PR #185, opened 2022-11-24
Don't let loaders override hardcoded configuration

pietern:
New approach of #170 (see [comment](https://github.com/databricks/databricks-sdk-go/pull/170#discussion_r1030401663)).

The approach in this PR is simple: we load a zero-initialized configuration with 1) environment variables, 2) databrickscfg, 3) environment variables again, to make sure they have precedence over databrickscfgs.

This configuration is then merged into the target configuration, ignoring attributes that were already set.

The net result is that:
* attributes that are set on the struct can never be overridden
* attributes configured through environment variables take precedence over attributes loaded from a configuration file


nfx:
i think this should be closed in favor of https://github.com/databricks/databricks-sdk-go/pull/306

PR #225, opened 2022-12-02
Host-based configuration

nfx:
Changes to the following configuration behavior:

ImplicitProfileLoader is an alternative config loader, which triggers if you specify  `host`, but no other credentials either through direct configuration or through environment variables, Databricks SDK for Go will try picking up profile with the matching host from ~/.databrickscfg. This allows keeping the hostname checked in to version control, but have ability to pick up different credentials either from local development machine or production server.

### DIFF
--- a/config/auth_permutations_test.go
+++ b/config/auth_permutations_test.go
@@ -235,21 +235,35 @@ func TestConfig_PatFromDatabricksCfg_NohostProfile(t *testing.T) {
 func TestConfig_Implicit_DatabricksCfg_Profile_ThroughHost(t *testing.T) {
 	configFixture{
 		env: map[string]string{
-			"HOME":            "testdata",
-			"DATABRICKS_HOST": "https://dbc-XXXXXXXX-ABCD.cloud.databricks.com",
+			"HOME":                       "testdata",
+			"DATABRICKS_PROFILE_RESOLVE": "true",
+			"DATABRICKS_HOST":            "https://dbc-XXXXXXXX-ABCD.cloud.databricks.com",
 		},
 		assertHost: "https://dbc-XXXXXXXX-ABCD.cloud.databricks.com",
 		assertAuth: "basic",
 	}.apply(t)
 }
 
+func TestConfig_Implicit_DatabricksCfg_Profile_ThroughHost_WithToken(t *testing.T) {
+	configFixture{
+		env: map[string]string{
+			"HOME":                       "testdata",
+			"DATABRICKS_PROFILE_RESOLVE": "true",
+			"DATABRICKS_HOST":            "https://dbc-XXXXXXXX-ABCD.cloud.databricks.com",
+			"DATABRICKS_TOKEN":           "abcd",
+		},
+		assertError: "validate: more than one authorization method configured: basic and pat. Config: host=https://dbc-XXXXXXXX-ABCD.cloud.databricks.com, token=***, username=abc, password=***, resolve_profile=true. Env: DATABRICKS_HOST, DATABRICKS_TOKEN, DATABRICKS_PROFILE_RESOLVE",
+	}.apply(t)
+}
+
 func TestConfig_Implicit_DatabricksCfg_Profile_Conflicts(t *testing.T) {
 	configFixture{
 		env: map[string]string{
-			"HOME":            "testdata",
-			"DATABRICKS_HOST": "https://adb-123.4.azuredatabricks.net",
+			"HOME":                       "testdata",
+			"DATABRICKS_PROFILE_RESOLVE": "true",
+			"DATABRICKS_HOST":            "https://adb-123.4.azuredatabricks.net",
 		},
-		assertError: "resolve: azure-justhost and azure-pat and azure-spn match https://adb-123.4.azuredatabricks.net in testdata/.databrickscfg. Config: host=https://adb-123.4.azuredatabricks.net. Env: DATABRICKS_HOST",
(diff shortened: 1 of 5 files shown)


config/config_file.go
```diff
@@ -43,6 +43,42 @@ func (l KnownConfigLoader) Configure(cfg *Config) error {
...
+			}
+			profiles = append(profiles, section.Name())
+		}
+		// in the real situations, we don't expect this to happen often (if not at all),
+		// hence we don't trim the list
+		if len(profiles) > 1 {
+			return fmt.Errorf("%s match %s in %s",
+				strings.Join(profiles, " and "),
+				cfg.Host, configFile)
+		}
```
pietern, line 75: This will error if you auth with `DATABRICKS_HOST` and `DATABRICKS_TOKEN`, and have multiple profiles with that host. Even though the creds would be valid. This path should only be taken as a last resort.

config/testdata/.databrickscfg
```diff
@@ -15,14 +15,14 @@ rate_limit=34
 host = https://adb-123.4.azuredatabricks.net/
 
 [azure-pat]
-host = https://adb-123.4.azuredatabricks.net/
+host = https://adb-123.4.azuredatabricks.net
```
pietern, line 18: Shouldn't the canonicalizer take care of this?

config/config_file.go
nfx: good one. will add a testcase for this.

P.S. could you be adding "Request changes" if you want these bigger changes, rather than "comment"? :)

config/testdata/.databrickscfg
nfx: this is to test canonicalizer

config/config_file.go
nfx: > validate: more than one authorization method configured: basic and pat. Config: host=https://dbc-XXXXXXXX-ABCD.cloud.databricks.com, token=***, username=abc, password=***. Env: DATABRICKS_HOST, DATABRICKS_TOKEN

yep, reproduced

nfx: i wonder what the semantics for `TestConfig_DatabricksCfgProfileResolve_ThroughHost` should be. it picks up `DEFAULT` profile if the default profile is there, but host is different from the one assigned in env.

config/auth_permutations_test.go
```diff
@@ -232,6 +234,50 @@ func TestConfig_PatFromDatabricksCfg_NohostProfile(t *testing.T) {
...
+	configFixture{
+		loaders: []Loader{
+			ConfigAttributes,
+			ImplicitProfileLoader{},
+		},
+		env: map[string]string{
+			"HOME":            "testdata",
+			"DATABRICKS_HOST": "https://adb-123.4.azuredatabricks.net",
+		},
+		assertError: "resolve: azure-justhost and azure-pat and azure-spn match https://adb-123.4.azuredatabricks.net in . Config: host=https://adb-123.4.azuredatabricks.net. Env: DATABRICKS_HOST",
```
pietern, line 277:
```
in . 
```
should say `in .databrickscfg` or whatever the path to the file is.

config/auth_permutations_test.go
```diff
@@ -232,6 +234,50 @@ func TestConfig_PatFromDatabricksCfg_NohostProfile(t *testing.T) {
...
+		loaders: []Loader{
+			ConfigAttributes,
+			ImplicitProfileLoader{},
+		},
+		env: map[string]string{
+			"HOME":             "testdata",
+			"DATABRICKS_HOST":  "https://dbc-XXXXXXXX-ABCD.cloud.databricks.com",
+			"DATABRICKS_TOKEN": "abcd",
+		},
+		assertError: "validate: more than one authorization method configured: basic and pat. Config: host=https://dbc-XXXXXXXX-ABCD.cloud.databricks.com, token=***, username=abc, password=***, profile=basic. Env: DATABRICKS_HOST, DATABRICKS_TOKEN",
```
pietern, line 263: This shouldn't fail. It is clear that the token should be used regardless in this case.

config/config_file.go
```diff
@@ -58,3 +50,115 @@ func (l KnownConfigLoader) Configure(cfg *Config) error {
...
+		return nil, fmt.Errorf("%s match %s in %s",
+			strings.Join(profiles, " and "), cfg.Host, cfg.ConfigFile)
+	}
+	return candidates[0], nil
+}
+
+func (l ImplicitProfileLoader) Configure(cfg *Config) error {
+	if cfg.Host == "" || cfg.Profile != "" {
+		// no host to match or profile already configured
```
pietern, line 129: Before running this we should try to configure the auth type and proceed only if this fails (nothing configured).

config/config_file.go
```diff
@@ -58,3 +50,115 @@ func (l KnownConfigLoader) Configure(cfg *Config) error {
...
+		return err
+	}
+	section, err := l.resolveSection(cfg, iniFile)
+	if err != nil {
+		return err
+	}
+	if section != nil {
+		cfg.Profile = section.Name()
+	} else {
+		cfg.Profile = "DEFAULT"
```
pietern, line 156: This should not happen -- it makes it possible for me to specify host A and connect to host B.

config/config_file.go
```diff
@@ -58,3 +50,115 @@ func (l KnownConfigLoader) Configure(cfg *Config) error {
...
+			continue
+		}
+		if canonical != cfg.Host {
+			continue
+		}
+		candidates = append(candidates, section)
+	}
+	if len(candidates) == 0 && iniFile.HasSection("DEFAULT") {
+		return iniFile.Section("DEFAULT"), nil
+	}
```
pietern, line 110: It shouldn't do this, same comment as below.

pietern (commented):
In the decision tree how to resolve which host we connect to we should do this if and only if no existing auth type can be configured. Otherwise we end up with too much ambiguity.

nfx:
implemented on the level of CLI

PR #228, opened 2022-12-05
Parse summary from descriptions

nfx:


pietern (approved):
What is materially different now?

I see the period after service/method titles. Anything else?

PR #239, opened 2022-12-09
Add concept of `main` service for the package

nfx:
This PR adds an advanced code generation feature for "main services" per package, which could be "flattened down" to the same namespace as the package itself, producing better end-user UX for the generated code. This manifests in the following templates:

package-level generation:
```
{{if .MainService -}}
    // code for a package with a main service
    {{load ".codegen/service.go.tmpl"}}
    {{template "service" .MainService}}
{{- else -}}
    // code for a package with multiple services
{{end}}
```

service-level generation:
```
{{if not .MatchesPackageName}}
	{{template "service" .}}
{{else}}
	{{skipThisFile}}
{{end}}

{{define "service"}}
	// code for a service
{{end}}
```

### DIFF
--- a/openapi/code/package.go
+++ b/openapi/code/package.go
@@ -39,6 +39,17 @@ func (pkg *Package) Services() (types []*Service) {
 	return types
 }
 
+// MainService returns a Service that matches Package name
+func (pkg *Package) MainService() *Service {
+	for _, svc := range pkg.services {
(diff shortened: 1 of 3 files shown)


pietern (commented):
@nfx Please add a test for this. I worry without exercising this path in this repo it will be easy to break and not know it's broken.

PR #247, opened 2022-12-14
Pass raw body as byte slice if requested

pietern:


nfx (approved):
thanks for removing redundant `path string`

PR #258, opened 2022-12-19
Cleaner callbacks for polling status

### LINKED ISSUE #251: Ability to tell loop iteration in retries.Info
This came up in a separate discussion. It would be nice to support (a subset of) the following pattern out of the box.

```go
var prevState *jobs.RunState

// This function is called each time the function below polls the run status.
update := func(info *retries.Info[jobs.Run]) {
  state := info.Info.State
  if state == nil {
    return
  }
  if prevState == nil {
    // First iteration
  }
  if prevState == nil || prevState.LifeCycleState != state.LifeCycleState {
    log.Printf("%s Run status: %s", prefix, info.Info.State.LifeCycleState)
    prevState = state
  }
}
```

nfx:
This PR ensures that `i.Info == nil` is a clear indicator for a long-running operation callback that it's run in the polling mode. For convenience, `retries.OnPoll` is added as well:

```go
clstr, err := w.Clusters.CreateAndWait(ctx, clusters.CreateCluster{
		ClusterName:            clusterName,
		SparkVersion:           latest,
		InstancePoolId:         GetEnvOrSkipTest(t, "TEST_INSTANCE_POOL_ID"),
		AutoterminationMinutes: 15,
		NumWorkers:             1,
	}, retries.Timeout[clusters.ClusterInfo](20*time.Minute),
		retries.OnPoll(func(i *clusters.ClusterInfo) {
			t.Logf("cluster is %s", i.State)
		}))
```

Fixes #251

### DIFF
--- a/retries/retries.go
+++ b/retries/retries.go
@@ -23,12 +23,12 @@ func Timeout[T any](dur time.Duration) Option[T] {
 	}
 }
 
-func OnPoll[T any](cb func(*T)) Option[T] {
+func OnPoll[T any](callback func(*T)) Option[T] {
 	return func(i *Info[T]) {
 		if i.Info == nil {
 			return
 		}
-		cb(i.Info)
+		callback(i.Info)
 	}
 }
 
--- a/retries/retries.go
+++ b/retries/retries.go
@@ -12,7 +12,6 @@ import (
 
 type Info[T any] struct {
 	Info    *T
-	Polling bool
 	Timeout time.Duration
 }
 
--- a/retries/retries.go
+++ b/retries/retries.go
@@ -11,7 +11,8 @@ import (
 )
 
 type Info[T any] struct {
-	Info    T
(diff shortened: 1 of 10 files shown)


retries/retries.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 type Info[T any] struct {
-	Info    T
+	Info    *T
+	Polling bool
```
pietern, line 15: What does it mean for this field to be false?

It looks like it is set to true everywhere in this PR and never accessed.

pietern (commented):
FWIW, the PR summary didn't make sense to me until I read the PR diff.

"enforces" made me look for an assert, but this changes the type to a pointer so that a nil value can have special meaning.

What is the motivation for this change?

pietern (changes requested):
Can you take a look at the comments?

PR #266, opened 2022-12-29
Added U2M auth

nfx:


config/auth_u2m.go
```diff
@@ -0,0 +1,67 @@
...
+type bricksCliTokenSource struct {
+	cfg *Config
+}
+
+func (ts *bricksCliTokenSource) Token() (*oauth2.Token, error) {
+	what := ts.cfg.Host
+	if ts.cfg.IsAccountClient() {
+		what = fmt.Sprintf("%s/oidc/accounts/%s", ts.cfg.Host, ts.cfg.AccountID)
+	}
+	out, err := exec.Command("bricks", "auth", "token", what).Output()
```
fjakobs, line 52: Where do you do `bricks login`? We should make this part of this flow.

fjakobs, line 52: We should have a config option and environment variable for the location of the bricks CLI. 
Actually we should do this for any CLI.

config/auth_default.go
```diff
@@ -13,6 +13,7 @@ var (
 	authProviders = []CredentialsProvider{
 		PatCredentials{},
 		BasicCredentials{},
+		BricksCliCredentials{},
```
pietern, line 16: As is, this should be `BricksOAuthCredentials`. Currently it implies it can do anything.

config/auth_u2m.go
```diff
@@ -0,0 +1,67 @@
...
+	if ee, ok := err.(*exec.ExitError); ok {
+		return nil, fmt.Errorf("cannot get access token: %s", string(ee.Stderr))
+	}
+	if err != nil {
+		return nil, fmt.Errorf("cannot get access token: %v", err)
+	}
+	var t oauth2.Token
+	err = json.Unmarshal(out, &t)
```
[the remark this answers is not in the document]
nfx: no, `bricks auth token` is only for U2M.

config/auth_u2m.go
nfx: let's start with this one. which variable do you use for bricks CLI?

config/auth_u2m.go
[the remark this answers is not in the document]
nfx: > If so, its output should be a struct containing oauth2.Token rather than a oauth2.Token verbatim.

why? :P PAT is yet another Bearer token ;)

PR #283, opened 2023-01-18
Let `google-credentials` take precedence over `google-id`

pietern:
It is possible that both `GoogleCredentials` and `GoogleServiceAccount` are set. Because `GoogleCredentials` is self-contained, it should have priority over `google-id`. This mirrors behavior of the TF provider.

### DIFF
--- a/config/auth_default.go
+++ b/config/auth_default.go
@@ -16,12 +16,12 @@ var (
 		M2mCredentials{},
 		BricksCliCredentials{},
 
-		// Attempt to configure the more specific auth methods before calling into the CLI.
+		// Attempt to configure auth from most specific to most generic (the Azure CLI).
 		AzureMsiCredentials{},
 		AzureClientSecretCredentials{},
 		AzureCliCredentials{},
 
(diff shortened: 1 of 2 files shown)


config/auth_default.go
```diff
@@ -15,11 +15,15 @@ var (
...
 		M2mCredentials{},
 		BricksCliCredentials{},
+
+		// Attempt to configure the more specific auth methods before calling into the CLI.
 		AzureMsiCredentials{},
 		AzureClientSecretCredentials{},
 		AzureCliCredentials{},
-		GoogleDefaultCredentials{},
+
+		// Attempt to configure the more specific auth methods before calling into the CLI.
```
nfx, line 24: `GoogleDefaultCredentials` does not call into the `gcloud` CLI. it just reads directly from `~/.config/gcloud/application_default_credentials.json` if it's present on the system, which makes this comment incorrect.

nfx (approved):

config/auth_default.go
pietern: Thanks, fixed.

PR #289, opened 2023-02-01
Log error responses

pietern:
The response body was read in multiple places before this change. To make response body logging possible this needed to be consolidated to a single place; the `attempt` function.

If there is a valid response the log will be of the form:

```
< HTTP/2.0 404 Not Found (RESOURCE_DOES_NOT_EXIST: ...)
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -233,8 +233,10 @@ func (c *DatabricksClient) recordRequestLog(
 	sb.WriteString("< ")
 	if response != nil {
 		sb.WriteString(fmt.Sprintf("%s %s", response.Proto, response.Status))
-		if err != nil {
-			sb.WriteString(fmt.Sprintf(" (%s)", err))
+		// Only display error on this line if the response body is empty.
+		// Otherwise the response body will include details about the error.
+		if len(responseBody) == 0 && err != nil {
+			sb.WriteString(fmt.Sprintf(" (Error: %s)", err))
 		}
 	} else {
 		sb.WriteString(fmt.Sprintf("Error: %s", err))
--- a/client/client.go
+++ b/client/client.go
@@ -182,8 +182,10 @@ func (c *DatabricksClient) attempt(
 		if err == nil && responseBodyErr != nil {
 			err = fmt.Errorf("response body: %w", responseBodyErr)
 		}
+
+		defer c.recordRequestLog(request, response, err, requestBody, responseBody.Bytes())
+
 		if err == nil {
-			c.recordRequestLog(request, response, nil, requestBody, responseBody.Bytes())
 			return &responseBody, nil
 		}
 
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -129,27 +128,14 @@ func (c *DatabricksClient) redactedDump(prefix string, body []byte) (res string)
...
-	method, requestURL string, requestBody []byte,
-	visitors ...func(*http.Request) error) func() (*http.Response, *retries.Err) {
-	return func() (*http.Response, *retries.Err) {
+func (c *DatabricksClient) attempt(
+	ctx context.Context,
+	method string,
+	requestURL string,
+	requestBody []byte,
+	visitors ...func(*http.Request) error,
+) func() (*bytes.Buffer, *retries.Err) {
```
nfx, line 137: How returning `*bytes.Buffer` would affect our ability to read response headers?

client/client.go
```diff
@@ -169,25 +155,57 @@ func (c *DatabricksClient) attempt(ctx context.Context,
...
+			_, responseBodyErr = responseBody.ReadFrom(response.Body)
+			response.Body.Close()
+		}
+
+		retry, err := apierr.CheckForRetry(ctx, response, err, responseBody.Bytes(), responseBodyErr)
+		if err != nil && !errors.As(err, &apierr.APIError{}) {
+			err = fmt.Errorf("failed request: %w", err)
+		}
+		if err == nil && response == nil {
+			err = fmt.Errorf("no response: %s %s", method, requestURL)
```
nfx, line 180: this _technically_ should not happen because default http client always has to return a response, but it's good to check, yes

client/client.go
```diff
@@ -169,25 +155,57 @@ func (c *DatabricksClient) attempt(ctx context.Context,
...
+		}
+		if err == nil && responseBodyErr != nil {
+			err = fmt.Errorf("response body: %w", responseBodyErr)
+		}
 		if err == nil {
-			return resp, nil
+			c.recordRequestLog(request, response, nil, requestBody, responseBody.Bytes())
```
nfx, line 186: what about adding `defer` for this?

client/client.go
pietern: It wouldn't allow for that, but we don't do this today. If we need to we can add back the `http.Response`. 

I removed it because we can only use a single return argument from `retries.Poll` and would need to wrap a tuple with `http.Response` and `bytes.Buffer` for the body otherwise.

client/client.go
pietern: Copied from the other code to make all tests pass ;-)

client/client.go
pietern: We could, but it implies we also log all requests that get retried. What do you think?

nfx: isn't it the only way to guarantee to log failed requests? :)

pietern: True, but this includes e.g. rate limiting.

Updated to log all requests with a `defer`.

nfx (approved):

PR #295, opened 2023-02-03
Test with Go 1.20

nfx:
Each major Go release is supported until there are two newer major releases.


pietern:
I saw the same issue for 1.18 w.r.t. staticcheck. I think it is pulling in the latest through a different dependency.

nfx:
> I saw the same issue for 1.18 w.r.t. staticcheck. I think it is pulling in the latest through a different dependency.

yep, fixed that.

pietern (approved):

PR #298, opened 2023-02-03
fix[#285]: change APIError to use pointer receiver

### LINKED ISSUE #285: APIError not using pointer receiver
`APIError` implementation of `Error()` does no use pointer receiver that while does not break `errors.As` it changes how you expect to use it.

```go
_, err := db.w.Catalogs.GetByName(ctx, name)
if err != nil {
	var apiErr *apierr.APIError
	if errors.As(err, &apiErr)
            // This block is never called
	}
	return false, fmt.Errorf("catalogExists: %w", err)
}
```

I know that `IsMissing` is available to do this check but would seem better to have `APIError` work as expected.

Happy to open a PR if this is something that would be accepted

amwill04:
Allows APIError to be used as expected with `errors.As`. Closes #285.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -173,7 +173,8 @@ func (c *DatabricksClient) attempt(
 		}
 
 		retry, err := apierr.CheckForRetry(ctx, response, err, responseBody.Bytes(), responseBodyErr)
-		if err != nil && !errors.As(err, &apierr.APIError{}) {
+		var apiErr *apierr.APIError
+		if err != nil && !errors.As(err, &apiErr) {
 			err = fmt.Errorf("failed request: %w", err)
 		}
 		if err == nil && response == nil {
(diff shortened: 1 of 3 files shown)


nfx (approved):

client/client.go
```diff
@@ -173,7 +173,8 @@ func (c *DatabricksClient) attempt(
 		}
 
 		retry, err := apierr.CheckForRetry(ctx, response, err, responseBody.Bytes(), responseBodyErr)
-		if err != nil && !errors.As(err, &apierr.APIError{}) {
+		var apiErr *apierr.APIError
+		if err != nil && !errors.As(err, &apiErr) {
```
pietern, line 177: This now takes a pointer to a pointer. I don't understand why this is necessary.

pietern: I'll take a look at the linked issue first.

pietern: I see the linked documentation that shows defining a pointer type prior to calling `errors.As`.

LGTM, thanks for the PR!

pietern (approved):

PR #299, opened 2023-02-07
Drop duplicate prefix in randomized email

pietern:
As is, the integration tests created email addresses such as `sdk-go-sdk-go-bflaglaacell@example.com`.


nfx (approved):
lgtm

PR #306, opened 2023-02-16
Skip loading config if auth is already explicitly configured directly

### LINKED ISSUE #304: `~/.databrickscfg` is loaded even when direct config has precedence
steps to reproduce:
* direct host & az resource ID
* configuration file with DEFAULT profile with `token` + `host`
* expected: azure CLI working for that host
* happens: `validate: more than one authorization method configured: azure and pat` error

this issue blocks https://github.com/databricks/terraform-provider-databricks/pull/1848

```go
func TestConfig_AzureCliHostAndResourceID(t *testing.T) {
	p, _ := filepath.Abs("../common/testdata")
	providerFixture{
		// omit request to management endpoint to get workspace properties
		azureResourceID: azResourceID,
		host:            "x",
		env: map[string]string{
			// these may fail on windows. use docker container for testing.
			"PATH": p,
			"HOME": p,
		},
		assertAzure: true,
		assertHost:  "https://x",
		assertAuth:  "azure-cli",
	}.apply(t)
}
```

nfx:
Fix #304

This issue unblocks https://github.com/databricks/terraform-provider-databricks/pull/1848

### DIFF
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -19,7 +19,7 @@ func (l KnownConfigLoader) Configure(cfg *Config) error {
 	// Skip loading config file if some authentication is already explicitly
 	// configured directly in the config by a user.
 	// See: https://github.com/databricks/databricks-sdk-go/issues/304
-	if l.isAnyAuthConfigured(cfg) {
+	if cfg.Profile == "" && l.isAnyAuthConfigured(cfg) {
 		return nil
 	}
 	configFile := cfg.ConfigFile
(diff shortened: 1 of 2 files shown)


config/config_file.go
```diff
@@ -16,6 +16,12 @@ func (l KnownConfigLoader) Name() string {
 }
 
 func (l KnownConfigLoader) Configure(cfg *Config) error {
+	// Skip loading config file if some authentication is already explicitly
+	// configured directly in the config by a user.
+	// See: https://github.com/databricks/databricks-sdk-go/issues/304
+	if l.isAnyAuthConfigured(cfg) {
+		return nil
+	}
```
pietern, line 24: This should only happen if `cfg.Profile == ""`. Otherwise it may not pick up what is needed from the config file.

pietern (commented):
Looks reasonable. Please check out the comment.

Reminds me of #185.

PR #309, opened 2023-02-17
Always use the latest value for user agent key

nfx:


useragent/user_agent.go
```diff
@@ -97,6 +85,24 @@ func (u info) String() string {
...
+	if err := matchAlphanum(key); err != nil {
+		panic(err)
+	}
+	if err := matchAlphanumOrSemVer(value); err != nil {
+		panic(err)
+	}
+	for i := range d {
+		if d[i].Key == key {
+			d[i].Value = value
+			return d
```
pietern, line 100: This should return a copy of the slice, not do an in-place modification.

As is, this will affect a parent context as well.

pietern (commented):
Why is setting the same key twice not an error?

useragent/user_agent_test.go
```diff
@@ -33,12 +33,16 @@ func TestFromContext_Default(t *testing.T) {
 func TestFromContext_Custom(t *testing.T) {
 	WithProduct("unit-tests", "0.0.1")
 	WithUserAgentExtra("pulumi", "3.8.4")
+	WithUserAgentExtra("terraform", "1.3.4")
+	WithUserAgentExtra("terraform", "1.3.5")
+	WithUserAgentExtra("terraform", "1.3.6")
 	ctx := InContext(context.Background(), "a", "b")
-	ctx = InContext(ctx, "terraform", "1.2.5")
+	ctx = InContext(ctx, "a", "d")
+	ctx = InContext(ctx, "a", "e")
```
pietern, line 41: Opportunity to have ctx1 and ctx2 and test that it doesn't leak across contexts to test the copy.

pietern (approved):
LGTM

useragent/user_agent_test.go
nfx: one sec, will add

nfx: added

PR #330, opened 2023-03-08
Add option to configure the location of the bricks CLI

fjakobs:
Especially for the VS Code extension this is easier and more robust than messing the the `PATH`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -73,6 +73,9 @@ type Config struct {
 	ClientID     string `name:"client_id" env:"DATABRICKS_CLIENT_ID" auth:"oauth"`
 	ClientSecret string `name:"client_secret" env:"DATABRICKS_CLIENT_SECRET" auth:"oauth,sensitive"`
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -73,6 +73,9 @@ type Config struct {
 	ClientID     string `name:"client_id" env:"DATABRICKS_CLIENT_ID" auth:"oauth"`
 	ClientSecret string `name:"client_secret" env:"DATABRICKS_CLIENT_SECRET" auth:"oauth,sensitive"`
 
+	// Path to the 'bricks' CLI
+	BricksCli string `name:"bricks_cli" env:"BRICKS_CLI"`
```
nfx, line 77: i'm not sure we should leak this to `Config` type, but we should allow the override through the environment variable. e.g. only `config/auth_u2m.go` to change in this PR

Environment variable should be `BRICKS_CLI_PATH` to be more consistent.

What do you think?

nfx (approved):

PR #333, opened 2023-03-09
Make logger context aware

pietern:
This commit changes a few things:
* The logging functions in the `logger` package now take a context argument. A logger from the context is used if one is configured, otherwise these functions use the global `DefaultLogger`.
* Add functions `NewContext` to configure a context with a logger and `FromContext` to retrieve a logger from a context.
* The `logger.Logger` interface defines an `Enabled` function to check if logging at a level is enabled or not. This means we no longer compute a redacted copy of the response body for every response if debug logging is not enabled.
* An example under `examples/slog` that shows how to use the SDK with slog.
* An example under `examples/zerolog` that shows how to use the SDK with zerolog.


apierr/errors.go
```diff
@@ -73,10 +73,15 @@ func (apiError *APIError) IsTooManyRequests() bool {
...
+	return apiError.isRetriable(context.Background())
+}
+
+// isRetriable returns true if error is retriable
+func (apiError *APIError) isRetriable(ctx context.Context) bool {
 	// Handle transient errors for retries
 	for _, substring := range transientErrorStringMatches {
 		if strings.Contains(apiError.Message, substring) {
-			logger.Debugf("Attempting retry because of %#v", substring)
+			logger.Debugf(ctx, "Attempting retry because of %#v", substring)
```
pietern, line 84: @nfx This is needed so that this log entry can be associated with log lines for the same request.

config/auth_azure_cli.go
```diff
@@ -82,7 +82,7 @@ func (ts *azureCliTokenSource) Token() (*oauth2.Token, error) {
 	if err != nil {
 		return nil, fmt.Errorf("cannot parse expiry: %w", err)
 	}
-	logger.Infof("Refreshed OAuth token for %s from Azure CLI, which expires on %s",
+	logger.Infof(context.Background(), "Refreshed OAuth token for %s from Azure CLI, which expires on %s",
```
pietern, line 85: @nfx Ideally we have access to a context here as well but `oauth2.TokenSource` doesn't deal with contexts...

logger/logger.go
```diff
@@ -1,55 +1,70 @@
...
-	DefaultLogger.Tracef(format, v...)
-}
+// Level maps to the logging levels in [Logger].
+type Level int
 
-func Debugf(format string, v ...interface{}) {
-	DefaultLogger.Debugf(format, v...)
-}
+const (
+	LevelTrace = -8
```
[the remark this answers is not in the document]
nfx: currently trace is too noisy, but rarely needed. i'd keep it.

and use debug for what we use info today

logger/logger.go
```diff
@@ -1,55 +1,70 @@
...
+	Debugf(format string, v ...any)
 
-type Logger interface {
-	Tracef(format string, v ...interface{})
-	Debugf(format string, v ...interface{})
-	Infof(format string, v ...interface{})
-	Warnf(format string, v ...interface{})
-	Errorf(format string, v ...interface{})
+	// Infof logs a formatted string at [LevelInfo].
+	Infof(format string, v ...any)
```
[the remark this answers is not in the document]
pietern: How would zerolog benefit?

Added the context argument to the interface.

logger/simple.go
```diff
@@ -0,0 +1,32 @@
+package logger
+
+import (
+	"context"
+	"log"
+)
+
+type SimpleLogger struct{}
+
+func (l *SimpleLogger) Enabled(_ context.Context, level Level) bool {
```
nfx, line 10: should we do any logic in this one?

nfx (approved):

PR #335, opened 2023-03-10
Added `id` field to the `SparkVersionRequest` for easier integration with Terraform's `databricks_spark_version` data source

alexott:


nfx (approved):

pietern:
@alexott Why does this make it eaiser?

alexott:
@pietern Because of the Go's inheritance implementation I can't easily to extend the existing structure defined in the SDK and use it with `common.WorkspaceData`, so I either need to go with full implementation of the `Read`, or copy struct into the terraform

pietern:
Does composition with anonymous fields not work? E.g.

```go
type MySparkVersionRequest {
  ID string
  *clusters.SparkVersionRequest
}
```

Might need some change to traverse those in the reflection code.

alexott:
No, it generate nested object. Making the changes in the reflection - how we would decide if the object should be flattened or not? Maybe it's a good topic for discussion

PR #339, opened 2023-03-13
Add pipeline events API to the SDK

shreyas-goenka:


pietern (approved):
Can you confirm you checked that multi-page events also worked?

PR #340, opened 2023-03-14
Authenticate using a local server

fjakobs:
This is similar to Azure MSI except that the server endpoint is configurable and optionally can also return the host.

Naming is still TBD.

Example:

```js
const http = require('http');

const server = http.createServer((req, res) => {
    if (req.headers["Metadata"] !== "true") {
        res.statusCode = 403;
        res.end();
        return;
    }

    res.end(JSON.stringify({
        host: process.env["DATABRICKS_HOST"],
        access_token: process.env["DATABRICKS_TOKEN"],
        expires_on: Math.floor(Date.now() / 1000) + 30,
        token_type: "Bearer"
    }));
});

server.listen(3000, "127.0.0.1", () => {});
```

```
% DATABRICKS_AUTH_SERVER_URL=http://localhost:3000 go run examples/default-auth/main.go
2023/03/14 14:12:28 [TRACE] Loading config via environment
2023/03/14 14:12:28 [TRACE] Loading config via config-file
2023/03/14 14:12:28 [TRACE] Attempting to configure auth: auth-server
2023/03/14 14:12:28 [INFO] Refreshed access token for http://localhost:3000 from local credentials server, which expires on 2023-03-14 14:12:58 +0100 CET
2023/03/14 14:12:29 [DEBUG] GET /api/2.0/clusters/list
...
```

## TODO

- [x] Remove `local` from names
- [x] Leverage `oauth2.Token` in server response
- [x] Add versioning to the protocol - client requests specific version(s)
- [x] Unit test for token refresh


config/auth_server.go
```diff
@@ -0,0 +1,161 @@
...
+	"time"
+
+	"github.com/databricks/databricks-sdk-go/logger"
+	"golang.org/x/oauth2"
+)
+
+// timeout to wait for local auth server to respond
+const authServerTimeout = 10 * time.Second
+
+type serverResponse struct {
```
nfx, line 19: you should use `oauth2.Token` type for this one. it's already well-intergrated.

config/auth_server.go
```diff
@@ -0,0 +1,161 @@
...
+	// Optional URL of the Databricks host to connect to.
+	Host string `json:"host"`
```
nfx, line 21: why not just use the `DATABRICKS_HOST` environment variable? it's a different thing from a token and is more "static", whereas a token is refreshable.

config/auth_server.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("token parse: Invalid URL %s", response.Host)
+	}
+
+	response.Host = parsedHost.String()
+
+	return &response, nil
+}
+
+type authServerTokenSource string
```
nfx, line 137: could you refactor `azureMsiTokenSource` and make it customizable to the extent you need it to be customizable? i suspect those to share 80-90% of the code.

config/auth_default.go
```diff
@@ -11,6 +11,7 @@ import (
 
 var (
 	authProviders = []CredentialsProvider{
+		ServerCredentials{},
```
nfx, line 14: Could those be before the `BricksCliCredentials`?..

nfx (changes requested):

config/auth_server.go
fjakobs: I always pass a `HOST`, `TOKEN` pair so the host based auth can determine if the metadata service is applicable to the configured host.

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+func (t metadataServiceTokenSource) Token() (*oauth2.Token, error) {
+	token, err := makeRequest(t.metadataServiceURL)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if token.Host != t.databricksHost {
+		return nil, fmt.Errorf("host in token (%s) doesn't match configured host (%s)", token.Host, t.databricksHost)
+	}
```
pietern, line 136: We should take it one step further and add an opaque "session" identifier or something that we can check.

It is possible that the session changes (e.g. other metadata we add later) while the host remains the same.

If that is the case this session should also be invalidated.

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+	defer resp.Body.Close()
+	if resp.StatusCode == 404 {
+		return nil, nil
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("token read: %w", err)
+	}
+	if resp.StatusCode != 200 {
```
pietern, line 102: 200 == `http.StatusOK`.

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+
+	ts := metadataServiceTokenSource{
+		metadataServiceURL: parsedMetadataServiceUrl,
+		databricksHost:     cfg.Host,
+	}
+	return refreshableVisitor(&ts), nil
+}
+
+// makes a request to the server and returns the token
+func makeRequest(serverUrl *url.URL) (*serverResponse, error) {
```
pietern, line 80: Make implies construction of the request, not performing the request.

Potential name: "performRequest".

Also, you can do:
```go
type metadataService struct {
  *url.URL
}

func (s metadataService) Get() (*metadata, error)
```

To avoid references to the literal URL and a global function.

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil, fmt.Errorf("token request: %w", err)
+	}
+	req.Header.Add("Metadata", "true")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("token response: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == 404 {
```
pietern, line 95: 404 == `http.StatusNotFound`.

config/auth_local_metadata_service_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	req := &http.Request{
+		Header: http.Header{},
+	}
+
+	err = authProvider(req)
+	require.NoError(t, err)
+
+	require.Equal(t, req.Header.Get("Authorization"), fmt.Sprintf("Bearer %s", token))
+}
```
pietern, line 73: Could you add a test to verify it gets called twice if refresh is required?

This could use an incrementing integer as the "token" or something.

config/config.go
```diff
@@ -38,6 +38,9 @@ type Config struct {
 	// Databricks host (either of workspace endpoint or Accounts API endpoint)
 	Host string `name:"host" env:"DATABRICKS_HOST"`
 
+	// URL of the local metadata service that provides authentication credentials.
+	LocalMetadataServiceUrl string `name:"local_metadata_service_url" env:"DATABRICKS_LOCAL_METADATA_SERVICE_URL" auth:"local-metadata-service,sensitive"`
```
pietern, line 42: Some of the other fields use initialisms, e.g. `AccountID`. To be consistent this should be `LocalMetadataServiceURL`.

config/auth_local_metadata_service.go
fjakobs: done

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+	AccessToken string `json:"access_token"`
+
+	// The type of token, which is a "Bearer" access token, which means the
+	// resource can give access to the bearer of this token.
+	TokenType string `json:"token_type"`
+
+	// The timespan when the access token expires.
+	// The date is represented as the number of seconds from "1970-01-01T0:0:0Z UTC"
+	// If zero the token is valid forever.
+	ExpiresOn json.Number `json:"expires_on"`
```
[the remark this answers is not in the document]
fjakobs: Cool, done

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+}
```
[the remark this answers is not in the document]
fjakobs: done

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+//	token_type: The type of token, which is a "Bearer" access token.
+//	expires_on: The timespan when the access token expires.
+type LocalMetadataServiceCredentials struct{}
+
+func (c LocalMetadataServiceCredentials) Name() string {
+	return "local-metadata-service"
+}
+
+func (c LocalMetadataServiceCredentials) Configure(ctx context.Context, cfg *Config) (func(*http.Request) error, error) {
+	if cfg.LocalMetadataServiceUrl == "" {
```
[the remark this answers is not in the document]
fjakobs: done

config/auth_local_metadata_service.go
```diff
@@ -0,0 +1,171 @@
...
+// The header "Metadata: true" will be added to the request, which must be
+// verified by server to prevent SSRF attacks
+//
+// The server is expected to return a JSON response with the following fields:
+//
+//	host: URL of the Databricks host to connect to.
+//	access_token: The requested access token.
+//	token_type: The type of token, which is a "Bearer" access token.
+//	expires_on: The timespan when the access token expires.
+type LocalMetadataServiceCredentials struct{}
```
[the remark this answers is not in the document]
fjakobs: Removed local from all names

config/auth_metadata_service.go
```diff
@@ -0,0 +1,159 @@
...
+	return &response, nil
+}
+
+type metadataServiceTokenSource struct {
+	metadataService *metadataService
+	databricksHost  string
+}
+
+func (t metadataServiceTokenSource) Token() (*oauth2.Token, error) {
+	token, err := t.metadataService.Get()
```
pietern, line 150: Var name should be response, or resp, or something. It includes the token.

pietern (commented):
This LGTM bar the comments. @nfx Can you also take a pass?

config/auth_metadata_service.go
```diff
@@ -0,0 +1,159 @@
...
+//	host: URL of the Databricks host to connect to.
+//	token: object with the following fields
+//    access_token: The requested access token.
+//	  token_type: The type of token, which is a "Bearer" access token.
+//	  expires_on: The timespan when the access token expires.
+
+type MetadataServiceCredentials struct{}
+
+func (c MetadataServiceCredentials) Name() string {
+	return "local-metadata-service"
```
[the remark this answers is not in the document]
fjakobs: done

config/auth_metadata_service.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/databricks/databricks-sdk-go/logger"
+	"golang.org/x/oauth2"
+)
+
+// timeout to wait for local auth server to respond
+const metadataServiceTimeout = 10 * time.Second
+
+// response expected from the metadata service
+type serverResponse struct {
+	// URL of the metadata service to connect to.
```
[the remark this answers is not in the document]
fjakobs: done

config/auth_metadata_service.go
```diff
@@ -0,0 +1,159 @@
...
+func (s metadataService) Get() (*serverResponse, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), metadataServiceTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, s.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("token request: %w", err)
+	}
+
+	req.Header.Add("X-Databricks-Metadata-Version", "1")
```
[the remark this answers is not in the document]
fjakobs: done

config/auth_metadata_service.go
fjakobs: done

config/auth_metadata_service.go
```diff
@@ -0,0 +1,159 @@
...
+// server returns a token for a Host on a given URL path, it MUST continue to return
+// tokens for the same Host.
+//
+// The server is expected to return a JSON response with the following fields:
+//
+//	host: URL of the Databricks host to connect to.
+//	token: object with the following fields
+//    access_token: The requested access token.
+//	  token_type: The type of token, which is a "Bearer" access token.
+//	  expires_on: The timespan when the access token expires.
```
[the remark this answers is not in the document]
fjakobs: indeed. fixed

config/auth_local_metadata_service_test.go
fjakobs: done

config/auth_default.go
```diff
@@ -11,6 +11,7 @@ import (
 
 var (
 	authProviders = []CredentialsProvider{
+		MetadataServiceCredentials{},
```
nfx, line 14: Please move it to the bottom of the list, we don't assume it'll be triggered for top-10% of the products built on top of Go SDK

config/auth_metadata_service.go
```diff
@@ -0,0 +1,166 @@
...
+// timeout to wait for metadata service to respond
+const metadataServiceTimeout = 10 * time.Second
+
+const MetadataServiceVersion = "1"
+
+// response expected from the metadata service
+// other fields such as `cluster_id` might be added later
+type serverResponse struct {
+	// URL of the Databricks workspace to connect to
+	Host string `json:"host"`
```
nfx, line 25: why do we need it? can't we pass it via `DATABRICKS_HOST` environment variable?

config/auth_metadata_service.go
```diff
@@ -0,0 +1,166 @@
...
+// host: URL of the Databricks host to connect to.
+// token: object with the following fields
+//   access_token: The requested access token.
+//	 token_type: The type of token, which is a "Bearer" access token.
+//	 expires_on: The timespan when the access token expires.
+
+type MetadataServiceCredentials struct{}
+
+func (c MetadataServiceCredentials) Name() string {
+	return "metadata-service"
```
nfx, line 53: please don't rename the name of auth type every other week - it pollutes the stats :) 
image

config/auth_metadata_service.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (c MetadataServiceCredentials) Configure(ctx context.Context, cfg *Config) (func(*http.Request) error, error) {
+	if cfg.MetadataServiceURL == "" {
+		return nil, nil
+	}
+
+	parsedMetadataServiceURL, err := url.Parse(cfg.MetadataServiceURL)
```
nfx, line 61: please make sure there's a reuse between Azure MSI and this and there's compatibility. that's the main requirement for this PR to pass, otherwise it explodes the complexity for all SDKs.

config/auth_server.go
nfx: it'd consider it a blocker if we don't return `oauth2.Token`-compatible response, because that's what we already have integrations in the other languages.

config/auth_metadata_service.go
fjakobs: Isn't that the purpose of a review to align on names and update after a discussion. I don't want to end up in a situation where we need to agree on names even before we start coding.

config/auth_metadata_service.go
fjakobs: I've looked at the diff between the two implementations. The two are similar but not similar enough to share implementation without increasing complexity.

The only way to meaningfully unify would be to use exactly the MSI protocol and wire format but that prevents us evolving the protocol e.g. adding `host` and `cluster_id` to the payload. Even if we do that only the ~40 lines of the `Token` function can be re-used https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_msi.go#LL97-L136C2

config/auth_metadata_service.go
```diff
@@ -0,0 +1,166 @@
...
+// server returns a token for a Host on a given URL path, it MUST continue to return
+// tokens for the same Host.
+//
+// The server is expected to return a JSON response with the following fields:
+//
+// host: URL of the Databricks host to connect to.
+// token: object with the following fields
+//   access_token: The requested access token.
+//	 token_type: The type of token, which is a "Bearer" access token.
+//	 expires_on: The timespan when the access token expires.
```
pietern, line 48: @fjakobs Indenting issue persist.

config/auth_metadata_service.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	// only allow localhost URLs
+	if parsedMetadataServiceURL.Hostname() != "localhost" && parsedMetadataServiceURL.Hostname() != "127.0.0.1" {
+		return nil, fmt.Errorf("invalid auth server URL: %s", cfg.MetadataServiceURL)
+	}
+
+	ms := metadataService{parsedMetadataServiceURL}
+
+	response, err := ms.Get()
```
pietern, line 73: Arguably if the metadata service URL is configured and we get either an error, a nil response, or a host mismatch (all of the checks below), this should return an error instead of silently continuing.

config/auth_metadata_service.go
fjakobs: effing tabs. They are gone now

config/auth_metadata_service.go
fjakobs: Agreed. Changed that behaviour as suggested

config/auth_metadata_service.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if response == nil {
+		return nil, nil
+	}
+
+	// ignore this credential provider if it returns tokens for a different host
+	// if no host is configured yet then take the host from the metadata service
+	if cfg.Host != "" && cfg.Host != response.Host {
+		logger.Debugf(ctx, "ignoring metadata service because it returned a token for a different host: %s", response.Host)
+		return nil, nil
```
pietern, line 87: @fjakobs This should return an error as well.

fjakobs: For this I'm not so sure. Imagine you use host based auth and have a host configured. If a metadata service is available that can provide credentials for that host then it should be used but if the metadata service credentials are for another host then I think we should fall through and try the next method.

What do you think?

config/auth_default.go
nfx: Can you move it to the bottom of the list , or at least below the `BricksCliCredentials`, because it's very risky to execute it first in the chain.

config/auth_metadata_service.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, fmt.Errorf("token read: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("token error: %s", body)
+	}
+
+	var response serverResponse
+	if err := json.Unmarshal(body, &response); err != nil {
+		return nil, fmt.Errorf("token parse: %w", err)
+	}
```
nfx, line 131: this looks like a duplicate from 
https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_msi.go#L125-L145

for example, can we consider creating the token source, calling `.Token()` on it to verify if it's correctly initialized and then return a refreshable visitor, like Azure CLI provider does?

https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_cli.go#L34-L35

config/auth_metadata_service.go
nfx: can you extend/reuse the `azureMsiTokenSource` - https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_msi.go#L104-L157. You can easily do that by passing down `func(*http.Request) error` callback, which will do something like:

```
customizeMetadata := func (req *http.Request) error {
    req.Header.Add(MetadataServiceVersionHeader, MetadataServiceVersion)
    req.Header.Add(MetadataServiceHostHeader, s.config.Host)
   return nil
}

customizeMsi := func (req *http.Request) error {
    query := req.URL.Query()
	query.Add("api-version", "2018-02-01")
	query.Add("resource", s.resource)
	if s.clientId != "" {
		query.Add("client_id", s.clientId)
	}
	req.URL.RawQuery = query.Encode()
	req.Header.Add("Metadata", "true")
    return nil
}
```

config/auth_metadata_service.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// The server MUST return a 4xx response if the Host passed in the "X-Databricks-Host"
+// header doesn't match the token.
+//
+// The server is expected to return a JSON response with the following fields:
+//
+// - access_token: The requested access token.
+// - token_type: The type of token, which is a "Bearer" access token.
+// - expires_on: Unix timestamp when the access token expires.
+
```
nfx, line 51: remove the empty line between the type and the comment, if you want this comment to appear as documentation on pkg.go.dev

config/auth_metadata_service.go
```diff
@@ -0,0 +1,169 @@
...
+const MetadataServiceHostHeader = "X-Databricks-Host"
+
+// response expected from the metadata service
+type serverResponse struct {
+	AccessToken string `json:"access_token"`
+	TokenType   string `json:"token_type"`
+
+	// epoch in seconds since 1970-01-01T00:00:00Z
+	ExpiresOn json.Number `json:"expires_on"`
+}
```
nfx, line 30: lines 24-30 are duplicates of https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_msi.go#L153-L157

nfx (changes requested):
Could you try extracting the common code path with MSI - I've suggested an easy way on line 63. Otherwise, we risk too much deviating from MSI protocol

nfx (approved):

PR #349, opened 2023-03-29
Add type to represent a loaded configuration file

pietern:
## Changes

Export type for a loaded configuration file such that downstream applications can read its contents in the same way the SDK does. Rename the loader instance to `ConfigFile` to make it symmetric with `ConfigAttributes`.

## Tests

This doesn't change functionality.

- [x] `make test` passing
- [x] `make fmt` applied


nfx (approved):
lgtm, as long as integration tests are passing

pietern:
Integration tests all pass.

PR #358, opened 2023-04-05
add `isTesting` marker to support `ResourceFixture` in Terraform

nkvuong:
## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -84,9 +84,6 @@ type Config struct {
 	// Use at your own risk or for unit testing purposes.
 	InsecureSkipVerify bool `name:"skip_verify" auth:"-"`
 
-	// marker for testing fixture
-	IsTesting bool
-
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -84,6 +84,9 @@ type Config struct {
 	// Use at your own risk or for unit testing purposes.
 	InsecureSkipVerify bool `name:"skip_verify" auth:"-"`
 
+	// marker for testing fixture
+	IsTesting bool
```
nfx, line 88: can we make it private and move down to the list of arguments? and enable testing with `func (c *Config) WithTesting() *Config { c.isTesting = true; return c; }`?

nfx (changes requested):

nfx (approved):

PR #362, opened 2023-04-06
Update debug messages in config loader to display correct path

pietern:
## Changes

Debug messages would include a stringified version of the `configFile` object instead of the path.

## Tests

- [x] `make test` passing

### DIFF
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -22,13 +22,19 @@ func (f *File) Path() string {
 	return f.path
 }
 
+func defaultConfigFilePath(path string) string {
+	if path == "" {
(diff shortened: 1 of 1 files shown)


mgyucht (approved):
Code LGTM. 2 notes:
1. Did you do a manual test? If so, could you paste the test into the PR description so we can see the new behavior?
2. IIUC your PR description is what is currently happening? Something like "This PR improves logging during config loading by logging the path to the .databrickscfg file instead of the serialized config object." would be a bit more clear about the intent.

pietern:
@mgyucht Updated the PR description with before/after and link to relevant previous PR.

PR #363, opened 2023-04-06
Skip loading default profile if host is already configured

pietern:
## Changes

If a user specified a host to connect to without authentication details (either directly or via the `DATABRICKS_HOST` environment variable), the config file loader would load the profile named `DEFAULT` from the config file and overwrite whatever was already configured. This means it was possible to configure the host of workspace A and end up connecting to workspace B (if configured as `DEFAULT` profile in the config file). This PR prevents this behavior.

Also see https://github.com/databricks/databricks-sdk-go/issues/304.

## Tests

- Added unit tests.
- Confirmed behavior with manual test.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -150,7 +150,7 @@ func (c *Config) IsGcp() bool {
 
 // IsAws returns if the client is configured for Databricks on AWS.
 func (c *Config) IsAws() bool {
-	return strings.Contains(c.Host, ".databricks.com") && !c.IsAzure() && !c.IsGcp()
+	return c.Host != "" && !c.IsAzure() && !c.IsGcp()
 }
 
 // IsAccountClient returns true if client is configured for Accounts API
--- a/config/config.go
+++ b/config/config.go
@@ -134,7 +134,7 @@ func (c *Config) Authenticate(r *http.Request) error {
 	return c.auth(r)
 }
 
-// IsAzure returns true if client is configured for Azure Databricks
+// IsAzure returns if the client is configured for Azure Databricks.
 func (c *Config) IsAzure() bool {
 	isAzureHost := strings.Contains(c.Host, ".azuredatabricks.net") ||
 		strings.Contains(c.Host, "databricks.azure.cn") ||
@@ -143,14 +143,14 @@ func (c *Config) IsAzure() bool {
 	return isAzureHost || c.AzureResourceID != ""
 }
 
-// IsGcp returns true if client is configured for GCP
+// IsGcp returns if the client is configured for Databricks on Google Cloud.
 func (c *Config) IsGcp() bool {
 	return strings.Contains(c.Host, ".gcp.databricks.com")
 }
 
-// IsAws returns true if client is configured for AWS
+// IsAws returns if the client is configured for Databricks on AWS.
 func (c *Config) IsAws() bool {
-	return !c.IsAzure() && !c.IsGcp()
+	return strings.Contains(c.Host, ".databricks.com") && !c.IsAzure() && !c.IsGcp()
 }
 
 // IsAccountClient returns true if client is configured for Accounts API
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -143,14 +143,14 @@ func (c *Config) IsAzure() bool {
...
+// IsGcp returns if the client is configured for Databricks on Google Cloud.
 func (c *Config) IsGcp() bool {
 	return strings.Contains(c.Host, ".gcp.databricks.com")
 }
 
-// IsAws returns true if client is configured for AWS
+// IsAws returns if the client is configured for Databricks on AWS.
 func (c *Config) IsAws() bool {
-	return !c.IsAzure() && !c.IsGcp()
+	return strings.Contains(c.Host, ".databricks.com") && !c.IsAzure() && !c.IsGcp()
```
nfx, line 153: we have other top-level domains for AWS (or PVC), this won't be sufficient and may break some deployments

config/config_file.go
```diff
@@ -65,7 +65,7 @@ func (l configFileLoader) Configure(cfg *Config) error {
 	// Skip loading config file if some authentication is already explicitly
 	// configured directly in the config by a user.
 	// See: https://github.com/databricks/databricks-sdk-go/issues/304
-	if cfg.Profile == "" && (l.isAnyAuthConfigured(cfg) || cfg.IsAzure()) {
+	if cfg.Profile == "" && (l.isAnyAuthConfigured(cfg) || cfg.IsAws() || cfg.IsAzure() || cfg.IsGcp()) {
```
nfx, line 68: should we just check for host not empty?

nfx (changes requested):

config/config.go
mgyucht: Seems fine, one other way which is more of a whitelist approach is
```go
strings.Contains(c.Host, ".dev.databricks.com") || strings.Contains(c.Host, ".cloud.databricks.com")
```
staging is a subzone of prod.

config/config_file.go
alexott, line 68: but do we need to list all clouds here? It's equivalent of using any cloud

alexott: I'm also that it will solve this https://github.com/databricks/terraform-provider-databricks/issues/2149 - as we're using non-cloud URL in tests

config/config.go
pietern: Good point. Modified to check for host not empty in this function.

If the host is empty, `IsAws` should also return false because we don't know (different from current behavior).

config/config_file.go
pietern: Good point -- made the field checks explicit.

pietern: Done.

config/config.go
```diff
@@ -143,14 +143,14 @@ func (c *Config) IsAzure() bool {
...
+// IsGcp returns if the client is configured for Databricks on Google Cloud.
 func (c *Config) IsGcp() bool {
 	return strings.Contains(c.Host, ".gcp.databricks.com")
 }
 
-// IsAws returns true if client is configured for AWS
+// IsAws returns if the client is configured for Databricks on AWS.
 func (c *Config) IsAws() bool {
-	return !c.IsAzure() && !c.IsGcp()
+	return c.Host != "" && !c.IsAzure() && !c.IsGcp()
```
mgyucht, line 153: Allow-list version may also be safer:
```go
strings.Contains(c.Host, ".dev.databricks.com") || strings.Contains(c.Host, ".cloud.databricks.com")
```

mgyucht (approved):
LGTM

alexott (commented):
lgtm

alexott (approved):

config/config.go
pietern: True, but we would have this problem: https://github.com/databricks/databricks-sdk-go/pull/363#discussion_r1159502765

mgyucht: ack

nfx (approved):

PR #368, opened 2023-04-13
Cleanup ephemeral groups in integration tests

pietern:
## Changes

These tests left groups around that had to be removed manually/separately.

## Tests

Manually ran changed integration tests.

### DIFF
--- a/internal/permissions_test.go
+++ b/internal/permissions_test.go
@@ -34,6 +34,10 @@ func TestAccGenericPermissions(t *testing.T) {
 		DisplayName: RandomName("go-sdk-"),
 	})
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		err := w.Groups.DeleteById(ctx, group.Id)
+		require.NoError(t, err)
+	})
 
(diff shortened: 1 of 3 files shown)


mgyucht (approved):
LGTM! Is this a good use of the `defer` keyword in golang?

pietern:
@mgyucht It could be if tests didn't use `panic` for control flow (i.e. `t.FailNow()` or `t.Skip()`).

Defer works if you return from a function under normal circumstances, not when unwinding a `panic`.

pietern:
@mgyucht Turns out my understanding is wrong, defer should still work with `panic`.

See https://go.dev/blog/defer-panic-and-recover.

pietern:
@mgyucht The difference is that `t.Cleanup` is called on test completion and `defer` when returning from a function. So if you have a bunch of helper functions to construct state that you want to keep around for the duration of a test, `t.Cleanup` will do the trick, where `defer` wouldn't.

nfx (approved):

PR #378, opened 2023-04-25
Fix mlflow acceptance tests

kartikgupta-db:
## Changes
* Fix `TestAccClustersApiIntegration` acceptance test for mlflow on azure.

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/internal/mlflow_test.go
+++ b/internal/mlflow_test.go
@@ -132,12 +132,11 @@ func deleteModel(t *testing.T, w *databricks.WorkspaceClient, ctx context.Contex
 
 func TestAccRegistryWebhooks(t *testing.T) {
 	ctx, w := workspaceTest(t)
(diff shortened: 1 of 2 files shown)


internal/mlflow_test.go
```diff
@@ -132,7 +132,7 @@ func deleteModel(t *testing.T, w *databricks.WorkspaceClient, ctx context.Contex
 
 func TestAccRegistryWebhooks(t *testing.T) {
 	ctx, w := workspaceTest(t)
-
+	w.Config.FixHostIfNeeded()
```
nfx, line 135: i think it's better to add `CanonicalHostname()` method that will call `fixHostIfNeeded` internally, rather than flipping `fixHostIfNeeded` public. it's not intended for a public use.

nfx (changes requested):

PR #380, opened 2023-05-01
Fixed MLflow integration test and removed workaround for `DELETE` query parameter

tanmay-db:
## Changes

- Fix Mlflow integration test by checking status of model version and waiting for it to not be pending -- issue was happening due to race condition while deleting and status being pending leading to 409 status code.
- Use query parameters for delete
- Remove delete workaround

## Tests

Ran integration test on PR:

- azure-prod: https://github.com/databricks/eng-dev-ecosystem/actions/runs/4854588429
- aws-prod: https://github.com/databricks/eng-dev-ecosystem/actions/runs/4854587272

Verified query parameters are supported for MLflow DELETE

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/internal/mlflow_test.go
+++ b/internal/mlflow_test.go
@@ -1,10 +1,7 @@
 package internal
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
-	"net/http"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/mlflow_test.go
```diff
@@ -162,6 +137,35 @@ func TestAccRegistryWebhooks(t *testing.T) {
...
+		if currentModelVersion.ModelVersion.Status != ml.ModelVersionStatusPendingRegistration {
+			break
+		}
+		currentTime := time.Now()
+		elapsedTime := currentTime.Sub(startTime).Seconds()
+		if elapsedTime > 10 {
+			t.Fatalf("Timeout: Model version still has PENDING_REGISTERATION status after 10 seconds")
+		}
+		time.Sleep(2 * time.Second)
+	}
```
kartikgupta-db, line 159: You might be able to use something like this https://stackoverflow.com/a/52218264 here, instead of manually keeping track of time.

kartikgupta-db (changes requested):

kartikgupta-db (approved):

internal/mlflow_test.go
tanmay-db: discussed offline - tldr - not able to use testing methods through goroutine -- might require passing messages to channel and read them from the testMethod and call test fail (after checking the channel to see if message is returned) but maybe not worth it.

PR #381, opened 2023-05-04
Make clusters acceptance tests robust to duplicate cluster names

shreyas-goenka:
## Changes
The nightly clusters integration tests are failing because of duplicate cluster names in our testing workspaces. This PR removes a call to `ClusterInfoClusterNameToClusterIdMap` to make us robust against that 

## Tests
Manually by running the acceptance test

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/internal/clusters_test.go
+++ b/internal/clusters_test.go
@@ -209,15 +209,21 @@ func TestAccClustersApiIntegration(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, len(events) > 0)
 
+	// List clusters in workspace
 	all, err := w.Clusters.ListAll(ctx, compute.ListClustersRequest{})
 	require.NoError(t, err)
+	foundCluster := false
+	for _, info := range all {
+		if info.ClusterName == clusterName {
+			foundCluster = true
+		}
+	}
+	assert.True(t, foundCluster)
 
-	// List compute in workspace
-	names, err := w.Clusters.ClusterInfoClusterNameToClusterIdMap(ctx,
-		compute.ListClustersRequest{})
+	// Get cluster by name and assert it still exists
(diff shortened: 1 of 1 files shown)


internal/clusters_test.go
```diff
@@ -209,15 +209,21 @@ func TestAccClustersApiIntegration(t *testing.T) {
...
+	// List clusters in workspace
 	all, err := w.Clusters.ListAll(ctx, compute.ListClustersRequest{})
 	require.NoError(t, err)
+	foundCluster := false
+	for _, info := range all {
+		if info.ClusterName == clusterName {
+			foundCluster = true
+		}
+	}
+	assert.True(t, foundCluster)
```
tanmay-db, line 221: Maybe we could use: https://pkg.go.dev/golang.org/x/exp/slices#Contains and if not then let's make a generic utility so it can be used at other places: https://stackoverflow.com/a/70802740/16597007.

internal/clusters_test.go
```diff
@@ -209,15 +209,21 @@ func TestAccClustersApiIntegration(t *testing.T) {
...
 
-	// List compute in workspace
-	names, err := w.Clusters.ClusterInfoClusterNameToClusterIdMap(ctx,
-		compute.ListClustersRequest{})
```
tanmay-db, line 217: Do you happen to know why did it start failing recently and not before?

tanmay-db (commented):
Thanks for PR, can you run an integration test on the PR and see if it passes? Not sure if you already did since there is no link in description.

mgyucht (approved):

internal/clusters_test.go
mgyucht: probably because i created a luster with the same name as another cluster...

nfx (approved):

PR #383, opened 2023-05-09
Serialize params to request body on delete

mgyucht:
## Changes
This change was introduced in https://github.com/databricks/databricks-sdk-go/commit/0dc1944411cd611ef4a92bc6a10eac379500dc55#diff-bd3a55a72186f59e2e63efb4951573b2f9e4a7cc98086e922b0859f8ccc1dd09L324, but it causes metastore deletion to fail.

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- [x] TestUcAccMetastores, TestUcAccCatalogs, and TestUcAccSchemas all pass after this change.
- [x] TestUcAccMetastoreAssignment and TestUcAccCreateRecipientDb2DbAws from terraform-provider-databricks are passing after this change.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -324,6 +324,10 @@ func serializeQueryParamsToRequestBody(data any) ([]byte, error) {
 	m := make(map[string]any)
 	rv := reflect.ValueOf(data)
 	rt := reflect.TypeOf(data)
+	// If data is a map, just serialize it to JSON.
+	if rv.Kind() == reflect.Map {
+		return json.MarshalIndent(data, "", "  ")
+	}
 	for i := 0; i < rv.NumField(); i++ {
 		field := rv.Field(i)
 		tag := rt.Field(i).Tag.Get("url")
@@ -348,8 +352,8 @@ var requestInBodyOverrides = []struct {
 	urlRegexp  *regexp.Regexp
 	serializer serializer
 }{
-	{"DELETE", regexp.MustCompile("/api/2.1/unity-catalog/(metastores|catalogs)/[^/]+"), serializeQueryParamsToRequestBody},
-	{"DELETE", regexp.MustCompile("/api/2.1/unity-catalog/workspaces/[^/]+/(metastore|catalog)"), serializeQueryParamsToRequestBody},
+	{"DELETE", regexp.MustCompile("(/api/2.1)?/unity-catalog/(metastores|catalogs)/[^/]+"), serializeQueryParamsToRequestBody},
+	{"DELETE", regexp.MustCompile("(/api/2.1)?/unity-catalog/workspaces/[^/]+/(metastore|catalog)"), serializeQueryParamsToRequestBody},
 }
 
 func getRequestCustomSerializer(method string, requestURL *string) serializer {
--- a/client/client.go
+++ b/client/client.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"regexp"
 	"runtime"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
...
+var catalogWorkspacePath = regexp.MustCompile("/api/2.1/unity-catalog/workspaces/[^/]+/catalog")
+
+// List of exact request prefixes for which the request data should be serialized
+// into the request body instead of the query string.
+var deleteRequestInBodyOverrides = []struct {
+	requestMethod     string
+	requestRegexp     *regexp.Regexp
+	requestSerializer serializer
+}{
+	{"DELETE", metastorePath, serializeUnityCatalogDeleteRequest},
```
nfx, line 370: I’d in-line these regexes

client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
...
+
+var metastorePath = regexp.MustCompile("/api/2.1/unity-catalog/metastores/[^/]+")
+var metastoreWorkspacePath = regexp.MustCompile("/api/2.1/unity-catalog/workspaces/[^/]+/metastore")
+var catalogPath = regexp.MustCompile("/api/2.1/unity-catalog/catalogs/[^/]+")
+var catalogWorkspacePath = regexp.MustCompile("/api/2.1/unity-catalog/workspaces/[^/]+/catalog")
+
+// List of exact request prefixes for which the request data should be serialized
+// into the request body instead of the query string.
+var deleteRequestInBodyOverrides = []struct {
+	requestMethod     string
```
nfx, line 366: Nit: “request” prefix is very redundant here

client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
 	return "", fmt.Errorf("unsupported query string data: %#v", data)
 }
 
+// Remove all custom request serializer logic once APP-1331 is rolled out.
+type serializer func(interface{}) ([]byte, error)
+
+func serializeUnityCatalogDeleteRequest(data interface{}) ([]byte, error) {
+	m := make(map[string]interface{})
```
nfx, line 324:
```suggestion
	m := make(map[string]any{})
```

client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
...
+	// Reflectively look up the Force field of data, and if it is true, set the
+	// "force" field of m to true.
```
nfx, line 326: Why didn’t you use query.Values(data) and then converted to map[str]any and then json marshal it?

 It’s hard to understand what the code below is doing.

client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
...
+		for i := 0; i < v.NumField(); i++ {
+			setField(v, i, "Force", "force")
+			setField(v, i, "MetastoreId", "metastore_id")
+		}
+	}
+
+	return json.MarshalIndent(m, "", "  ")
+}
+
+var metastorePath = regexp.MustCompile("/api/2.1/unity-catalog/metastores/[^/]+")
```
nfx, line 358: Does TF use 2.1 for UC? Don’t remember from top of my head

client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
 	return "", fmt.Errorf("unsupported query string data: %#v", data)
 }
 
+// Remove all custom request serializer logic once APP-1331 is rolled out.
+type serializer func(interface{}) ([]byte, error)
```
nfx, line 321:
```suggestion
type serializer func(any) ([]byte, error)
```

client/client.go
```diff
@@ -316,11 +317,83 @@ func makeQueryString(data interface{}) (string, error) {
...
+	return nil
+}
+
 func makeRequestBody(method string, requestURL *string, data interface{}) ([]byte, error) {
 	var requestBody []byte
 	if data == nil && (method == "DELETE" || method == "GET") {
 		return requestBody, nil
 	}
+	if customSerializer := getRequestCustomSerializer(method, requestURL); customSerializer != nil {
+		return customSerializer(data)
```
nfx, line 395: There’s only one corner case we have now, no need to make it more complex

internal/catalog_test.go
```diff
@@ -162,16 +166,18 @@ func TestUcAccCatalogs(t *testing.T) {
...
 	})
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		// TODO: force delete
-		err := w.Catalogs.DeleteByName(ctx, newCatalog.Name)
+		deleteCatalogRequest := catalog.DeleteCatalogRequest{
+			Name:  newCatalog.Name,
+			Force: true,
+		}
+		err := w.Catalogs.Delete(ctx, deleteCatalogRequest)
```
nfx, line 180: In-line the request for readability

nfx (changes requested):

client/client.go
mgyucht: query.Values(data) converts values to `string`, but we want to serialize the request body using the underlying type (e.g. bool for "force").

client/client.go
mgyucht: I copied these from the autogenerated code, so I think they should be right. I think this should be covered by the integration test cases?

client/client.go
mgyucht: Do you think this is too complex?

mgyucht (commented):
Address comments

client/client.go
nfx: got it!

nfx: okay, we don't really need `setField` in the current shape - it's too complex. the better solution is something like:

```go
    rv := reflect.ValueOf(data)
    rt := reflect.TypeOf(data)
    for i := 0; i < rv.NumField(); i++ {
        field := rv.Field(i)
        tag := rt.Field(i).Tag.Get("query")
        if tag == "" {
            continue
        }
        value := field.Interface()
        if reflect.IsZero(value) {
           continue
        }
        key := strings.TrimSuffix(tag, ",omitempty") 
        m[key] = value
    }
```

client/client.go
```diff
@@ -316,11 +317,76 @@ func makeQueryString(data interface{}) (string, error) {
 	return "", fmt.Errorf("unsupported query string data: %#v", data)
 }
 
+// Remove all custom request serializer logic once APP-1331 is rolled out.
+type serializer func(any) ([]byte, error)
+
+func serializeUnityCatalogDeleteRequest(data any) ([]byte, error) {
```
nfx, line 323: i think it's better to remove `UnityCatalog` from the name of the function. we may have some of the non-compliant GET requests as well.

nfx (changes requested):

client/client.go
nfx: TF does not fully rely on autogenerated code yet. that's why we have Go SDK migration ;)

client/client.go
```diff
@@ -316,11 +317,66 @@ func makeQueryString(data interface{}) (string, error) {
...
 }
 
+// Remove all custom request serializer logic once APP-1331 is rolled out.
+type serializer func(any) ([]byte, error)
+
+func serializeQueryParamsToRequestBody(data any) ([]byte, error) {
+	m := make(map[string]any)
+	rv := reflect.ValueOf(data)
+	rt := reflect.TypeOf(data)
+	// If data is a map, just serialize it to JSON.
```
mgyucht, line 327: Note that the Terraform provider depends on this behavior because it passes a map as the request body. This should be forward-compatible with TF migration to Go SDK for the Metastore/Catalog resources, because they will use the struct type and go in the following pathway.

client/client.go
```diff
@@ -316,11 +317,66 @@ func makeQueryString(data interface{}) (string, error) {
...
+		return json.MarshalIndent(data, "", "  ")
+	}
+	for i := 0; i < rv.NumField(); i++ {
+		field := rv.Field(i)
+		tag := rt.Field(i).Tag.Get("url")
+		if tag == "" {
+			continue
+		}
+		value := field.Interface()
+		if reflect.DeepEqual(value, reflect.Zero(field.Type()).Interface()) {
```
nfx, line 338:
```suggestion
		if field.IsZero() {
```

no need for deep equal

client/client.go
```diff
@@ -316,11 +317,66 @@ func makeQueryString(data interface{}) (string, error) {
...
+			continue
+		}
+		key := strings.Split(tag, ",")[0]
```
nfx, line 341: there might be a segfault when there's no comma

client/client.go
```diff
@@ -316,11 +317,66 @@ func makeQueryString(data interface{}) (string, error) {
 	return "", fmt.Errorf("unsupported query string data: %#v", data)
 }
 
+// Remove all custom request serializer logic once APP-1331 is rolled out.
+type serializer func(any) ([]byte, error)
+
+func serializeQueryParamsToRequestBody(data any) ([]byte, error) {
+	m := make(map[string]any)
```
nfx, line 324:
```suggestion
	m := map[string]any{}
```

nfx (changes requested):

client/client.go
mgyucht: Doccomment just says it returns the original string in a slice of length 1.
> // If s does not contain sep and sep is not empty, Split returns a
> // slice of length 1 whose only element is s.

nfx (approved):

PR #394, opened 2023-05-22
Make u2m authentication work with new CLI

pietern:
## Changes

Acquire token by calling `databricks` CLI. Incorporate detection to see if the path resolves to the Python CLI (< 0.100) or the new CLI (>= 0.100). This detection looks at the binary size to determine which one we're working with.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


config/auth_databricks_cli.go
```diff
@@ -0,0 +1,119 @@
...
+	// Determine executable size as signal to determine old/new Databricks CLI.
+	stat, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// The new Databricks CLI is a single binary with size > 1MB.
+	// We use the size as a signal to determine which Databricks CLI is installed.
+	if stat.Size() < (1024 * 1024) {
+		return nil, errLegacyDatabricksCli
```
nfx, line 98: This might not work on windows, but we'll fix after if integration tests fail

nfx (approved):
lgtm

PR #401, opened 2023-05-24
Fix nondeterminism in workspace filesystem integration test

pietern:
## Changes

Broken by #389.

The test performs recursive listing and asserts the result is non-empty without first ensuring there is a file to list. This would sometimes pass and sometimes fail because tests are executed in parallel.

## Tests

Ran TestAccWorkspaceRecursiveListNoTranspile manually.

### DIFF
--- a/internal/workspace_test.go
+++ b/internal/workspace_test.go
@@ -76,6 +76,18 @@ func TestAccWorkspaceIntegration(t *testing.T) {
 
 func TestAccWorkspaceRecursiveListNoTranspile(t *testing.T) {
 	ctx, w := workspaceTest(t)
+	notebook := myNotebookPath(t, w)
+
+	// Import the test notebook
+	err := w.Workspace.Import(ctx, workspace.Import{
+		Path:      notebook,
+		Format:    "SOURCE",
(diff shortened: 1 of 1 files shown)


internal/workspace_test.go
```diff
@@ -76,6 +76,18 @@ func TestAccWorkspaceIntegration(t *testing.T) {
 
 func TestAccWorkspaceRecursiveListNoTranspile(t *testing.T) {
 	ctx, w := workspaceTest(t)
+	notebook := myNotebookPath(t, w)
+
+	// Import the test notebook
+	err := w.Workspace.Import(ctx, workspace.Import{
+		Path:      notebook,
+		Format:    "SOURCE",
+		Language:  "PYTHON",
```
andrewnester, line 85: There are constants for these values like `workspace.LanguagePython` and `workspace. ExportFormatSource `, I'd better use them

andrewnester (approved):

internal/workspace_test.go
pietern: Good point -- I copied this code from elsewhere in the file to unblock.

PR #403, opened 2023-05-25
Do not use cached entities in code generation

andrewnester:
## Changes
Code generator stores the defined types (entities) in `pkg.types` field. This field is primarily used for keeping the list of all observed types while generating code.

But if we use the entity stored in this map and modify it in-place, for example, add that some field is required, it will change the entity stored.

If later one we use this "cached" version of entity, it will contain modified configuration which can lead to unexpected results, for example, `User` entity in `Users.create` operation will have `ID` field as required even though based on OpenAPI definition it is not.

Thus, it's better not to use cached entities.

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/openapi/code/package.go
+++ b/openapi/code/package.go
@@ -253,11 +253,6 @@ func (pkg *Package) definedEntity(name string, s *openapi.Schema) *Entity {
 		return pkg.define(entity)
 	}
 
-	component := pkg.localComponent(&s.Node)
-	if s.IsRef() && pkg.types[component] != nil {
-		// entity is defined, return from cache
-		return pkg.types[component]
-	}
 	e := pkg.schemaToEntity(s, []string{name}, true)
 	if e == nil {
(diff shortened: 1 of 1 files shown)


pietern (approved):
For my info, where are these entities modified after being returned?

andrewnester:
@pietern as an example, for `Users.*` operation we were returning cached object here https://github.com/databricks/databricks-sdk-go/blob/main/openapi/code/service.go#L139

It happens that `Users.update` and such are processed before `Users.create` operation in code generator.

When we process `Users.update`, request object was modified here
https://github.com/databricks/databricks-sdk-go/blob/main/openapi/code/service.go#L172

Thus, when `Users.create` is processed and takes request schema from cache, it contains wrong required fields.

nfx:
this PR breaks SDK generation with path parameters

PR #408, opened 2023-06-01
Introduce waiters as top-level methods

nfx:
## Changes

* Deprecated `*AndWait` methods, will be removed in further releases. This PR keeps backwards compatibility for these methods.
* Introduced dedicated wait-for-state polling methods. This PR breaks compatibility for these methods and increases coherence with other SDKs
* `*API` clients with long-running semantics now always return an instance, that allows to optionally start polling

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/internal/jobs_test.go
+++ b/internal/jobs_test.go
@@ -89,13 +89,13 @@ func TestAccJobsApiFullIntegration(t *testing.T) {
 	require.NoError(t, err)
 
 	cancelledRun, err := w.Jobs.CancelRunAndWait(ctx, jobs.CancelRun{
-		RunId: runNowResponse.RunId,
+		RunId: runNowResponse.Response.RunId,
 	})
 	require.NoError(t, err)
 
 	repairedRun, err := w.Jobs.RepairRunAndWait(ctx, jobs.RepairRun{
 		RerunTasks: []string{cancelledRun.Tasks[0].TaskKey},
-		RunId:      runNowResponse.RunId,
+		RunId:      runNowResponse.Response.RunId,
 	})
 	require.NoError(t, err)
 	assert.GreaterOrEqual(t, len(repairedRun.Tasks), 1)
--- a/service/compute/api.go
+++ b/service/compute/api.go
@@ -233,6 +233,114 @@ func (a *ClustersAPI) Impl() ClustersService {
 	return a.impl
 }
 
+// WaitGetClusterRunning calls [ClustersAPI.Edit] and waits to reach RUNNING state
+func (a *ClustersAPI) WaitGetClusterRunning(ctx context.Context, clusterId string,
+	timeout time.Duration, callback func(*ClusterInfo)) (*ClusterInfo, error) {
+	ctx = useragent.InContext(ctx, "sdk-feature", "long-running")
+	return retries.Poll[ClusterInfo](ctx, timeout, func() (*ClusterInfo, *retries.Err) {
+		clusterInfo, err := a.Get(ctx, GetClusterRequest{
+			ClusterId: clusterId,
+		})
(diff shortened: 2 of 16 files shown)


service/compute/api.go
```diff
@@ -233,6 +233,114 @@ func (a *ClustersAPI) Impl() ClustersService {
 	return a.impl
 }
 
+// WaitGetClusterRunning calls [ClustersAPI.Edit] and waits to reach RUNNING state
```
pietern, line 236: It doesn't call `Edit`.

service/compute/api.go
```diff
@@ -233,6 +233,114 @@ func (a *ClustersAPI) Impl() ClustersService {
...
+func (a *ClustersAPI) WaitGetClusterRunning(ctx context.Context, clusterId string,
```
pietern, line 237: This name is a bit clunky. I would have expected `GetClusterWaitRunning`.

service/compute/api.go
```diff
@@ -386,50 +478,42 @@ func (a *ClustersAPI) DeleteByClusterIdAndWait(ctx context.Context, clusterId st
 // state will be rejected with an `INVALID_STATE` error code.
 //
 // Clusters created by the Databricks Jobs service cannot be edited.
-func (a *ClustersAPI) Edit(ctx context.Context, request EditCluster) error {
-	return a.impl.Edit(ctx, request)
+func (a *ClustersAPI) Edit(ctx context.Context, editCluster EditCluster) (*WaitGetClusterRunning[any], error) {
```
pietern, line 481: This was not a long running op and now potentially is. That's odd.

pietern (commented):
Need to have a closer look.

service/compute/api.go
nfx: @pietern it's a common name across all SDKs, all with common `Wait` prefix.

service/compute/api.go
nfx: it was

nfx: there was `EditAndWait` :) `Edit` does not trigger a wait anymore, because it just returns `WaitGetClusterRunning`, on which you have to call `.Get()`

internal/jobs_test.go
```diff
@@ -89,13 +89,13 @@ func TestAccJobsApiFullIntegration(t *testing.T) {
 	require.NoError(t, err)
 
 	cancelledRun, err := w.Jobs.CancelRunAndWait(ctx, jobs.CancelRun{
-		RunId: runNowResponse.RunId,
+		RunId: runNowResponse.Response.RunId,
```
pietern, line 92: Unlike for Python, we have the opportunity here to return 3 values: the response, the waiter, the error.

Then there's no need for calls like this.

service/compute/api.go
pietern: Alright, works.

service/compute/api.go
```diff
@@ -233,6 +233,114 @@ func (a *ClustersAPI) Impl() ClustersService {
...
+			return nil, retries.Halt(err)
+		default:
+			return nil, retries.Continues(statusMessage)
+		}
+	})
+}
+
+// WaitGetClusterRunning is a wrapper that calls [ClustersAPI.WaitGetClusterRunning] and waits to reach RUNNING state.
+type WaitGetClusterRunning[R any] struct {
+	Response *R
```
pietern, line 267: If the suggestion is applied then this no longer needs a type param or this response field.

internal/jobs_test.go
nfx: added direct fields for polling continuation across process boundary

PR #409, opened 2023-06-01
Revert "Do not use cached entities in code generation (#403)"

nfx:
This reverts commit 0af0ffcfe4e6ea977347ab107d16e5b2de9e7e41, otherwise, we miss the path parameters in the requests and break all SDKs
image


andrewnester:
@nfx could you point to where do we modify cached object which is required for correct generation of IntegrationId?

The fact that we have different behaviour for cached and non-cached object is a bit worrisome

PR #410, opened 2023-06-01
Improve command execution interface

nfx:
## Changes

- Create a more performant wrapper over command execution contexts
- Ensure that the supplied cluster is started and is properly running
- more conventional API flow for Go
- BC break for w.CommandExecution


service/compute/commands.go
```diff
@@ -9,6 +9,63 @@ import (
...
+	language     Language
+	clusterID    string
+	contextID    string
+}
+
+// Start the command execution context on a cluster and ensure it transitions to a running state
+func (a *CommandExecutionAPI) Start(ctx context.Context, clusterID string, language Language) (*CommandExecutorV2, error) {
+	executionImpl, ok := a.impl.(*commandExecutionImpl)
+	if !ok {
+		return nil, fmt.Errorf("unknown command executionimplementation: %v", a.impl)
```
pietern, line 24: Invalid error message.

service/compute/commands.go
```diff
@@ -9,6 +9,63 @@ import (
...
+	}
+	// re-initializing clusters API here, so that we have less IF's in the generator
```
pietern, line 26: I don't understand this comment. Why does the generator have anything to do with this?

service/compute/commands.go
```diff
@@ -9,6 +9,63 @@ import (
...
+func (c *CommandExecutorV2) Destroy(ctx context.Context) error {
+	return c.executionAPI.Destroy(ctx, DestroyContext{
+		ClusterId: c.clusterID,
+		ContextId: c.contextID,
+	})
+}
+
+// Execute runs given command in the running cluster and context and waits for the results
+func (c *CommandExecutorV2) Execute(ctx context.Context, command string) (*Results, error) {
+	status, err := c.executionAPI.ExecuteAndWait(ctx, Command{
```
pietern, line 57: These calls should set the `sdk-feature` like the previous ones.

service/compute/commands.go
```diff
@@ -9,6 +9,63 @@ import (
 	"github.com/databricks/databricks-sdk-go/useragent"
 )
 
+type CommandExecutorV2 struct {
+	clustersAPI  *ClustersAPI
+	executionAPI *CommandExecutionAPI
+	language     Language
+	clusterID    string
+	contextID    string
+}
```
pietern, line 18: Testing this would be painful, consider turning this into an interface like `CommandExecutor` and then provide an implementation instead. Then it'll be much easier to drop in a mock when testing downstream usage.

pietern (commented):
Approach LGTM, left a couple comments.

pietern (commented):
Approach LGTM, left a couple comments.

service/compute/commands.go
nfx: yep. that's the plan. TF has plenty of resources that call into this api

service/compute/commands.go
nfx: workspace client generation

service/compute/commands.go
nfx: `WaitCommandStatusCommandExecutionFinishedOrError` adds it already

PR #420, opened 2023-06-08
Use x-databricks-is-accounts flag to determine whether a service is an account level service

gauthamsunjay:
## Changes

We already mark services as account level services using a custom databricks flag `x-databricks-is-accounts`. This PR updates the codegen logic to rely on this flag to classify a service as an account level service instead of the existing regex match(path contains `/accounts/`).

This change is required because `Access Management` team has a workspace proxy for an account level API. 

Path of account level API: `/api/2.0/preview/accounts/{account_id}/access-control/...`
Path of workspace API that just proxies to the account level service: `/api/2.0/preview/accounts/access-control/...`

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


openapi/model.go
```diff
@@ -47,10 +47,11 @@ type Specification struct {
...
 type Tag struct {
 	Node
-	Package   string `json:"x-databricks-package"`
-	PathStyle string `json:"x-databricks-path-style"`
-	Service   string `json:"x-databricks-service"`
-	Name      string `json:"name"`
+	Package          string `json:"x-databricks-package"`
+	PathStyle        string `json:"x-databricks-path-style"`
+	Service          string `json:"x-databricks-service"`
+	IsAccountService bool   `json:"x-databricks-is-accounts"`
```
nfx, line 53:
```suggestion
	IsAccounts bool   `json:"x-databricks-is-accounts"`
```
for more consistency with the tags

nfx (commented):
LGTM on the generator side, but universe bug has to be fixed as well. Let's first fix it there and only then merge this PR

gauthamsunjay:
@nfx can we merge this PR now?

PR #422, opened 2023-06-08
Don't panic but return error from NewAccountClient

pietern:
## Changes

TSIA

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


nfx (approved):
lgtm

PR #423, opened 2023-06-09
Added `Upload` and `Download` methods where applicable

nfx:
## Changes
- adds `Upload` and `Download` methods for Workspace API
- adds `Upload` and `Download` methods for Files API
- removed generating `direct_download` parameter in downstream SDKs, as it makes return type non-deterministic

## Tests
Run new integrations tests


.DS_Store
pietern, line 1: .gitignore?

client/client.go
```diff
@@ -275,6 +284,7 @@ func (c *DatabricksClient) perform(
 	visitors = append([]func(*http.Request) error{
 		c.Config.Authenticate,
 		c.addHostToRequestUrl,
+		c.addApplicationJsonContentType,
```
pietern, line 287: I know this doesn't change behavior, but strictly speaking we only need to set this header on requests with a body. We can set the `Accept` header on all of them as it pertains to the response.

internal/workspace_test.go
```diff
@@ -74,18 +74,92 @@ func TestAccWorkspaceIntegration(t *testing.T) {
...
 }
 
+func TestAccWorkspaceUploadNotebookWithFileExtensionNoTranspile(t *testing.T) {
+	// TODO: remove NoTranspile suffix once other languages get Upload/Donwload features
+	ctx, w := workspaceTest(t)
+
+	notebookPath := filepath.Join("/Users", me(t, w).UserName, RandomName("notebook-")+".py")
+
+	err := w.Workspace.Upload(ctx, notebookPath, strings.NewReader("print(1)"))
+	assert.NoError(t, err)
```
pietern, line 84: Does this add the header magically? If so this API won't work for regular files.

service/files/files_api.go
```diff
@@ -0,0 +1,80 @@
...
+	return a.Upload(ctx, name, bytes.NewBuffer(data))
+}
+
+func (a *FilesAPI) Download(ctx context.Context, path string) (*bytes.Buffer, error) {
+	impl, ok := a.impl.(*filesImpl)
+	if !ok {
+		return nil, fmt.Errorf("wrong impl: %v", a.impl)
+	}
+	var buf bytes.Buffer
+	err := impl.client.Do(ctx, "GET", "/api/2.0/fs/files"+path, nil, &buf)
```
pietern, line 58: This still requires buffering of the whole response, which can be very large.

Especially if we have the `ReadFile` function we can make `Download` return an `io.Reader` and allow the caller to stream the response.

.DS_Store
nfx: gitignore...

client/client.go
nfx: we set the multipart header or octet-stream header in request visitors. multipart requests do require a Content-Type because of the multipart boundary.

internal/workspace_test.go
nfx: https://github.com/databricks/databricks-sdk-go/blob/upload-download/service/workspace/utilities.go#L171-L175

client/client.go
nfx: https://github.com/databricks/databricks-sdk-go/blob/upload-download/service/workspace/utilities.go#L171-L175

service/files/files_api.go
nfx: we'll fix it if it becomes a problem

.DS_Store
pietern, line 1: Please remove this.

internal/workspace_test.go
```diff
@@ -74,18 +74,92 @@ func TestAccWorkspaceIntegration(t *testing.T) {
 	assert.Contains(t, paths, notebook)
 }
 
+func TestAccWorkspaceUploadNotebookWithFileExtensionNoTranspile(t *testing.T) {
+	// TODO: remove NoTranspile suffix once other languages get Upload/Donwload features
```
pietern, line 78: Typo

internal/workspace_test.go
pietern: I mean the language header `# Databricks notebook source`.

service/files/files_api.go
pietern: Ack, works for me.

service/workspace/utilities.go
```diff
@@ -70,3 +73,154 @@ func (a *WorkspaceAPI) RecursiveList(ctx context.Context, path string) ([]Object
...
+//
+// Errors:
+//
+//   - RESOURCE_ALREADY_EXISTS: if `path` already exists no `overwrite=True`.
+//   - INVALID_PARAMETER_VALUE: if `format` and `content` values are not compatible.
+//
+// By default, workspace.UploadFormat(workspace.ExportFormatSource). If using
+// workspace.UploadFormat(workspace.ExportFormatAuto) the `path` is imported or
+// exported as either a workspace file or a notebook, depending on an analysis
+// of the `path`’s extension and the header content provided in the request.
```
pietern, line 108: Are you talking about the header of the file or the request header?

service/workspace/utilities.go
```diff
@@ -70,3 +73,154 @@ func (a *WorkspaceAPI) RecursiveList(ctx context.Context, path string) ([]Object
...
+}
+
+// Download a notebook or file from the workspace by path.
+//
+// By default, it acts as if workspace.DownloadFormat(workspace.ExportFormatSource) option is supplied. When using
+// workspace.ExportFormatAuto, the `path` is imported or exported as either a workspace file or a notebook, depending
+// on an analysis of the `item`’s extension and the header content provided in the request.
+//
+// Returns [bytes.Buffer] of the path contents.
+func (a *WorkspaceAPI) Download(ctx context.Context, path string, opts ...DownloadOption) (*bytes.Buffer, error) {
```
pietern, line 202: Please make this return an `io.ReadCloser` as well for symmetry with the Files API.

internal/workspace_test.go
nfx: where?..

internal/workspace_test.go
nfx: it's the backend that adds it

internal/workspace_test.go
pietern: Donwload

internal/workspace_test.go
pietern: Thanks -- now see that specifying the format does it, or having the file content header + format auto.

service/workspace/utilities.go
```diff
@@ -70,3 +73,154 @@ func (a *WorkspaceAPI) RecursiveList(ctx context.Context, path string) ([]Object
...
+}
+
+// Download a notebook or file from the workspace by path.
+//
+// By default, it acts as if workspace.DownloadFormat(workspace.ExportFormatSource) option is supplied. When using
+// workspace.ExportFormatAuto, the `path` is imported or exported as either a workspace file or a notebook, depending
+// on an analysis of the `item`’s extension and the file content header provided in the request.
+//
+// Returns [bytes.Buffer] of the path contents.
+func (a *WorkspaceAPI) Download(ctx context.Context, path string, opts ...DownloadOption) (io.Reader, error) {
```
pietern, line 202: Please make this a io.ReadCloser like the Files API, even if it's a nop closer. It's needed if the downstream users ever aborts reading a stream.

service/workspace/utilities.go
```diff
@@ -70,3 +73,154 @@ func (a *WorkspaceAPI) RecursiveList(ctx context.Context, path string) ([]Object
...
+	err := impl.client.Do(ctx, "GET", "/api/2.0/workspace/export", query, &buf)
+	if err != nil {
+		return nil, err
+	}
+	return &buf, nil
+}
+
+// ReadFile is identical to [os.ReadFile] but for workspace files.
+func (a *WorkspaceAPI) ReadFile(ctx context.Context, name string) ([]byte, error) {
+	b, err := a.Download(ctx, name, DownloadFormat(ExportFormatAuto))
```
pietern, line 221: Then this one should do a pro-forma `defer close(b)`.

pietern (approved):

service/workspace/utilities.go
nfx: these files are capped to 10Mb. why do we need a stream? :)

PR #426, opened 2023-06-11
Added log level support to SimpleLogger

marwansalem:
Implement Log levels.
Can be used as the default logger.
Default log level can be later specified via configuration, or environment variable.
I had been using the library and found out that I needed to define a custom logger to suppress logs, to be specific
I was using the `dbfsAPI` to upload files, and I was getting bombarded with logs with each `add_block` request.

## Changes

Add new Implementation for the `Logger` interface which respects the log level.
## Tests

added 2 new tests found in [level
-logger_test.go](logger/level_logger_test.go)

- [x] `make test` passing
- [x] `make fmt` applied


logger/level_logger.go
```diff
@@ -0,0 +1,14 @@
+package logger
+
+import (
+	"context"
+)
+
+type LevelLogger struct {
+	SimpleLogger
```
nfx, line 8: just add this to `SimpleLogger` - we expect downstream dependencies to implement their own logger, but it's fine to add level understanding to `SimpleLogger`

nfx (changes requested):

logger/level_logger.go
marwansalem: Please review, but now the default level should be Info, since it maps to int 0

logger/simple.go
```diff
@@ -5,10 +5,12 @@ import (
...
 
-type SimpleLogger struct{}
+type SimpleLogger struct {
+	Level Level
+}
 
 func (l *SimpleLogger) Enabled(_ context.Context, level Level) bool {
-	return true
+	return level >= l.Level
 }
```
nfx, line 14: please add `Enabled(..)` calls to every method in the simple logger. otherwise, this has no effect :) 

e.g. if we call `logger.Debugf(..)` and we have `logging.LevelInfo` set, it shouldn't print anything

nfx (changes requested):

logger/simple.go
marwansalem: Please check 👍

logger/simple.go
```diff
@@ -5,28 +5,45 @@ import (
...
 
-func (l *SimpleLogger) Tracef(_ context.Context, format string, v ...any) {
+func (l *SimpleLogger) Tracef(ctx context.Context, format string, v ...any) {
+	if !l.Enabled(ctx, l.Level) {
```
nfx, line 17:
```suggestion
	if !l.Enabled(ctx, l.LevelTrace) {
```

if not changed, it'll always return true ;) 

change it in other places as well

nfx (changes requested):

logger/simple.go
marwansalem: The last two comments including this one are valid, but do you think we can add tests to validate them?

PR #497, opened 2023-06-13
Add preview level functions for services and methods to OpenAPI code

pietern:
## Changes

TSIA

## Tests

Regenerated code using these helpers and observed the intended changes.


pietern:
Note: the tag is passed by value now because it was a pointer to a variable reused in a for loop, thus invalid after use.

nfx (commented):
can you also add these two methods on Entity and Field?

nfx (approved):

PR #501, opened 2023-06-14
Migrate Metadata Service Auth Tests to auth_permutation_test.go Framework

mgyucht:
## Changes
This PR migrates tests covering the metadata service to auth_permutations_test.go. This ensures that the test configuration is dumped for these tests and can be used to autogenerate test cases for auth in other SDKs.

One noteworthy point: this PR adds a `Client` field to `config.Config`. This is exposed for testing and allows us to mock the responses to HTTP requests made during authentication.

This PR adds the following fields to configFixture:
- `AssertHeaders`: A set of headers expected to be set on the `http.Request`. This allows us to verify that the correct `Authorization` header (and potentially other headers, depending on auth type) are set.
- `HTTPFixtures`: when set, the test framework will start a local webserver and pass the client into the `Config` struct. The client will respond using the provided API mocks.
- `MetadataServiceURL`: sets the corresponding field on Config. If HTTPFixtures is supplied, the URL authority is replaced by that of the local server.

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


config/auth_azure_cli_test.go
```diff
@@ -1,4 +1,4 @@
-package config
+package config_test
```
nfx, line 1: this will lead to code coverage not showing up for this package. check `make coverage`, please move tests back to `config` package

config/auth_permutations_test.go
```diff
@@ -473,3 +497,62 @@ func TestConfig_AuthTypeFromEnv(t *testing.T) {
...
+}
+
+func TestConfig_MetadataServerConfigure(t *testing.T) {
+	configFixture{
+		AuthType:           "metadata-service",
+		MetadataServiceURL: "http://localhost/token",
+		Host:               "x",
+		AssertHost:         "https://x",
+		AssertAuth:         "metadata-service",
+		AssertHeaders: map[string][]string{
```
nfx, line 526:
```suggestion
		AssertHeaders: map[string]string{
```
this way it's easier to port across languages

config/auth_permutations_test.go
```diff
@@ -473,3 +497,62 @@ func TestConfig_AuthTypeFromEnv(t *testing.T) {
...
+			"Authorization": {"Bearer XXX-2"}, // XXX-1 is returned during Configure()
+		},
+		HTTPFixtures: []qa.HTTPFixture{
+			{
+				Method:   "GET",
+				Resource: "/token",
+				ExpectedHeaders: map[string][]string{
```
nfx, line 533:
```suggestion
				ExpectedHeaders: map[string]string{
```

this way it's easier to reason across languages

config/auth_permutations_test.go
```diff
@@ -473,3 +497,62 @@ func TestConfig_AuthTypeFromEnv(t *testing.T) {
...
+					"X-Databricks-Metadata-Version": {"1"},
+					"X-Databricks-Host":             {"https://x"},
+				},
+				Response: config.AzureMsiToken{
+					AccessToken: "XXX-1",
+					ExpiresOn:   json.Number(strconv.FormatInt(time.Now().Add(1*time.Second).Unix(), 10)),
```
nfx, line 539: this is not really portable across languages with `deco code authtests`. I think we need to add a template variable somewhere

nfx (changes requested):

config/auth_permutations_test.go
mgyucht: This is the underlying type of `http.Header`. Other languages also follow this same pattern of headers being a Map[String, Iterable[String]] (https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletRequest.html#getHeaders-java.lang.String-, https://pytest-httpserver.readthedocs.io/en/latest/api.html#pytest_httpserver.RequestHandler.respond_with_data), so I was mainly following this standard.

config/auth_permutations_test.go
mgyucht: Good point. I believe the actual time is not necessary for these tests, will update these.

config/auth_azure_cli_test.go
mgyucht: make coverage shows that config package is covered by unit tests, including auth_metadata_server.
Screenshot 2023-06-15 at 10 51 05
Screenshot 2023-06-15 at 10 50 56

mgyucht:
Closing this out as won't do.

PR #524, opened 2023-06-20
Added ClusterID field for Config

andrewnester:
## Changes
As per https://docs.databricks.com/dev-tools/databricks-connect.html it's possible to provide cluster ID in Databricks configuration

This PR adds this field in Config struct

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- [ ] ~relevant integration tests applied~

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -38,6 +38,8 @@ type Config struct {
 	// Databricks host (either of workspace endpoint or Accounts API endpoint)
 	Host string `name:"host" env:"DATABRICKS_HOST"`
 
+	ClusterId string `name:"cluster_id" env:"DATABRICKS_CLUSTER_ID"`
+
 	// URL of the metadata service that provides authentication credentials.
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -38,6 +38,8 @@ type Config struct {
 	// Databricks host (either of workspace endpoint or Accounts API endpoint)
 	Host string `name:"host" env:"DATABRICKS_HOST"`
 
+	ClusterId string `name:"cluster_id" env:"DATABRICKS_CLUSTER_ID"`
```
pietern, line 41: Please use `ClusterID` to match the style of the other fields.

Also see: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

We can't enforce this everywhere, e.g. for generated code, but should try to stick to it where it makes sense.

pietern (approved):
The Python SDK also contains a warehouse ID, can you add that here as well?

https://github.com/databricks/databricks-sdk-py/blob/b59fcdf3493d2832ddb6335da1bf6da4a1847c70/databricks/sdk/core.py#L482-L483

PR #530, opened 2023-06-27
Support accounts-dod as account endpoint

mgyucht:
## Changes
Modify IsAccountClient to return `true` for new accounts endpoints that are set up for the Department of Defense.

Resolves https://github.com/databricks/terraform-provider-databricks/issues/2433 once TF provider is updated to a version including this commit.

## Tests
- [x] Added a unit test to cover IsAccountClient().
- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -158,7 +158,16 @@ func (c *Config) IsAws() bool {
 
 // IsAccountClient returns true if client is configured for Accounts API
 func (c *Config) IsAccountClient() bool {
-	return (c.AccountID != "" && c.isTesting) || strings.HasPrefix(c.Host, "https://accounts.")
+	accountsPrefixes := []string{
+		"https://accounts.",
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -158,7 +158,16 @@ func (c *Config) IsAws() bool {
...
+	accountsPrefixes := []string{
+		"https://accounts.",
+		"https://accounts-dod.",
+	}
+	for _, prefix := range accountsPrefixes {
+		if strings.HasPrefix(c.Host, prefix) {
+			return true
+		}
+	}
+	return c.AccountID != "" && c.isTesting
```
shreyas-goenka, line 170: Can you make this case return early since this is a special condition?
```
if c.AccountID != "" && c.isTesting {
  return true
}
```

shreyas-goenka (approved):

PR #534, opened 2023-06-28
Regenerate SDK according to OpenAPI spec c411c04c62262f1bdacc047356f1d2030f41e660

mgyucht:
## Changes
Regenerates the OpenAPI Spec at commit c411c04c62262f1bdacc047356f1d2030f41e660.

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- [x] terraform provider unit tests pass with the new SDK

### DIFF
--- a/service/catalog/impl.go
+++ b/service/catalog/impl.go
@@ -484,7 +484,7 @@ func (a *systemSchemasImpl) Disable(ctx context.Context, request DisableRequest)
 
 func (a *systemSchemasImpl) Enable(ctx context.Context, request EnableRequest) error {
 	path := fmt.Sprintf("/api/2.1/unity-catalog/metastores/%v/systemschemas/%v", request.MetastoreId, request.SchemaName)
-	err := a.client.Do(ctx, http.MethodPost, path, request, nil)
+	err := a.client.Do(ctx, http.MethodPut, path, request, nil)
 	return err
 }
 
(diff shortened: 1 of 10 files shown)


service/catalog/impl.go
```diff
@@ -484,7 +484,7 @@ func (a *systemSchemasImpl) Disable(ctx context.Context, request DisableRequest)
 
 func (a *systemSchemasImpl) Enable(ctx context.Context, request EnableRequest) error {
 	path := fmt.Sprintf("/api/2.1/unity-catalog/metastores/%v/systemschemas/%v", request.MetastoreId, request.SchemaName)
-	err := a.client.Do(ctx, http.MethodPost, path, request, nil)
+	err := a.client.Do(ctx, http.MethodPut, path, request, nil)
```
tanmay-db, line 487: What is the context behind Post -> Put?

tanmay-db (approved):
lgtm, have some questions, mentioned them.

service/catalog/impl.go
mgyucht: I checked this, and it seems like the wrong HTTP method was used originally but was recently corrected in the OpenAPI spec.

tanmay-db: I see

service/ml/model_registry_usage_test.go
```diff
@@ -58,20 +58,20 @@ func ExampleModelRegistryAPI_CreateComment_modelVersionComments() {
 
 }
 
-func ExampleModelRegistryAPI_CreateModel_models() {
+func ExampleModelRegistryAPI_CreateModel_modelVersionComments() {
```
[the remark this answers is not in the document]
mgyucht: I think this is just because the examples are not sorted. We should fix this in the code generator for tests.

PR #536, opened 2023-06-29
Added HasRequiredNonBodyField method

andrewnester:
## Changes
Added HasRequiredNonBodyField method

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/openapi/code/entity.go
+++ b/openapi/code/entity.go
@@ -256,6 +256,15 @@ func (e *Entity) IsAllRequiredFieldsPrimitive() bool {
 	return true
 }
 
+func (e *Entity) IsAllRequiredFieldsJsonUnserialisable() bool {
+	for _, v := range e.RequiredFields() {
+		if !v.IsJson {
+			return false
+		}
+	}
+	return true
+}
+
 // IsPrivatePreview flags object being in private preview.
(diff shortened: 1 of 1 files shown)


openapi/code/entity.go
```diff
@@ -256,6 +256,15 @@ func (e *Entity) IsAllRequiredFieldsPrimitive() bool {
 	return true
 }
 
+func (e *Entity) IsAllRequiredFieldsJsonUnserialisable() bool {
```
mgyucht, line 259: Shouldn't this be IsAllRequiredFieldsJsonSerializable? It returns true if for all required fields, IsJson is true.

Suggest adding a doccomment.

mgyucht (commented):
I was so confused by this PR until I realized that the name is probably not correct (I hope?). Should we rename this method?

openapi/code/entity.go
pietern, line 259: If I understand this correctly, the check returns if any of the required fields are not part of the request body (i.e. in the URL path or query). If so, then the name here should reflect that. E.g. HasRequiredNonBodyField.

pietern (changes requested):

PR #540, opened 2023-06-30
Add SCIM Patch Acceptance Test

mgyucht:
## Changes
Adds an acceptance test covering SCIM user patch that users can leverage when interacting with the SCIM API.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


examples/scim-patch/main.go
```diff
@@ -0,0 +1,127 @@
+package main
```
nfx, line 1: Use godoc examples to make it discoverable https://go.dev/blog/examples

make doc to preview

examples/scim-patch/main.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+	"os"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/logger"
+	"github.com/databricks/databricks-sdk-go/service/iam"
+)
+
+func fetchAndPrintUserName(c *databricks.AccountClient, id string) error {
+	user, err := c.Users.GetById(context.Background(), id)
```
nfx, line 20: Ctx must always be the first parameters and initialized as much at the top of business logic as possible. Here - in main()

examples/scim-patch/main.go
```diff
@@ -0,0 +1,127 @@
...
+		os.Exit(1)
+	}
+	name := os.Args[1]
+
+	// Login via databricks auth login https://accounts.cloud.databricks.com --account-id ...
+
+	// Create a client
+	cfg := databricks.Config{
+		Host: "https://accounts.cloud.databricks.com",
+		// Specify Account ID via the DATABRICKS_ACCOUNT_ID environment variable
```
nfx, line 42: Use input util from other examples

nfx (changes requested):

nfx (changes requested):

nfx (changes requested):
regen comments

PR #550, opened 2023-07-12
Parse Camel Case and Pascal Case Enum Values

mgyucht:
## Changes
Currently, enum values that use camelCase or PascalCase are not split into separate words during the parse phase of the OpenAPI spec. As a result, `PascalCase()` returns `Camelcase` or `Pascalcase`, respectively. To fix this, I've modified SplitASCII to support parsing more forms (snake-case and constant-case), to split words based on all non-alphanumeric symbols, and to estimate the "case" of a number by looking at the neighboring letters.

## Tests
- [x] Added several cases to SplitASCII unit tests.
- [x] Regenerated the Go, Python, and Java SDKs with this change, and the results are positive. https://github.com/databricks/databricks-sdk-py/pull/229 and https://github.com/databricks/databricks-sdk-java/pull/115

### DIFF
--- a/openapi/code/named.go
+++ b/openapi/code/named.go
@@ -99,13 +99,42 @@ func (n *Named) splitASCII() (w []string) {
 	return SplitASCII(n.Name)
 }
 
+// Return the value of cond evaluated at the nearest letter to index i in name.
+// dir determines the direction of search: if true, search forwards, if false,
+// search backwards.
+func search(name string, cond func(rune) bool, dir bool, i int) bool {
+	nameLen := len(name)
+	incr := 1
+	if !dir {
+		incr = -1
+	}
+	for j := i; j >= 0 && j < nameLen; j += incr {
+		if unicode.IsLetter(rune(name[j])) {
+			return cond(rune(name[j]))
+		}
+	}
+	return false
+}
+
+// Return the value of cond evaluated on the rune at index i in name. If that
+// rune is not a letter, search in both directions for the nearest letter and
+// return the result of cond on those letters.
+func checkCondAtNearestLetters(name string, cond func(rune) bool, i int) bool {
+	r := rune(name[i])
+
+	if unicode.IsLetter(r) {
+		return cond(r)
+	}
+	return search(name, cond, true, i) && search(name, cond, false, i)
+}
+
 // emulate positive lookaheads from JVM regex:
 // (?<=[a-z])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])|([-_\s])
 // and convert all words to lower case
 func SplitASCII(name string) (w []string) {
 	var current []rune
 	nameLen := len(name)
-	var last, this, next, lookahead bool
+	var isPrevUpper, isCurrentUpper, isNextLower, isNextUpper, isNotLastChar bool
 	// we do assume here that all named entities are strictly ASCII
 	for i := 0; i < nameLen; i++ {
 		r := rune(name[i])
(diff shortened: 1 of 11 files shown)


openapi/code/named.go
```diff
@@ -95,41 +95,80 @@ func (n *Named) Singular() *Named {
...
+		return cond(r)
+	}
+	return search(name, cond, true, i) && search(name, cond, false, i)
+}
+
 // emulate positive lookaheads from JVM regex:
 // (?<=[a-z])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])|([-_\s])
 // and convert all words to lower case
-func (n *Named) splitASCII() (w []string) {
+func SplitASCII(name string) (w []string) {
```
nfx, line 134: can you revert it to being a method on Named? it's not technically needed, as what is being split gets fed into Named anyway.
image

openapi/code/named.go
```diff
@@ -95,41 +95,80 @@ func (n *Named) Singular() *Named {
...
 }
 
+func (n *Named) splitASCII() (w []string) {
+	return SplitASCII(n.Name)
+}
+
+// Return the value of cond evaluated at the nearest letter to index i in name.
+// dir determines the direction of search: if true, search forwards, if false,
+// search backwards.
+func search(name string, cond func(rune) bool, dir bool, i int) bool {
```
nfx, line 105: better make these two functions as private methods to Named, not to pollute the global namespace

pietern (approved):
Nice work!

This reminds me that we still split `OAuth` into `["o", "auth"]`.

We probably need allow listing to make that work.

service/iam/model.go
```diff
@@ -478,7 +478,7 @@ func (f *PatchOp) Type() string {
 
 type PatchSchema string
 
-const PatchSchemaUrnIetfParamsScimApiMessagesPatchop PatchSchema = `urn:ietf:params:scim:api:messages:2.0:PatchOp`
+const PatchSchemaUrnIetfParamsScimApiMessages20PatchOp PatchSchema = `urn:ietf:params:scim:api:messages:2.0:PatchOp`
```
nfx, line 481: single-value enums hint at patch operations for SCIM having to be manually coded

nfx (changes requested):
major comments:
- `splitASCII` should remain a private method of Named. 
- there needs to be a change in Package.makeEnum

perhaps, changes to splitascii might be smaller

service/iam/model.go
mgyucht: +1. Until this is supported at the OpenAPI level, is there a practical approach we can take?

openapi/code/named_test.go
```diff
@@ -102,8 +102,12 @@ func TestNamedDecamel(t *testing.T) {
 		"HTMLParser":    {"html", "parser"},
 		"BigO":          {"big", "o"},
 		"OCaml":         {"o", "caml"},
+		"K8S_FAILURE":   {"k8s", "failure"},
+		"k8sFailure":    {"k8s", "failure"},
+		"i18nFailure":   {"i18n", "failure"},
+		"Patch:Request": {"patch", "request"},
```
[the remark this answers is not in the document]
mgyucht: SCIM is such a challenging API, but these fields are needed according to the spec.

openapi/code/named.go
mgyucht: Good point on the namespacing!

If I revert to the old implementation, splitASCII fails on `K8S_FAILURE` and `Patch:Request`, producing `k8, s, failure` and `patch:, request` respectively

service/iam/model.go
mgyucht: I see that there is an x-databricks-alias-enum field which we can use to simplify this name significantly, I'll do that in a separate PR since that requires an OpenAPI change in universe.

nfx: @mgyucht these enums won't be present in Java SDK - it won't allow non-alphanum values. 

one of the options is an annotation for a field to always be present in the code. still - it's only for SCIM `patch` requests. i think it's best to hand-code all patch requests...

nfx (approved):

service/iam/model.go
mgyucht: Yeah, we'll have to figure out how best to expose this in the API. It's a very difficult protocol to generate an SDK for... I used jackson's at-JsonProperty annotation to control serde for these fields, so I think it should be OK in Java.

nfx: but java enums don't support non-alphanum characters.

mgyucht: You mean the enum value names? They just need to be valid identifiers IIUC, which are alphanum + underscore and dollar sign. See https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-Identifier

> The "Java letters" include uppercase and lowercase ASCII Latin letters A-Z (\u0041-\u005a), and a-z (\u0061-\u007a), and, for historical reasons, the ASCII underscore (_, or \u005f) and dollar sign ($, or \u0024). The $ sign should be used only in mechanically generated source code or, rarely, to access pre-existing names on legacy systems.

I don't see any errors from IntelliJ with this change, and there are examples where underscores are allowed in enum values: https://stackoverflow.com/questions/9712977/is-it-possible-to-have-an-enum-class-with-enums-of-two-or-more-words.

Issue #552, opened 2023-07-14
[FEATURE] Detect workspace/account config upon construction of a client


pietern:
**Problem Statement**
A user can initialize a workspace client with an account profile and not get an error until they make an API call.

**Proposed Solution**
We can check `cfg.IsAccountClient()` to confirm whether or not the configuration is for a workspace or account.

nfx:
here is a whole class of issues related to `lazy auth` that we used to have in TF provider, so this has to be carefully designed https://github.com/databricks/terraform-provider-databricks/issues?q=is%3Aissue+host+label%3A%22lazy+auth%22

pietern:
Implemented in #596.

PR #553, opened 2023-07-17
[FEATURE] Detect workspace/account config upon construction of a client

ShripadMhetre:
Issue fixed - https://github.com/databricks/databricks-sdk-go/issues/552
## Changes
In `workspace_client.go`, added the conditional check to catch if account profile is being passed to instantiate WorkspaceClient. 

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- [] relevant integration tests applied


workspace_client.go
```diff
@@ -776,6 +778,8 @@ type WorkspaceClient struct {
 	WorkspaceConf *settings.WorkspaceConfAPI
 }
 
+var ErrNotWorkspaceClient = errors.New("invalid Databricks Workspace configuration")
```
nfx, line 781: please also modify the relevant files in `.codegen` folder

nfx (changes requested):

workspace_client.go
```diff
@@ -791,6 +795,11 @@ func NewWorkspaceClient(c ...*Config) (*WorkspaceClient, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if cfg.AccountID == "" || cfg.IsAccountClient() {
```
nfx, line 799: this will not work with Terraform, as sometimes we know host only after upstream modules are initialized.

nfx (changes requested):
this will not work with Terraform, as sometimes we know host only after upstream modules are initialized.

we can't deterministically give this kind of error on the level of NewWorkspaceClient

workspace_client.go
ShripadMhetre: Similar kind of check exist in `account_client.go` -

```
if cfg.AccountID == "" || !cfg.IsAccountClient() {
	return nil, ErrNotAccountClient
}
```

Can you help me understand the optimal solution then? to make it work with Terraform as well.

nfx: terraform resolves config upon the first request

ShripadMhetre: Please review the latest changes and help with proper feedback

.codegen/workspaces.go.tmpl
```diff
@@ -30,6 +32,10 @@ func NewWorkspaceClient(c ...*Config) (*WorkspaceClient, error) {
 		// default config
 		cfg = &config.Config{}
 	}
+	if cfg.AccountID == "" || cfg.IsAccountClient() {
```
nfx, line 35: it is not enough, proper implementation should be lazy evaluated - e.g. after the error. otherwise this will break certain terraform workflows, where account id is not yet available and is loaded from another resource

nfx (changes requested):

PR #559, opened 2023-07-25
Handle nested query parameters in Client.attempt

mgyucht:
## Changes
The Query History list API filter_by query parameter is modeled by a dictionary, rather than a primitive type, defining the allowed filters including query_start_time_range, statuses, user_ids, and warehouse_ids. To be compatible with gRPC transcoding, query parameters modeled by message types as opposed to primitives need to be separated into one query parameter per nested field, where the key is the path to that query parameter. For example:

```
ListQueryHistoryRequest{
  FilterBy: QueryFilter{
    UserIds: []int{123, 456}
  }
}
```
becomes
```
filter_by.user_ids=123&filter_by.user_ids=456
```
go-querystring already flattens this but to a different form:
```
filter_by[user_ids]=123&filter_by[user_ids]=456
```
To fix this, we replace `[` with `.` and `]` with empty string.

This resolves one of the problems from https://github.com/databricks/databricks-sdk-py/issues/99. The issue with the conflict between filter_by and next_page_token will be resolved by the backend service.

Separately, there is an underlying issue affecting the Go SDK specifically, as we delegate query parameter serialization to the go-querystring library. This library looks at the `url` struct tag to name the query parameters appropriately. However, schemas defined in `Components` in OpenAPI are initially cached with IsQuery for all fields = false. As we parse operations, we incrementally discover which entities are present in query parameters, and we need to update the underlying types recursively.

Lastly, this includes a bugfix to support the newest version of Delve.

## Tests
Added an integration test for ListQueryHistory.

- [x] `make test` passing
- [x] `make fmt` applied


internal/init_test.go
```diff
@@ -154,7 +154,7 @@ func skipf(t *testing.T) func(format string, args ...any) {
 // detects if test is run from "debug test" feature in VSCode
 func isInDebug() bool {
 	ex, _ := os.Executable()
-	return path.Base(ex) == "__debug_bin"
+	return strings.HasPrefix(path.Base(ex), "__debug_bin")
```
mgyucht, line 157: Small bugfix to support newest version of Delve (see https://github.com/databricks/terraform-provider-databricks/pull/2497)

PR #562, opened 2023-07-26
Support x-databricks-path-style overrides at the operation level

mgyucht:
## Changes
In order to support multiple styles of request under a single service, we introduce support for overriding x-databricks-path-style at the Operation level. This will allow RPC- and REST-style operations to exist under a single service. To determine the path style for a method, we check for a method-level override, falling back to the service-level setting. If neither is set, the path style is assumed to be "rest".

## Tests
Tested by regenerating the Python SDK using a spec I edited by hand to include this extension at the operation level. The generated code was as I expected (using the operation ID to derive name in the Clusters service, which is normally an RPC style service).

- [ ] `make test` passing
- [ ] `make fmt` applied


nfx (commented):
Where is it needed? Is it for Terminate/PermanentDelete? Can’t see this reflecting in Go SDK

mgyucht:
It's needed for the permissions refactor being done by @Rchakmak. His permissions APIs are all RESTful, but they will be part of services that are marked as RPC style.

pietern (approved):

PR #567, opened 2023-08-01
Update OpenAPI spec to include Permissions OpenAPI Refactor

mgyucht:
## Changes
Update the Go SDK to include the updated structures for the permissions APIs.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


examples/test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/databricks/databricks-sdk-go"
+	"github.com/databricks/databricks-sdk-go/service/files"
+)
+
+func main() {
```
nfx, line 11: this file doesn't belong here

service/iam/api.go
```diff
@@ -790,7 +790,51 @@ func NewPermissions(client *client.DatabricksClient) *PermissionsAPI {
 }
 
 // Permissions API are used to create read, write, edit, update and manage
-// access for various users on different objects and endpoints.
+// access for various users on different objects and endpoints:
+//
+// * **[Cluster permissions](clusters)** — Manage which users can manage,
```
nfx, line 795: run `make doc`. this may break the https://pkg.go.dev/github.com/databricks/databricks-sdk-go docs, as it points to non-existing links.

workspace_client.go
```diff
@@ -367,7 +367,52 @@ type WorkspaceClient struct {
 	ModelRegistry *ml.ModelRegistryAPI
 
 	// Permissions API are used to create read, write, edit, update and manage
-	// access for various users on different objects and endpoints.
+	// access for various users on different objects and endpoints:
+	//
+	// * **[Cluster permissions](clusters)** — Manage which users can manage,
```
nfx, line 372: we have to use different syntax for cross-service referencing or links. See `Deep linking: methods, classes, and external links` from http://go/openapi/spec. 

can you also add a linter to prevent these from being added in universe?

nfx (changes requested):

service/iam/api.go
mgyucht: Screenshot 2023-08-03 at 13 21 23
No links to break if there are no links :)
Any prior art w.r.t. links within the REST API docs?

service/oauth2/model.go
```diff
@@ -11,6 +11,9 @@ type CreateCustomAppIntegration struct {
 	Name string `json:"name"`
 	// List of oauth redirect urls
 	RedirectUrls []string `json:"redirect_urls"`
+	// OAuth scopes granted to the application. Supported scopes: all-apis, sql,
+	// offline_access, openid, profile, email.
+	Scopes []string `json:"scopes,omitempty"`
```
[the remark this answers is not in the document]
mgyucht: This may have other OpenAPI changes in it as well, I just wanted to update ASAP to detect any potential issues.

service/pkg.go
```diff
@@ -82,7 +82,7 @@
 //
 // - [oauth2.OAuthEnrollmentAPI]: These APIs enable administrators to enroll OAuth for their accounts, which is required for adding/using any OAuth published/custom application integration.
 //
-// - [iam.PermissionsAPI]: Permissions API are used to create read, write, edit, update and manage access for various users on different objects and endpoints.
+// - [iam.PermissionsAPI]: Permissions API are used to create read, write, edit, update and manage access for various users on different objects and endpoints: * **[Cluster permissions](clusters)** — Manage which users can manage, restart, or attach to clusters.
```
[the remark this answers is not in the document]
mgyucht: Screenshot 2023-08-03 at 13 23 36
Looks bad.

workspace_client.go
mgyucht: Yes, will add a linter there.

mgyucht: Screenshot 2023-08-03 at 13 28 53
That style of link also does not seem to be handled appropriately?

service/iam/api.go
mgyucht: From what I can tell, it is not possible to link to a service in our current OpenAPI model. I filed a ticket with @arusanov to support this. For now, I'm going to revert the links in this file

workspace_client.go
mgyucht: Filed a ticket with @arusanov to add linting

service/iam/api.go
nfx: yep, also in the openapi spec. this just breaks the logical flow of code comments.

.codegen/_openapi_sha
```diff
@@ -1 +1 @@
-fbdd0fa3e83fed2c798a58d376529bdb1285b915
\ No newline at end of file
+universe:/Users/miles/universe
```
nfx, line 1: looks very unintentional ;) i'd prefer if we remove the ability to put rfs into this file.

PS: I just associate `refish` with "a fish named RE". Can we change that? =P

nfx (changes requested):
does godoc support bold highlighting? I don't think so... but if we want to roll it out ASAP, then just fix the sha file

.codegen/_openapi_sha
mgyucht: I suppose we could call it `ref`?

nfx: ref is better :)

nfx (approved):

PR #572, opened 2023-08-07
Support Files API (application/octet-stream) in OpenAPI Spec

mgyucht:
## Changes
This PR adds support for non-application/json request and response content types in the Databricks OpenAPI specification. This PR introduces the following changes to the OpenAPI library interface:

**Entity**
* `IsByteStream bool` -> `true` iff the entity should be modeled as a byte stream. This is only allowed for top-level entities (i.e. that are used directly as a request or response) or for the field within such an entity, provided that that field is the "contents" field (for requests)

**Method**
* `RequestBodyField *Field` -> a pointer to the field of the request entity that is the body of the request. Only to be used when the request body can be captured by a single field (today, this is only allowed for the "contents" field of non-application/json requests).
* `FixedRequestHeaders map[string]string` -> the headers which are always the same, regardless of request parameters. In the future, a new method could be added to support fields from requests that are provided as headers.
* `(m *Method) IsRequestByteStream() bool` -> `true` iff the request is modeled as a byte stream (i.e. != application/json)
* `(m *Method) IsResponseByteStream() bool` -> `true` iff the response is modeled as a byte stream (i.e. != application/json).

**Parameter**
* `MultiSegment bool `json:"x-databricks-multi-segment,omitempty"`` -> indicates that this parameter is a path parameter that may contain slashes, which is not officially compliant with OpenAPI. (unused today)

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


openapi/code/entity.go
```diff
@@ -54,16 +54,17 @@ type EnumEntry struct {
...
+	Package     *Package
+	enum        map[string]EnumEntry
+	ArrayValue  *Entity
+	MapValue    *Entity
+	IsInt       bool
+	IsInt64     bool
+	IsFloat64   bool
+	IsBool      bool
+	IsString    bool
+	IsByteArray bool
```
nfx, line 66: shouldn't it rather be on a method level?

nfx (dismissed):

openapi/code/method.go
```diff
@@ -36,6 +36,13 @@ type Method struct {
 	// the user-friendly name of the resource.
 	NameFieldPath []*Field
 
+	// If not the zero value, represent the field in the request entity that
+	// should be mapped to the request body.
+	RequestBodyField *Field
```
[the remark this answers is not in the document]
mgyucht: We still need this because request/response structures need to name the field that corresponds to the body...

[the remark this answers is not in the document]
nfx: can you add a link to a standard?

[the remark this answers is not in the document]
mgyucht: Yeah we haven't approved it yet, but I will once it is ready.

openapi/code/service.go
```diff
@@ -131,20 +131,66 @@ var crudNames = map[string]bool{
...
+	entity.IsByteStream = true
+	entity.IsEmpty = false
+	entity.IsAny = false
+}
+
+// Construct the base request entity for a given operation. For requests whose
+// mime type is not application/json, the request body is always a byte stream.
+// For requests whose mime type is application/json, the request body consists
+// of the top-level fields of the request object as defined in the OpenAPI spec.
+// Additionally, if the x-databricks-body-field-name extension is specified,
```
mgyucht, line 149: field is always "contents"

openapi/code/service.go
```diff
@@ -131,20 +131,66 @@ var crudNames = map[string]bool{
...
+	if op.RequestBody == nil {
+		return &Entity{fields: map[string]*Field{}}, "", nil
 	}
-	request := &Entity{fields: map[string]*Field{}}
-	if op.RequestBody != nil {
-		request = svc.Package.schemaToEntity(op.RequestBody.Schema(), []string{op.Name()}, true)
+	mimeType, mediaType := op.RequestBody.MimeTypeAndMediaType()
+	requestSchema := mediaType.GetSchema()
+	if requestSchema == nil {
+		return &Entity{fields: map[string]*Field{}}, "", nil
```
mgyucht, line 160: Maybe remove this: what if `"application/octet-stream": {}`?

openapi/code/service.go
```diff
@@ -131,20 +131,66 @@ var crudNames = map[string]bool{
...
 	}
-	if request == nil {
+	// If x-databricks-body-field-name is specified, the request structure
```
mgyucht, line 162: remove this

openapi/code/method.go
```diff
@@ -36,6 +36,14 @@ type Method struct {
 	// the user-friendly name of the resource.
 	NameFieldPath []*Field
 
+	// If not nil, the field in the request entity that should be mapped to the
+	// request body.
+	RequestBodyField *Field
+
+	// Expected content type of the request and response
+	ContentType openapi.MimeType
+	Accept      openapi.MimeType
```
mgyucht, line 45: Change to map

openapi/code/service.go
```diff
@@ -131,20 +131,66 @@ var crudNames = map[string]bool{
...
-		return nil
+	var bodyField *Field
+	entity := res
+	if mimeType.IsByteStream() {
+		bodyField = res.fields[openapi.MediaTypeNonJsonBodyFieldName]
+		entity = bodyField.Entity
+		res.RequiredOrder = append(res.RequiredOrder, openapi.MediaTypeNonJsonBodyFieldName)
 	}
+	svc.updateEntityTypeFromMimeType(entity, mimeType)
+	if mimeType != openapi.MimeTypeJson {
```
mgyucht, line 184: Give precautionary warning about architecture of schemaToEntity

openapi/code/service.go
```diff
@@ -189,24 +235,67 @@ func (svc *Service) newRequest(params []openapi.Parameter, op *openapi.Operation
...
-		request.Description = op.Summary
-		svc.Package.define(request)
+		svc.nameAndDefineRequest(request, op)
 	}
-	return request
+	return request, mimeType, bodyField
+}
+
+func (svc *Service) newResponse(op *openapi.Operation) (*Entity, openapi.MimeType, Named) {
+	respMimeType, respBody := op.SuccessResponseSchema(svc.Package.Components)
```
mgyucht, line 289: change name of this

openapi/code/service.go
```diff
@@ -131,20 +131,66 @@ var crudNames = map[string]bool{
...
+	mimeType, mediaType := op.RequestBody.MimeTypeAndMediaType()
+	requestSchema := mediaType.GetSchema()
+	if requestSchema == nil {
+		return &Entity{fields: map[string]*Field{}}, "", nil
 	}
-	if request == nil {
+	// If x-databricks-body-field-name is specified, the request structure
+	// contains a field with that name whose value is the request body.
+	if mimeType.IsByteStream() {
+		requestSchema = &openapi.Schema{
```
mgyucht, line 165: let's use this logic in responses as well rather than returning bare bytestreams.

mgyucht:
Follow-ups 1 & 3 are addressed in this PR. 2 is tracked on our backlog and I'll pick it up next week.

tanmay-db (approved):

openapi/code/service.go
```diff
@@ -189,24 +239,89 @@ func (svc *Service) newRequest(params []openapi.Parameter, op *openapi.Operation
...
+	name := op.Name()
+	response := svc.Package.definedEntity(name+"Response", schema)
+
+	// This next block of code is needed to make up for shortcomings in
+	// schemaToEntity. That function (and the Entity structure) assumes that all
+	// entities are modeled by JSON objects. Later, we should change Entity
+	// and schemaToEntity to be more tolerant of non-JSON entities; for now, we
+	// put this hack in place to make things work.
+	if mimeType.IsByteStream() {
+		bodyField := response.fields[openapi.MediaTypeNonJsonBodyFieldName]
```
mgyucht, line 313: Need to return body field and add to Method

PR #577, opened 2023-08-10
Properly skip request placeholder types

nfx:
This PR adds a way to list types that are not mere request placeholders, so that Python SDK keyed args UX is not cluttered with the types of no use.

To be verified after https://github.com/databricks/databricks-sdk-py/pull/264 is merged

### DIFF
--- a/openapi/code/service.go
+++ b/openapi/code/service.go
@@ -206,6 +206,17 @@ func (svc *Service) newRequest(params []openapi.Parameter, op *openapi.Operation
 		request.Description = op.Summary
 		svc.Package.define(request)
 	}
+	isOnlyUsedAsRequest := true
+	for _, e := range svc.Package.types {
+		for _, f := range e.fields {
+			if f.Entity == request {
+				// this type is not only a request placeholder, but is also
+				// part of other type definitions.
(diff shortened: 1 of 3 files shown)


openapi/code/service.go
```diff
@@ -206,6 +206,17 @@ func (svc *Service) newRequest(params []openapi.Parameter, op *openapi.Operation
...
+	for _, e := range svc.Package.types {
+		for _, f := range e.fields {
+			if f.Entity == request {
+				// this type is not only a request placeholder, but is also
+				// part of other type definitions.
+				isOnlyUsedAsRequest = false
+			}
+		}
+	}
+	request.IsRequestOnly = isOnlyUsedAsRequest
```
mgyucht, line 219: Doesn't this need to be done as a postprocessing step? The order in which requests are defined is arbitrary AFAIK, so this could accidentally set types as request only if their uses are not yet processed

nfx: > Doesn't this need to be done as a postprocessing step?

it could be in `NonPlaceholderTypes()` method. may add some false positives, but will try that.

renaudhartert-db:
Closing stale PRs

PR #582, opened 2023-08-15
Don't access `param` unless we know it's there

pietern:


nfx (commented):
can you run `make gen` to make sure it doesn't break stuff

openapi/code/package.go
```diff
@@ -350,15 +350,15 @@ func (pkg *Package) Load(spec *openapi.Specification, tag openapi.Tag) error {
 			for _, list := range [][]openapi.Parameter{path.Parameters, op.Parameters} {
 				for _, v := range list {
 					param, err := pkg.resolveParam(&v)
-					if param.In == "header" {
-						continue
-					}
 					if err != nil {
 						return fmt.Errorf("no components found: %s %s", verb, prefix)
```
mgyucht, line 354:
```suggestion
						return fmt.Errorf("could not resolve parameter %s for %s %s. This could be due to a problem in the definition of this parameter. If using $ref, ensure that $ref is used inside the 'schema' keyword.", v.Name, verb, prefix)
```

mgyucht (commented):
Better error message that hopefully would have helped @tywin1104 out

mgyucht (approved):
One small suggestion, otherwise LGTM!

PR #584, opened 2023-08-17
Set necessary headers when authenticating via Azure CLI

mgyucht:
## Changes
The Go SDK request authentication logic is inconsistent between the Azure login types: for service principal & MSI auth, the SDK correctly adds the X-Databricks-Azure-Workspace-Resource-Id when configured, but this is missed for Azure CLI auth. Additionally, when logging in via Azure CLI using a service principal, the service management token must also be fetched from the CLI. This caused a regression for the Terraform provider: https://github.com/databricks/terraform-provider-databricks/issues/2590.

This PR fixes this by defining the logic to attach these header in a common function that is used by all Azure-specific authentication types.

## Tests
- [x] Added a unit test to ensure the header is being set for Azure CLI login
- [x] Made a test app that uses `azure-cli` to login and verified that the correct header was set on the request:
```
...
> * X-Databricks-Azure-Workspace-Resource-Id: /subscriptions/<REDACTED>/resourceGroups/<REDACTED>/pr... (63 more bytes)
```

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/config/auth_azure_cli.go
+++ b/config/auth_azure_cli.go
@@ -15,7 +15,8 @@ import (
 )
 
 // The header used to pass the workspace resource ID to the Databricks backend.
-const XDatabricksAzureSpManagementToken = "X-Databricks-Azure-SP-Management-Token"
+const xDatabricksAzureSpManagementToken = "X-Databricks-Azure-SP-Management-Token"
+const xDatabricksAzureWorkspaceResourceId = "X-Databricks-Azure-Workspace-Resource-Id"
 
 type AzureCliCredentials struct {
 }
@@ -28,6 +29,29 @@ func (c AzureCliCredentials) Name() string {
 func (c AzureCliCredentials) tokenSourceFor(
 	ctx context.Context, cfg *Config, env azureEnvironment, resource string) oauth2.TokenSource {
 	return &azureCliTokenSource{resource: resource}
+}
+
+// There are three scenarios:
+//
+//  1. The user has logged in with the Azure CLI as a user and has access to the service management endpoint.
+//  2. The user has logged in with the Azure CLI as a user and does not have access to the service management endpoint.
+//  3. The user has logged in with the Azure CLI as a service principal, and must have access to the service management
+//     endpoint to authenticate.
+//
+// If the user can't access the service management endpoint, we assume they are in case 2 and do not pass the service
+// management token. Otherwise, we always pass the service management token.
+func (c AzureCliCredentials) getVisitor(ctx context.Context, cfg *Config, innerTokenSource oauth2.TokenSource) (func(*http.Request) error, error) {
+	env, err := cfg.GetAzureEnvironment()
+	if err != nil {
+		return nil, err
+	}
+	managementTs := &azureCliTokenSource{env.ServiceManagementEndpoint, ""}
+	_, err = managementTs.Token()
+	if err != nil {
+		logger.Debugf(ctx, "Not including service management token in headers: %v", err)
+		return refreshableVisitor(innerTokenSource), nil
+	}
+	return serviceToServiceVisitor(innerTokenSource, managementTs, xDatabricksAzureSpManagementToken), nil
 }
 
 func (c AzureCliCredentials) Configure(ctx context.Context, cfg *Config) (func(*http.Request) error, error) {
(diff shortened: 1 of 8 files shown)


config/oauth_visitors.go
```diff
@@ -56,3 +56,12 @@ func refreshableVisitor(inner oauth2.TokenSource) func(r *http.Request) error {
 		return nil
 	}
 }
+
+func azureVisitor(workspaceResourceId string, inner func(*http.Request) error) func(*http.Request) error {
+	return func(r *http.Request) error {
+		if workspaceResourceId != "" {
+			r.Header.Set("X-Databricks-Azure-Workspace-Resource-Id", workspaceResourceId)
```
alexott, line 63: Besides this header, we may need to add service management header as well: https://github.com/databricks/terraform-provider-databricks/blob/v1.9.2/common/azure_auth.go#L147

config/auth_azure_cli.go
```diff
@@ -50,11 +54,32 @@ func (c AzureCliCredentials) Configure(ctx context.Context, cfg *Config) (func(*
...
+	// management token. Otherwise, we always pass the service management token.
+	var managementTs oauth2.TokenSource
+	env, err := cfg.GetAzureEnvironment()
+	if err != nil {
+		return nil, err
+	}
+	managementTs = &azureCliTokenSource{env.ServiceManagementEndpoint, ""}
+	_, err = managementTs.Token()
+	if err != nil {
+		managementTs = nil
```
pietern, line 75: Debug log with the error would be nice. Will make all the difference debugging this not working.

config/oauth_visitors.go
```diff
@@ -27,32 +27,44 @@ func retriableTokenSource(ctx context.Context, ts oauth2.TokenSource) (*oauth2.T
...
 		}
-		r.Header.Set(header, cloud.AccessToken)
 		return nil
 	}
 }
 
 func refreshableVisitor(inner oauth2.TokenSource) func(r *http.Request) error {
-	refreshableInner := oauth2.ReuseTokenSource(nil, inner)
+	return serviceToServiceVisitor(inner, nil, "", false)
+}
```
pietern, line 61: I recommend leaving `refreshableVisitor` as is.

The additional nil checks to make this possible make it much more complex, IMO.

config/oauth_visitors.go
pietern: Isn't that already added by the `serviceToServiceVisitor`?

The header in question: `X-Databricks-Azure-SP-Management-Token`.

config/auth_azure_cli.go
pietern: Separate comment, why call `.Token()` in the first place if we ignore the error in the visitor if we call it there? I suspect this is to avoid calling it repeatedly if we don't have one anyway. If so, please add comment.

pietern: And, if we do already pull this token eagerly, we should not tolerate errors on the secondary token source (the final arg to the visitor function).

mgyucht: I kind of blended two changes together. Originally, I had a version that would check for the management token when making requests and tolerating failures there. Partway through, I changed it to a version that checked for the management token during Configure() and skips even passing it to the visitor factory. You're right, we don't need to tolerate errors (and we can do one better: just call the refreshableVisitor method instead).

config/oauth_visitors.go
mgyucht: After looking at this original change, we realized that there was a secondary problem: if a user logs in via the CLI using a service principal, the CLI auth type needs to also request the management token from the CLI in addition to the Databricks token. The point is: all Azure-native auth types need to call `azureVisitor`, and all auth types that need to include a second token in the request need to call `serviceToServiceVisitor`; this now includes the CLI.

config/auth_azure_cli.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/databricks/databricks-sdk-go/logger"
 )
 
+// The header used to pass the workspace resource ID to the Databricks backend.
```
pietern, line 17: The comment is now out of date / we need another one.

config/auth_azure_cli.go
```diff
@@ -50,11 +78,16 @@ func (c AzureCliCredentials) Configure(ctx context.Context, cfg *Config) (func(*
 		return nil, fmt.Errorf("resolve host: %w", err)
 	}
 	logger.Infof(ctx, "Using Azure CLI authentication with AAD tokens")
-	return refreshableVisitor(&ts), nil
+	baseVisitor, err := c.getVisitor(ctx, cfg, ts)
+	if err != nil {
+		return nil, err
+	}
+	return azureVisitor(cfg.AzureResourceID, baseVisitor), nil
```
pietern, line 85: This can be folded into `getVisitor` to make things a bit simpler (i.e. "what is a base visitor?").

config/oauth_visitors.go
```diff
@@ -27,32 +27,51 @@ func retriableTokenSource(ctx context.Context, ts oauth2.TokenSource) (*oauth2.T
 	})
 }
 
-func serviceToServiceVisitor(inner, cloud oauth2.TokenSource, header string) func(r *http.Request) error {
-	refreshableInner := oauth2.ReuseTokenSource(nil, inner)
-	refreshableCloud := oauth2.ReuseTokenSource(nil, cloud)
+// serviceToServiceVisitor returns a visitor that sets the Authorization header to the token from the auth token source
+// and the provided secondary header to the token from the secondary token source. If secondary is nil, the secondary
+// header is not set.
+func serviceToServiceVisitor(auth, secondary oauth2.TokenSource, secondaryHeader string) func(r *http.Request) error {
```
pietern, line 33: In the current approach I believe the secondary one is never nil anymore.

config/auth_azure_cli.go
mgyucht: Sounds good. I did it this way to not have to write azureVisitor twice. I'll adopt your suggestion, but in general I try to only duplicate when things intentionally could/should diverge and not duplicate when things must have identical behavior.

alexott (approved):
lgtm

config/oauth_visitors.go
pietern: The comment still mentions nil secondary.

pietern (approved):

PR #589, opened 2023-08-19
Propagate Request Headers to client.go

mgyucht:
## Changes
Based on changes in https://github.com/databricks/databricks-sdk-go/pull/572.

This PR adds initial support for headers in client.go. The `impl.go` implementations, when making a request, will construct a map of headers to be included in the request. These are passed through a new parameter in `client.Do()`.

In the future, to support dynamic headers, we can modify our codegen template to add those fields to the headers map before calling the client.

Alternatives considered:
* **Adding headers in request objects directly**: this would not require any change in the client interface. Instead, we would introduce a new annotation for headers, much like what we do for query parameters, and reflectively scan the fields of each request object. We could do this, but for headers that have a fixed value (like `Content-Type` and `Accept`), these fields would essentially be private final fields that we could iterate through. Users can never interact with these fields, so it seems a bit unusual to have them in the request structure, a public interface.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


.codegen/impl.go.tmpl
```diff
@@ -24,8 +24,16 @@ func (a *{{.Service.CamelName}}Impl) {{.PascalName}}(ctx context.Context{{if .Re
 	path := {{if .PathParts -}}
 		fmt.Sprintf("{{range  .PathParts}}{{.Prefix}}{{if or .Field .IsAccountId}}%v{{end}}{{ end }}"{{ range .PathParts }}{{if .Field}}, request.{{.Field.PascalName}}{{ else if .IsAccountId }}, a.client.ConfiguredAccountID(){{end}}{{ end }})
 	{{- else}}"{{.Path}}"{{end}}
-	err := a.client.Do(ctx, http.Method{{.TitleVerb}}, path, {{if .Request}}request{{else}}nil{{end}}, {{if .Response}}&{{.Response.CamelName}}{{else}}nil{{end}})
+	{{ template "make-header" . }}
+	err := a.client.Do(ctx, http.Method{{.TitleVerb}}, path, headers, {{if .Request}}request{{else}}nil{{end}}, {{if .Response}}&{{.Response.CamelName}}{{else}}nil{{end}})
```
nfx, line 28: adding headers makes unnecessarily large diff. did you try doing it with the request visitor? we're adding other headers with that mechanism already. This could would work without the need to change signature of the method (and this unnecessary diff) - 

```suggestion
	err := a.client.Do(ctx, http.Method{{.TitleVerb}}, path, {{if .Request}}request{{else}}nil{{end}}, {{if .Response}}&{{.Response.CamelName}}{{else}}nil{{end}}{{with .FixedRequestHeaders}}, func(r *http.Request) error {
	{{- range $k, $v := . }}
	r.Header["{{$k}}"] = "{{$v}}"
	{{- end }}
	return nil
	}{{end}})
```

you've even touched the code following this pattern:
image

nfx (changes requested):
could you use request visitor for adding request headers, that doesn't require the diff in every callsite?

tanmay-db (approved):
LGTM

PR #590, opened 2023-08-20
Support io.ReadCloser for request and response bodies

mgyucht:
## Changes
This PR adds support for non-application/json requests and responses. This PR is divided into two parts:

**Support setting headers per request**. The `Do` method of DatabricksClient is expanded to accept one new parameter, `headers map[string]string`, and all callers are expected to pass a map of headers to be included in the request. As the allowed content types for requests and responses are known from the OpenAPI specification, callers must construct the request header map in code generation and pass it to the client. The default "Content-Type: application/json" header is removed, as each request should specify its own Content-Type and Accept headers. This is done in https://github.com/databricks/databricks-sdk-go/pull/589.

**Add support for streaming request and response bodies to support non-application/json requests/responses**. Today, serialization of requests is done in the client. This implies that ApiClient needs to be able to serialize a request purely based on the parameters provided to it via headers, the request body, etc. In this proposal, the serialization is specified by the caller depending on whether the value passed to the `request` and `response` parameters is an `io.ReadCloser` or not.

One important caveat about streaming responses is that callers are required to close the streams when they are done using them.
## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


nfx:
By the way, shouldn’t we do io.Writer for requests and io.Reader for responses? It sounds more logical

mgyucht:
io.Reader allows the client to consume data from a given source. If you're reading data from a file for example, you'll get back an io.Reader, which you can pass directly to the client. Otherwise, you'd have to read from the reader and write to the writer to pass to the client. io.Writer would work for an API where you don't have the data at the moment you call `uploadFile` but later want to write it out somewhere.

tanmay-db (approved):

client/client.go
```diff
@@ -128,84 +166,103 @@ func (c *DatabricksClient) addHostToRequestUrl(r *http.Request) error {
...
+			}
+			return c.wrapError(apiError.IsRetriable(ctx), apiError)
+		}
+		responseBody, responseBodyErr := c.fromResponse(response)
+		if responseBodyErr != nil {
+			return c.wrapError(false, apierr.ReadError(response.StatusCode, responseBodyErr))
 		}
 
-		retry, err := apierr.CheckForRetry(ctx, response, err, responseBody.Bytes(), responseBodyErr)
+		retry, err := apierr.CheckForRetry(ctx, response, err, responseBody.ReadCloser)
```
mgyucht, line 248: err is always nil here, so we can remove

client/client.go
```diff
@@ -128,84 +166,103 @@ func (c *DatabricksClient) addHostToRequestUrl(r *http.Request) error {
...
 		var apiErr *apierr.APIError
 		if err != nil && !errors.As(err, &apiErr) {
 			err = fmt.Errorf("failed request: %w", err)
 		}
 		if err == nil && response == nil {
 			err = fmt.Errorf("no response: %s %s", method, requestURL)
 		}
```
mgyucht, line 255: Should be guaranteed by http.Client:
> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.

client/client.go
```diff
@@ -128,84 +166,103 @@ func (c *DatabricksClient) addHostToRequestUrl(r *http.Request) error {
...
-		var responseBodyErr error
-		if response != nil {
-			_, responseBodyErr = responseBody.ReadFrom(response.Body)
-			response.Body.Close()
+		if ue, ok := err.(*url.Error); ok {
+			apiError := &apierr.APIError{
+				ErrorCode:  "IO_ERROR",
+				StatusCode: 523,
+				Message:    ue.Error(),
+			}
```
mgyucht, line 240: This could be refactored

client/client_test.go
```diff
@@ -62,7 +62,9 @@ func TestSimpleRequestFails(t *testing.T) {
 			assert.Equal(t, "f", r.Header.Get("e"))
 			auth := r.Header.Get("Authenticated")
 			assert.Equal(t, "yes", auth)
-			return nil, fmt.Errorf("nope")
+			return &http.Response{
+				Request: r,
+			}, fmt.Errorf("nope")
```
mgyucht, line 67: Return nil but wrap err with URLError

PR #591, opened 2023-08-20
Update OpenAPI Spec & Merge Files API

mgyucht:
## Changes
Autogenerate Files API in Go SDK, including other recent changes.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


tanmay-db (approved):
LGTM, since service/files/files_api.go is removed and it's part of Open API spec, I will update the terraform side changes to account for this (would require a release).

PR #595, opened 2023-08-31
Fix loading of databrickscfg with a password containing a hash

pietern:
## Changes

Pass options to the ini loader to indicate that inline comments require a leading whitespace. Any existing configuration that uses inline comments _without_ a leading whitespace will be interpreted differently with this change.

## Tests

- [x] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/config/config_file.go
+++ b/config/config_file.go
@@ -45,7 +45,13 @@ func LoadFile(path string) (*File, error) {
 		path = fmt.Sprintf("%s%s", homedir, path[1:])
 	}
 
-	iniFile, err := ini.Load(path)
+	opts := ini.LoadOptions{
+		// Passwords may contain '#'. Require a space before '#' for inline comments.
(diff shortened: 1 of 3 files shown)


mgyucht (approved):
LGTM!

config/config_file.go
```diff
@@ -45,7 +45,13 @@ func LoadFile(path string) (*File, error) {
 		path = fmt.Sprintf("%s%s", homedir, path[1:])
 	}
 
-	iniFile, err := ini.Load(path)
+	opts := ini.LoadOptions{
```
nfx, line 48: nit: `opts` is not used anywhere after line 54, can you inline it?

nfx (approved):

config/config_file.go
pietern: Done.

PR #597, opened 2023-08-31
Handle Azure authentication when WorkspaceResourceID is provided

hectorcast-db:
## Changes
Handle Azure authentication when WorkspaceResourceID is provided

Get token for the correct subscription 

## Tests
* Created Unit tests
* Manually listed workspace cluster in the following scenarios:
  * User with wrong default tenant. No WorkspaceResourceID provided: Fail (expected). WARN log emitted.
  * User with wrong default tenant. WorkspaceResourceID provided: Succeed
  * User with no subscription. No WorkspaceResourceID provided: Succeed. WARN log  emitted.
  * User with no subscription. WorkspaceResourceID provided: Succeed (fallback mode, expected).

- [X] `make test` passing
- [X] `make fmt` applied
https://github.com/databricks/eng-dev-ecosystem/actions/runs/6038851262/job/16386160429

### DIFF
--- a/config/auth_azure_cli.go
+++ b/config/auth_azure_cli.go
@@ -95,17 +95,49 @@ type internalCliToken struct {
 	ExpiresOn    string `json:"expiresOn"`
 }
 
-func (ts *azureCliTokenSource) Token() (*oauth2.Token, error) {
-	out, err := exec.Command("az", "account", "get-access-token", "--resource",
-		ts.resource, "--output", "json").Output()
-	if ee, ok := err.(*exec.ExitError); ok {
-		return nil, fmt.Errorf("cannot get access token: %s", string(ee.Stderr))
+func (ts *azureCliTokenSource) getSubscription() string {
+	if ts.workspaceResourceId == "" {
+		return ""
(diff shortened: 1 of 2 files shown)


config/auth_azure_cli.go
```diff
@@ -128,3 +136,30 @@ func (ts *azureCliTokenSource) Token() (*oauth2.Token, error) {
...
+			return result, nil
+		}
+		logger.Warnf(context.Background(), "Failed to get token for subscription. Using resource only token.")
+	}
+	result, err := exec.Command("az", args...).Output()
+	if ee, ok := err.(*exec.ExitError); ok {
+		return nil, fmt.Errorf("cannot get access token: %s", string(ee.Stderr))
+	}
+	if err != nil {
+		return nil, fmt.Errorf("cannot get access token: %v", err)
```
mgyucht, line 162: We should use `%w` to wrap errors, unless for some reason we don't want to expose the underlying error?

mgyucht (approved):
One small nit, otherwise LGTM!

PR #598, opened 2023-09-01
Use an operation's request type name if specified

pietern:
## Changes

The spec can contain the `x-databricks-request-type-name` annotation for requests that do not have a request body but need a request type to parameterize their URL path and query. This is necessary to 1) keep the method names for permissions APIs at the service level in check, and 2) have unique request type names.

When using this change with a new spec, the diff is (for the `JobsService`):
* `GetJobPermissionLevels` -> `GetPermissionLevels`
* `GetJobPermissions` -> `GetPermissions`
* `SetJobPermissions` -> `SetPermissions`
* `UpdateJobPermissions` -> `UpdateJobPermissions`

The request type names for these methods remain the same (because of this commit).

## Tests

Manually confirmed that with this change the request type name is used if specified

### DIFF
--- a/openapi/code/service.go
+++ b/openapi/code/service.go
@@ -3,6 +3,7 @@ package code
 import (
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/databricks/databricks-sdk-go/openapi"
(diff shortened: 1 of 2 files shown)


mgyucht (approved):
LGTM!

PR #600, opened 2023-09-04
Update OpenAPI spec

pietern:
## Changes

This includes a change to how we expose the permissions-related APIs, for example (for the `JobsService`):
* `GetJobPermissionLevels` -> `GetPermissionLevels`
* `GetJobPermissions` -> `GetPermissions`
* `SetJobPermissions` -> `SetPermissions`
* `UpdateJobPermissions` -> `UpdatePermissions`

These were added in #567 and first released as part of v0.15.0.

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


service/compute/api.go
```diff
@@ -117,15 +117,15 @@ func (a *ClusterPoliciesAPI) GetByPolicyId(ctx context.Context, policyId string)
...
+func (a *ClusterPoliciesAPI) GetPermissionLevels(ctx context.Context, request GetClusterPolicyPermissionLevelsRequest) (*GetClusterPolicyPermissionLevelsResponse, error) {
+	return a.impl.GetPermissionLevels(ctx, request)
 }
 
 // Get cluster policy permission levels.
 //
 // Gets the permission levels that a user can have on an object.
-func (a *ClusterPoliciesAPI) GetClusterPolicyPermissionLevelsByClusterPolicyId(ctx context.Context, clusterPolicyId string) (*GetClusterPolicyPermissionLevelsResponse, error) {
-	return a.impl.GetClusterPolicyPermissionLevels(ctx, GetClusterPolicyPermissionLevelsRequest{
+func (a *ClusterPoliciesAPI) GetPermissionLevelsByClusterPolicyId(ctx context.Context, clusterPolicyId string) (*GetClusterPolicyPermissionLevelsResponse, error) {
```
mgyucht, line 127: While we're dealing with name: what do you think about `GetPermissionLevelsById`? The type of ID is implicit in the service name. I just looked into this, and this seems to be derived from the name of the path parameter. If we want this, we will need to change all path parameters to be just `{id}` instead of `{cluster_policy_id}`, `{cluster_id}`, etc. This would also be nice from a perspective of consistency.

mgyucht (approved):
LGTM. One longer-term discussion point that isn't affected by this PR but should be considered for our roadmap w.r.t. API stabilization.

PR #601, opened 2023-09-04
Work on GitHub Action

pietern:
## Changes

* Update to actions/setup-go@v4; this uses caching by default
* Run the build workflow on push to main to properly use the build cache
* Include Go 1.21 in the test matrix

## Tests

n/a

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -5,6 +5,14 @@ on:
     types: [opened, synchronize]
   merge_group:
     types: [checks_requested]
+  push:
+    # Always run on push to main. The build cache can only be reused
+    # if it was saved by a run from the repository's default branch.
+    # The run result will be identical to that from the merge queue
+    # because the commit is identical, yet we need to perform it to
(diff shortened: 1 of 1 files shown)


.github/workflows/push.yml
```diff
@@ -5,14 +5,22 @@ on:
     types: [opened, synchronize]
   merge_group:
     types: [checks_requested]
+  push:
```
mgyucht, line 8: Did not know about this! Can you send a link? Does this apply to all other repos?

mgyucht (approved):

.github/workflows/push.yml
hectorcast-db: I found this
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#about-caching-workflow-dependencies
It seems to also work for multiple runs in the current branch. But no sharing between non default branches for security reasons.

hectorcast-db (approved):

PR #604, opened 2023-09-05
Revert error message changes from client.go

mgyucht:
## Changes
https://github.com/databricks/databricks-sdk-go/pull/590 made several changes in the error messages returned from the client. It turns out that many users depend on these error messages exactly. I'll revert this change for now to unblock users who depended on these error messages exactly. Going forward, we recommend that users check if the returned error is an API Error, and then to use the ErrorCode to make decisions about how to handle errors in the APIs.

## Tests
- [x] Unit tests are reverted and no longer reference these nested fields.
- [x] Upgraded TF provider to use this commit locally and unit tests pass.


tanmay-db (approved):
Thanks for doing this

pietern (approved):

PR #611, opened 2023-09-11
Example diff for PR-380074 in universe repo

eugenekoh:
This is the generated diff for https://github.com/databricks/universe/pull/380074


service/billing/model.go
```diff
@@ -179,7 +179,15 @@ type DeleteBudgetRequest struct {
 	BudgetId string `json:"-" url:"-"`
 }
 
-// This describes an enum
+// The status string for log delivery. Possible values are: * `CREATED`: There
```
eugenekoh, line 182: @pietern This is a result of the changes in https://github.com/databricks/universe/pull/380074. We should do some post-processing in the SDK generators to replace this comment. 

Previously, we replaced the description with `This describes an enum` which led to an information loss. We need to retain the description for the API reference.

service/billing/model.go
```diff
@@ -383,7 +391,16 @@ type LogDeliveryStatus struct {
 	// delivery fails with USER_FAILURE, error details will be provided for
 	// fixing misconfigurations in cloud permissions.
 	Message string `json:"message,omitempty"`
-	// This describes an enum
+	// The status string for log delivery. Possible values are: * `CREATED`:
```
eugenekoh, line 394: Change due to PR

service/catalog/impl.go
```diff
@@ -41,13 +41,13 @@ func (a *accountMetastoreAssignmentsImpl) Get(ctx context.Context, request GetAc
 	return &accountsMetastoreAssignment, err
 }
 
-func (a *accountMetastoreAssignmentsImpl) List(ctx context.Context, request ListAccountMetastoreAssignmentsRequest) ([]MetastoreAssignment, error) {
-	var metastoreAssignmentList []MetastoreAssignment
+func (a *accountMetastoreAssignmentsImpl) List(ctx context.Context, request ListAccountMetastoreAssignmentsRequest) ([]int64, error) {
+	var workspaceIdList []WorkspaceId
```
eugenekoh, line 45: This error is unrelated to my PR and is causing the tests to fail.

service/compute/model.go
```diff
@@ -334,7 +334,23 @@ type ClusterAttributes struct {
 	// - Clusters can only reuse cloud resources if the resources' tags are a
 	// subset of the cluster tags
 	CustomTags map[string]string `json:"custom_tags,omitempty"`
-	// This describes an enum
+	// Data security mode decides what data governance model to use when
```
eugenekoh, line 337: Change caused by pr

service/compute/model.go
```diff
@@ -462,7 +478,23 @@ type ClusterDetails struct {
...
```
eugenekoh, line 481: Change caused by PR

service/compute/model.go
```diff
@@ -860,7 +892,23 @@ type ClusterSpec struct {
...
```
eugenekoh, line 895: Change caused by PR

service/jobs/model.go
```diff
@@ -101,7 +101,11 @@ type BaseRun struct {
 	RunName string `json:"run_name,omitempty"`
 	// The URL to the detail page of the run.
 	RunPageUrl string `json:"run_page_url,omitempty"`
-	// This describes an enum
+	// * `JOB_RUN`: Normal job run. A run created with :method:jobs/runNow. *
```
eugenekoh, line 104: Change caused by PR

service/ml/model.go
```diff
@@ -82,7 +101,19 @@ func (f *ActivityAction) Type() string {
 	return "ActivityAction"
 }
 
-// This describes an enum
+// Type of activity. Valid values are: * `APPLIED_TRANSITION`: User applied the
```
eugenekoh, line 104: Change caused by PR

service/sql/model.go
```diff
@@ -2416,16 +2438,24 @@ type StatementParameterListItem struct {
 // successful, and statement closed; result no longer available for fetch
 type StatementState string
 
+// user canceled; can come from explicit cancel call, or timeout with
```
eugenekoh, line 2441: This is an example change where the enum descriptions are now correctly generated. Previously, //openapi library did not parse - or + bullets.

service/billing/model.go
pietern: Any idea why some of the other enums in this diff do contain newlines and this one doesn't?

eugenekoh: I was looking into this as well. Its a weird behavior. Let me check what's happening here

PR #613, opened 2023-09-13
Add details to APIError

hectorcast-db:
## Changes
Databricks API can return error details in case of errors. In some cases, users need to access such details to be able to solve the issue. This is the case for the errors of type ErrorInfo used by the UC catalog.
This PR adds the necessary fields to the APIError to Unmarshal the ErrorDetails and to expose the details of type ErrorInfo to the users.

## Tests

- [X] `make test` passing
- [X] `make fmt` applied

### DIFF
--- a/apierr/errors.go
+++ b/apierr/errors.go
@@ -29,10 +29,15 @@ var (
 	}
 )
 
+const (
+	errorInfoType string = "type.googleapis.com/google.rpc.ErrorInfo"
+)
+
 // APIErrorBody maps "proper" databricks rest api errors to a struct
 type APIErrorBody struct {
-	ErrorCode string `json:"error_code,omitempty"`
-	Message   string `json:"message,omitempty"`
+	ErrorCode string        `json:"error_code,omitempty"`
+	Message   string        `json:"message,omitempty"`
+	Details   []ErrorDetail `json:"details,omitempty"`
 	// The following two are for scim api only
 	// for RFC 7644 Section 3.7.3 https://tools.ietf.org/html/rfc7644#section-3.7.3
 	ScimDetail string `json:"detail,omitempty"`
@@ -41,11 +46,19 @@ type APIErrorBody struct {
 	API12Error string `json:"error,omitempty"`
 }
 
+type ErrorDetail struct {
+	Type     string            `json:"@type,omitempty"`
+	Reason   string            `json:"reason,omitempty"`
+	Domain   string            `json:"domain,omitempty"`
+	Metadata map[string]string `json:"metadata,omitempty"`
+}
+
 // APIError is a generic struct for an api error on databricks
 type APIError struct {
 	ErrorCode  string
 	Message    string
 	StatusCode int
+	Details    []ErrorDetail
 }
 
 // Error returns error message string instead of
@@ -60,6 +73,25 @@ func IsMissing(err error) bool {
 		return apiError.IsMissing()
 	}
 	return false
+}
+
+// GetErrorInfo returns all entries in the list of error details of type `ErrorInfo`.
+func GetErrorInfo(err error) []ErrorDetail {
+	return getDetailsByType(err, errorInfoType)
+}
+
+func getDetailsByType(err error, errorDetailType string) []ErrorDetail {
+	var apiError *APIError
+	if !errors.As(err, &apiError) {
+		return nil
+	}
+	filteredDetails := []ErrorDetail{}
+	for _, detail := range apiError.Details {
+		if errorDetailType == detail.Type {
+			filteredDetails = append(filteredDetails, detail)
+		}
+	}
+	return filteredDetails
 }
 
 // IsMissing tells if it is missing resource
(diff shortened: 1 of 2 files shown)


apierr/errors.go
```diff
@@ -41,11 +48,19 @@ type APIErrorBody struct {
 	API12Error string `json:"error,omitempty"`
 }
 
+type ErrorDetails struct {
+	Type     ErrorType         `json:"@type,omitempty"`
+	Reason   string            `json:"reason,omitempty"`
+	Domain   string            `json:"domain,omitempty"`
+	Medatada map[string]string `json:"metadata,omitempty"`
```
deryasusm-db, line 55: Metadata

apierr/errors.go
```diff
@@ -29,10 +29,17 @@ var (
 	}
 )
 
+type ErrorType string
```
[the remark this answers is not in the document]
pietern: In hindsight, I don't think we should define a type for this.

We won't know all the different instances here, so we can keep it a string and compare against known values.

apierr/errors.go
```diff
@@ -62,6 +75,26 @@ func IsMissing(err error) bool {
...
+		filteredDetails := []ErrorDetail{}
+		for _, detail := range apiError.Details {
+			if errrorDetailsType == detail.Type {
+				filteredDetails = append(filteredDetails, detail)
+			}
+		}
+		return filteredDetails
+	}
+	return nil
+
```
pietern, line 95: Superfluous newline.

apierr/errors.go
```diff
@@ -62,6 +75,26 @@ func IsMissing(err error) bool {
 	return false
 }
 
+// GetErrorInfo returns the details of the error
```
pietern, line 78:
```suggestion
// GetErrorInfo returns all entries in the list of error details of type `ErrorInfo`.
```

client/client_test.go
```diff
@@ -111,6 +111,62 @@ func TestSimpleRequestFailsAPIError(t *testing.T) {
 	assert.EqualError(t, err, "nope")
 }
 
+func TestETag(t *testing.T) {
+	reason := "some_reason"
+	domain := "a_domain"
+	eTag := "sample_etag"
+	c := &DatabricksClient{
+		Config: config.NewMockConfig(func(r *http.Request) error {
+			r.Header.Add("Authenticated", "yes")
```
pietern, line 120: This can be removed for this test.

client/client_test.go
```diff
@@ -111,6 +111,62 @@ func TestSimpleRequestFailsAPIError(t *testing.T) {
...
+			}, nil
+		}),
+		rateLimiter: rate.NewLimiter(rate.Inf, 1),
+	}
+	err := c.Do(context.Background(), "GET", "/a/b", map[string]string{
+		"e": "f",
+	}, map[string]string{
+		"c": "d",
+	}, nil)
+	apiError := err.(*apierr.APIError)
```
pietern, line 159: No need to type cast. You should be able to pass the returned error directly into `GetErrorInfo`.

apierr/errors.go
pietern: This ends up in the public docs so we need to be precise.

client/client_test.go
```diff
@@ -111,6 +111,62 @@ func TestSimpleRequestFailsAPIError(t *testing.T) {
...
+	details := apierr.GetErrorInfo(apiError)
+	assert.Equal(t, 1, len(details))
+	errorDetails := details[0]
```
pietern, line 162: `info`? Because we're getting info.

Btw, is it possible to have multiple entries of the same type in the the details array? Theoretically, yes, but perhaps the backend doesn't allow this.

apierr/errors.go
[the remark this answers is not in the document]
hectorcast-db: Updated

client/client_test.go
hectorcast-db: Good question. No idea. Let me ask.

pietern (commented):
Please expand the PR title and summary with more information before merging.

I would like the commit history to include a reference to `ErrorDetail` being a type and `ErrorInfo` being one particular implementation with the fields we use in `ErrorDetail`, and its type set as it is. Future readers of this commit or history will be thankful.

apierr/errors.go
```diff
@@ -62,6 +75,25 @@ func IsMissing(err error) bool {
 	return false
 }
 
+// GetErrorInfo returns all entries in the list of error details of type `ErrorInfo`.
+func GetErrorInfo(err error) []ErrorDetail {
+	return getDetailsByType(err, errorInfoType)
+}
+
+func getDetailsByType(err error, errrorDetailsType string) []ErrorDetail {
```
pietern, line 83:
```suggestion
func getDetailsByType(err error, errorDetailType string) []ErrorDetail {
```

apierr/errors.go
```diff
@@ -62,6 +75,25 @@ func IsMissing(err error) bool {
...
+	var apiError *APIError
+	if errors.As(err, &apiError) {
```
pietern, line 85: If you negate this you can return early and reduce the nesting on the rest.

pietern (approved):
Leaving stamp so you can merge after addressing issues.

client/client_test.go
hectorcast-db: Yes. We can get multiple.

PR #615, opened 2023-09-13
Add ForceSendFields and a custom Marshaller

hectorcast-db:
## Changes
Add ForceSendFields and a custom JSON Marshaller to support adding default Go fields in the API calls.

## Tests
- [X] `make test` passing
- [X] `make fmt` applied


hectorcast-db:
Integration tests
https://github.com/databricks/eng-dev-ecosystem/actions/runs/6171182559

json/marshal.go
```diff
@@ -0,0 +1,193 @@
+package marshal
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+const FORCE_SEND_FIELD_NAME = "ForceSendFields"
```
pietern, line 10: Unexport + snake case

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Marshal returns a JSON encoding of the given object. Included fields:
+// - non-empty value
+// - a basic type whose field's name is present in forceSendFields
+// Our templates always populate either omitempty or the '-' tag.
+// For simplicity, we assume one of those cases.
+func Marshal(object interface{}) ([]byte, error) {
```
pietern, line 17: `interface{}` -> `any`

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+
+	dataMap, err := structAsMap(object)
```
pietern, line 19: This is not recursive at the moment. Can we make it recursive?

* Upside: we don't need a custom MarshalJSON function on all our types anymore
* Downside: custom MarshalJSON is not honored/cannot be used

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return json.Marshal(dataMap)
+}
+
+// Converts the object to a map
+func structAsMap(object interface{}) (map[string]interface{}, error) {
+	result := make(map[string]interface{})
+
+	// If the object is nil or a pointer to nil, don't do anything
+	if object == nil || (reflect.ValueOf(object).Kind() == reflect.Ptr && reflect.ValueOf(object).IsNil()) {
```
pietern, line 32: Isn't the second case already taken care of by the first?

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+
+	forceSendFields, err := getForceSendFields(object)
+
+	if err != nil {
+		return nil, err
+	}
+
+	includeFields := make(map[string]bool)
+	for _, field := range forceSendFields {
+		includeFields[field] = true
```
pietern, line 44: The map only contains `forceSendFields`, thus the `getForceSendFields` function can return a map.

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+		fieldValue := value.Field(i)
+		fieldStruct := objectType.Field(i)
+
+		// Anonymous fields should be marshalled using the same JSON, and then merged into the same map
+		if fieldStruct.Anonymous && fieldValue.IsValid() {
+			dataMap2, err := structAsMap(fieldValue.Interface())
+			if err != nil {
+				return nil, err
+			}
+			result = mergeMaps(result, dataMap2)
```
pietern, line 67: The embedded struct gets precedence; it shouldn't.

pietern: Naming: `result` and `dataMap2` are not great.

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, v := range parts {
+		switch v {
+		case "omitempty":
+			jsonTag.omitempty = true
+		case "string":
+			jsonTag.asString = true
+		}
+	}
+	return jsonTag, nil
+}
```
pietern, line 121: Does this come from `encoding/json`? If not, is it different?

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Determines wether a field should be indluded or not
```
pietern, line 123: Typo

json/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Merges two maps. Overrides duplicated elements, which is fine because json.Unmarshal also
+// does it, and a JSON should not have duplicated entries.
+func mergeMaps(m1 map[string]interface{}, m2 map[string]interface{}) map[string]interface{} {
+	merged := make(map[string]interface{})
+	for k, v := range m1 {
+		merged[k] = v
+	}
+	for key, value := range m2 {
+		merged[key] = value
```
pietern, line 162: Naming is different.

pietern (commented):
Only looked at `marshal.go` so far.

The package directory is `json` but the package name is `marshal`. They should match.

json/marshal.go
hectorcast-db: The downside is a bit more tricky than it seems. 
* If you don't honor MarshalJSON and do it yourself, you need to suddenly care about many more types. E.g., an array of Jobs needs to be iterated and recursively called for each element. Currently, we call json.Marshall on the array and json figures out it needs to use the custom marshaller on each element.
* All calls to json.Marhsal need to use our library. If  a struct has a dependency on the SDK, it will need to use our marshaler, even if the struct itself has nothing to do with it. 
  * We need to be sure to support many more cases to ensure everything works.
  * Users will have to call our custom marshaller directly too if they build something which requires to marshal elements.

json/marshal.go
hectorcast-db: It is not. This example is false for the first, but true for the second:
```
var v interface{}
v = (*string)(nil)
```
[Go Playground](https://go.dev/play/p/apyPa4CNZ6)
[GoLang doc](https://go.dev/doc/faq#nil_error)

This seems to be the case for nil anonymous fields, since they escape the first condition.

json/marshal.go
hectorcast-db: `encoding/json` does the same in a different way:
https://github.com/golang/go/blob/master/src/encoding/json/tags.go
That particular class is not exported, so either we reimplement or copy the file. Since I had already reimplemented it before looking at it, I left my version to avoid licensing discussions.

.codegen/model.go.tmpl
```diff
@@ -15,6 +16,16 @@ type {{.PascalName}} struct {
     {{- range .Fields}}
     {{.Comment "    // " 80}}
     {{.PascalName}} {{if .IsOptionalObject}}*{{end}}{{template "type" .Entity}} `{{template "field-tag" . }}`{{end}}
+
+	ForceSendFields []string `json:"-"`
```
andrewnester, line 20: Since it's auto generated code and I want force send some fields, how do I define which fields to send?
It's not very clear from this PR at the moment

marshal/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+			jsonTag.omitempty = true
+		case "string":
+			jsonTag.asString = true
+		}
+	}
+	return jsonTag, nil
+}
+
+// Determines wether a field should be included or not
+func includeField(tag jsonTag, value reflect.Value, field reflect.StructField, mustInclude map[string]bool) bool {
```
andrewnester, line 119: instead of passing 2 arguments you can directly pass the value of `mustInclude[field.Name]` and call it `forceInclude`

.codegen/model.go.tmpl
hectorcast-db: Example for Clusters:
```
clusterDetails := compute.ClusterSpec{
		NumWorkers:   0,   
... //other details
                ForceSendFields: []string{"NumWorkers"} 
}
```
We can add a comment to it. How does this sound?
```
// Fields to be send as part of the request even if they are empty or have the default value. Null elements will still not be send.
```

.codegen/model.go.tmpl
```diff
@@ -15,6 +16,16 @@ type {{.PascalName}} struct {
...
+}
+
+func (s *{{.PascalName}}) UnmarshalJSON(b []byte) error {
+	return marshal.Unmarshal(b, s)
+}
+
+func (s {{.PascalName}}) MarshalJSON() ([]byte, error) {
+	return marshal.Marshal(s)
```
nfx, line 28: We only need to add a custom `MarshalJSON` with `ForceSendFields` for types that:

1. are used in requests
2. have optional primitive fields (int, bool, string, ..). 

This PR adds it for literally **every response and request type**, which hurts the UX and makes documentation less readable.

image

`UnmarshalJSON` is not necessary, as it's used for decoding responses.

nfx (dismissed):

.codegen/model.go.tmpl
hectorcast-db: SDK users are interacting with the SDK names. For them the JSON names are almost an implementation details (although they see them in errors.)

marshal/marshal.go
```diff
@@ -0,0 +1,193 @@
...
+	if tag.ignore {
+		return false
+	}
+	if !tag.omitempty {
+		return true
+	}
+	return (isBasicType(value.Type()) && mustInclude[field.Name]) || !isEmptyValue(value)
+}
+
+func isEmptyValue(v reflect.Value) bool {
```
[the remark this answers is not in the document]
hectorcast-db: It gives different results. This function is like the one added in the encoding/json [package](https://github.com/golang/go/blob/a278550c40ef3f01a5fcbef43414dc49009201f8/src/encoding/json/encode.go#L306). Added a comment for reference.

.codegen/model.go.tmpl
hectorcast-db: (2) Removed from structs without fields of basic types. 

As for (1), it is not only requests, but also structs which are subfields of a request (at any level).
I will check if we can easily determine that.

marshal/marshal.go
```diff
@@ -0,0 +1,192 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+
+	"golang.org/x/exp/maps"
+)
+
+const force_send_field_name = "ForceSendFields"
+
+// Marshal returns a JSON encoding of the given object. Included fields:
```
mgyucht, line 14: Let's document the behavior for embedded struct/anonymous fields w.r.t. ForceSendFields defined in each of them.

marshal/marshal.go
```diff
@@ -0,0 +1,192 @@
...
+	if !field.IsValid() {
+		return nil, nil
+	}
+	forceSendFields, ok := field.Interface().([]string)
+	if !ok {
+		return nil, fmt.Errorf("invalid type for %s field", force_send_field_name)
+	}
+	includeFields := make(map[string]bool)
+	for _, field := range forceSendFields {
+		includeFields[field] = true
```
mgyucht, line 173: If `field` is not defined on `v`, return error

marshal/unmarshal.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		} else {
+			return errors.New("cannot set field")
+		}
+	}
+	return nil
+}
+
+// Returns the list of fields found in the json
+func getFieldsInJson(data []byte, v any) ([]string, error) {
```
mgyucht, line 65: Let's try to combine this with unmarshalInternal

openapi/code/entity.go
```diff
@@ -80,6 +80,22 @@ type Entity struct {
 	Schema *openapi.Schema
 }
 
+// Whether the Entity contains a basic GoLang type which is not required
+func (e *Entity) ForceSendFields() bool {
```
mgyucht, line 84: ShouldIncludeForceSendFields

.codegen/model.go.tmpl
```diff
@@ -15,7 +16,19 @@ type {{.PascalName}} struct {
     {{- range .Fields}}
     {{.Comment "    // " 80}}
     {{.PascalName}} {{if .IsOptionalObject}}*{{end}}{{template "type" .Entity}} `{{template "field-tag" . }}`{{end}}
+
+	{{if .ForceSendFields}} ForceSendFields []string `json:"-"` {{end}}
+}
+
+{{if and .Fields .ForceSendFields}}
```
mgyucht, line 23:
```suggestion
{{if .ForceSendFields}}
```

marshal/unmarshal.go
```diff
@@ -0,0 +1,181 @@
...
+		// reflex.GetFieldByName may return the field for an anonymous field
+		// if the extending struct doesn't have the field
+		field := getFieldByName(v, force_send_field_name)
+		if !field.IsValid() {
+			return nil
+		}
+		// A Value can be changed only if it is
+		// addressable and was not obtained by
+		// the use of unexported struct fields.
+		if field.CanSet() {
```
mgyucht, line 44:
```go
if !field.CanSet() {
  return error
}
if field.Kind() != reflect.Slice {
  error
}
```

marshal/unmarshal.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return err
+	}
+
+	return recordOriginalFields(data, v)
+}
+
+func recordOriginalFields(data []byte, v any) error {
+
+	value := reflect.ValueOf(v)
```
mgyucht, line 32: if value.Kind != Ptr {
  return
}

mgyucht (approved):

marshal/marshal.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return false
+}
+
+// Merges two maps. Overrides duplicated elements, which is fine because json.Unmarshal also
+// does it, and a JSON should not have duplicated entries.
+func mergeMaps(m1 map[string]any, m2 map[string]any) map[string]any {
+	merged := make(map[string]any)
+	maps.Copy(merged, m2)
+	maps.Copy(merged, m1)
```
pietern, line 159: This order means m1 has precedence over m2. From the argument order of the function I would have expected the reverse, i.e. for m2 to have precedence, because it is the second argument.

marshal/marshal.go
```diff
@@ -0,0 +1,212 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"golang.org/x/exp/maps"
+)
+
+const force_send_field_name = "ForceSendFields"
```
pietern, line 12: Style nit, please use camelCase for unexported consts.

marshal/marshal.go
```diff
@@ -0,0 +1,212 @@
...
+
+// Marshal returns a JSON encoding of the given object. Included fields:
+// - non-empty value
+// - a basic type whose field's name is present in forceSendFields
+// Embedded structs are still considered a separate struct. ForceSendFields
+// in an embedded struct only impact the fields of the embedded struct.
+// Conversely, an embedded struct is not impacted by the ForceSendFields
+// of the struct containing it.
+func Marshal(object any) ([]byte, error) {
+
```
pietern, line 22: Superfluous newline.

marshal/marshal.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func getFieldNames(v any) map[string]bool {
+	result := map[string]bool{}
+	value := reflect.ValueOf(v)
+	value = reflect.Indirect(value)
+	objectType := value.Type()
+
+	for i := 0; i < value.NumField(); i++ {
+		name := objectType.Field(i).Name
```
pietern, line 208: Should this skip anonymous fields?

marshal/marshal.go
```diff
@@ -0,0 +1,212 @@
...
+
+	for i := 0; i < value.NumField(); i++ {
+		jsonTag := objectType.Field(i).Tag.Get("json")
+		tag, err := parseJSONTag(jsonTag)
+		if err != nil {
+			return nil, err
+		}
+
+		fieldValue := value.Field(i)
+		fieldType := objectType.Field(i)
```
pietern, line 58: Doing this every time is pretty expensive. Please think about caching a bunch of this information per type (e.g. the list of field names, the tags, etc are all static per type).

marshal/types.go
```diff
@@ -0,0 +1,36 @@
+package marshal
+
+import "reflect"
+
+var BASIC_TYPES = map[reflect.Kind]bool{
```
pietern, line 5: Please camelCase.

marshal/unmarshal.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func setForceSendFields(v any, presentFields []string) error {
+
+	value := reflect.ValueOf(v)
+	if value.Kind() != reflect.Ptr {
+		return nil
+	}
+
+	// reflex.GetFieldByName may return the field for an anonymous field
```
pietern, line 37: Typo

marshal/unmarshal.go
```diff
@@ -0,0 +1,141 @@
...
+	pointer := field.Addr().Interface()
+	err := json.Unmarshal(value, pointer)
+	if err == nil {
+		return nil
+	}
+	// We use godss/yaml to convert YAML into JSON.
+	// This library stops converting when it finds a custom Marshaller,
+	// Since strings in YAML may not have quotes, they won't be added.
+	// So we add them manually
+	return json.Unmarshal([]byte(`"`+string(value)+`"`), pointer)
```
pietern, line 140: Why is this happening exactly? I believe the library first converts YAML to JSON without looking at types.

marshal/marshal.go
hectorcast-db: We can skip anything which is not a basic type.
Anonymous fields do have a name internally (which corresponds to the struct type name), so they are not really special here.

marshal/unmarshal.go
hectorcast-db: The library does YAML -> map[string]interface{} -> internal conversions -> map[string]interface{} -> JSON
The internal conversions include using reflect on the struct to determine the type and do some changes (e.g., add the quotes if the type if string). However, for structs which implement the UnmarshalJSON, the conversion step is skipped.

marshal/marshal.go
hectorcast-db: I added some caching and did a performance test. 
Dataset: Repeat single test 1.000.000 times 
No cache: ~10.5s
Cache: ~8s

pietern: Does this also include the stuff to get the index of the force send fields field, etc? The comment was not about this line alone, about caching type info in general.

marshal/unmarshal.go
pietern: The "may not have quotes" is what worries me. If it does, this will fail.

Can you tell with certainty if this is the case or not?

hectorcast-db: It it has quotes, lines 132 will have worked. 
We do have YAMLs with strings in between quotes which cover this case.

marshal/marshal.go
```diff
@@ -0,0 +1,213 @@
...
+		fieldValue := value.Field(i)
+		fieldStruct := fields[i]
+
+		// Anonymous fields should be marshalled using the same JSON, and then merged into the same map
+		if fieldStruct.Anonymous && fieldValue.IsValid() {
+			anonymousFieldResult, err := structAsMap(fieldValue.Interface())
+			if err != nil {
+				return nil, err
+			}
+			result = mergeMaps(result, anonymousFieldResult)
```
pietern, line 67: This now means the anonymous struct takes precedence over the parent, which I think is the reverse of what should actually happen. Can you add a test for this? I suspect we won't hit this path in the SDK because we don't do composition and we don't have field name aliasing.

marshal/reflect_cache.go
```diff
@@ -0,0 +1,37 @@
...
+
+var nameCache = map[reflect.Type]string{}
+
+type cachedType struct {
+	reflect.StructField
+	JsonTag string
+}
+
+func getTypeFields(structType reflect.Type) *[]cachedType {
+	if res, ok := typeCache[structType]; ok {
```
pietern, line 15: No need to use a pointer type for the map value or return type. A slice copy is always by reference.

marshal/reflect_cache.go
```diff
@@ -0,0 +1,37 @@
...
+	for i := 0; i < structType.NumField(); i++ {
+		field := structType.Field(i)
+		res = append(res, cachedType{
+			StructField: field,
+			JsonTag:     field.Tag.Get("json"),
+		})
+	}
+	typeCache[structType] = &res
+	return &res
+}
```
pietern, line 28: The type cache can/should contain more information; the helpers in `marshal.go` loop over fields to find one by name. That can instead be a map lookup on some cached struct.

marshal/unmarshal.go
```diff
@@ -0,0 +1,142 @@
...
+
+	objectType := derefer.Type()
+
+	foundFields := []string{}
+	fields := *getTypeFields(objectType)
+
+	for i := 0; i < derefer.NumField(); i++ {
+		field := derefer.Field(i)
+		fieldType := field.Type()
+		jsonTag := fields[i].JsonTag
```
pietern, line 93: We can cache the `jsonTag` object such that parsing/`getJsonName` isn't needed on every call.

marshal/unmarshal.go
pietern: Thanks, I suspect that's good enough.

marshal/marshal.go
hectorcast-db: I have the order. In any case, there should not be duplicates. json.marshal doesn't support duplicates. 
If we want to do something, we may want to throw an error instead.

marshal/reflect_cache.go
hectorcast-db: While I have a reusable method, we only search for the `forceSendFieldName` field. (I can change the mtehod name/signature). And caching that did not improve performance.

marshal/unmarshal.go
hectorcast-db: Added separately. Adding it inside the same cache means that each call to getTypeFields needs to handle errors, which made the code less readable for no real reason.

marshal/marshal.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	fields := getTypeFields(objectType)
+
+	for i := 0; i < value.NumField(); i++ {
+		jsonTag := fields[i].JsonTag
+		tag, err := parseJSONTag(jsonTag)
+		if err != nil {
+			return nil, err
+		}
```
pietern, line 55: There are 2 things happening here that I think need to be changed:
* The JSON tag is stored as a string on the field and is then parsed again (albeit cached). Why not store it on the cached type field in the first place?
* Iteration happens on `value.NumField()` as opposed to `len(fields)`. Instead, the cached type field can include its field index in the struct type, so that you can range over `fields` directly.

marshal/marshal.go
```diff
@@ -0,0 +1,190 @@
...
+func getFieldByName(v any, fieldName string) reflect.Value {
+	value := reflect.ValueOf(v)
+	value = reflect.Indirect(value)
+	objectType := value.Type()
+
+	for i, field := range getTypeFields(objectType) {
+		name := field.Name
+		if name == fieldName {
+			return value.Field(i)
+		}
```
pietern, line 174: This iteration scheme also assumes that the field index is equal to the index in the returned slice.

The cached type field should include its field index on the type, so that we can use that directly.

marshal/reflect_cache.go
```diff
@@ -0,0 +1,68 @@
...
+func getTypeFields(structType reflect.Type) []cachedType {
+	if res, ok := typeCache[structType]; ok {
+		return res
+	}
+	res := []cachedType{}
+	for i := 0; i < structType.NumField(); i++ {
+		field := structType.Field(i)
+		res = append(res, cachedType{
+			StructField: field,
+			JsonTag:     field.Tag.Get("json"),
```
pietern, line 28: Re: my other comment; we can parse the JSON tag here, and include the field index.

marshal/reflect_cache.go
pietern: Thanks for checking.

marshal/unmarshal.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// reflex.GetFieldByName may return the field from an anonymous field
+	// if the extending struct doesn't have the field.
+	field := getFieldByName(v, forceSendFieldName)
+	if !field.IsValid() {
+		return nil
+	}
+
+	if !field.CanSet() || field.Kind() != reflect.Slice {
+		return errors.New("cannot set field")
```
pietern, line 43: When can this happen?

marshal/unmarshal.go
pietern: Why would it need to handle errors?

hectorcast-db: Parsing the JSON tags can throw errors, unlike the reflection actions. The auxiliary functions which just want the list of fields and their names don't return errors currently. If we add parsing to the same function, then the function can return an error and those auxiliary functions also need to handle it. 
We can assume that the cache is populated in advance (which should happen) and then ignore errors inside the auxiliary functions, but seems something easy to forget in the future.

marshal/unmarshal.go
hectorcast-db: It should not happen, or happen by mistake:
* [Can't happen] The field is private (then it won't match the name)
* [Shold not happen] The field is not addressable
* [Mistake] Field has the wrong type

marshal/reflect_cache.go
```diff
@@ -0,0 +1,68 @@
...
+		})
+	}
+	typeCache[structType] = res
```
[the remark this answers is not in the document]
pietern: Access needs to be protected as well. You can do this (at the top of the function, prior to access):
```
mutexType.Lock()
defer mutexType.Unlock()
```

marshal/unmarshal.go
```diff
@@ -60,26 +60,24 @@ func unmarshalInternal(data []byte, v any) error {
...
 
-	for i := 0; i < derefer.NumField(); i++ {
-		field := derefer.Field(i)
-		fieldType := field.Type()
-		jsonTag := fields[i].JsonTag
-
-		if fields[i].Anonymous {
-			setField(field, data)
+	for _, field := range getTypeFields(objectType) {
+		if field.JsonTag.ignore {
```
pietern, line 70: Can't we opt to not include the type in the cached type slice to begin with, if it's ignored anyway?

marshal/marshal.go
```diff
@@ -45,20 +45,17 @@ func structAsMap(object any) (map[string]any, error) {
...
+	for _, field := range getTypeFields(objectType) {
+		if field.JsonTag.ignore {
+			continue
 		}
+		fieldIndex := field.IndexInStruct
+		tag := field.JsonTag
 
-		fieldValue := value.Field(i)
-		fieldStruct := fields[i]
+		fieldValue := value.Field(fieldIndex)
```
pietern, line 55: fieldIndex can be inlined, not used anywhere else.

marshal/unmarshal.go
hectorcast-db: We need it when searching for the ForceSendField name (which is ignored, but uses the same cache)

marshal/types_test.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+type basicMarshalTest struct {
+	st         customStruct
+	jsonString string
+	// Compare marshal results with normal marshal
+	matchClassic bool
+	// Unmarshal may not match, since ForceSendFields will be populated during
+	// custom unmarshal process
+	matchUnmarshal          bool
```
mgyucht, line 81: Let's change this to the target unmarshal'ed datastructure. As a shorthand, we can use `nil` to mean `we expect st`.

marshal/marshal.go
```diff
@@ -0,0 +1,185 @@
...
+const forceSendFieldName = "ForceSendFields"
+
+// Marshal returns a JSON encoding of the given object. Included fields:
+// - non-empty value
+// - a basic type whose field's name is present in forceSendFields
+// Embedded structs are still considered a separate struct. ForceSendFields
+// in an embedded struct only impact the fields of the embedded struct.
+// Conversely, an embedded struct is not impacted by the ForceSendFields
+// of the struct containing it.
+func Marshal(object any) ([]byte, error) {
```
mgyucht, line 20: Add comment about Interface/pointer values that are not serialized, even if ForceSendFields includes their field name.

marshal/types_maptointerface_test.go
```diff
@@ -0,0 +1,49 @@
...
+		basicMarshalTest{
+			st: customStruct{
+				MapToInt: element,
+			},
+			jsonString:     `{"maptoint":{"key":"3"},"childfs":{},"childnofs":{}}`,
+			matchClassic:   true,
+			matchUnmarshal: false,
+		},
+	)
+	//assert.Fail(t, "S")
```
mgyucht, line 34: Remove this?

mgyucht (approved):
Couple last points, otherwise LGTM.

.codegen/model.go.tmpl
```diff
@@ -15,7 +16,19 @@ type {{.PascalName}} struct {
...
+
+	{{if .ShouldIncludeForceSendFields}} ForceSendFields []string `json:"-"` {{end}}
+}
+
+{{if .ShouldIncludeForceSendFields}}
+func (s *{{.PascalName}}) UnmarshalJSON(b []byte) error {
+	return marshal.Unmarshal(b, s)
 }
+
+func (s {{.PascalName}}) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
hectorcast-db: The way that json.Marshal searches for the methods using reflection requires this. Changing it makes json.Marshal not call our methods.

marshal/composite_test.go
```diff
@@ -0,0 +1,171 @@
...
+	assert.NoError(t, err, "error while executing custom marshal")
+	compareJSON(t, `{"field2":2}`, string(res))
+
+	var reconstruct noSendFieldChild
+	err = json.Unmarshal(res, &reconstruct)
+	assert.NoError(t, err, "error while unmarshaling")
+	st.CompositeSecondParent = &CompositeSecondParent{}
+	assert.Equal(t, st, reconstruct)
+
+}
```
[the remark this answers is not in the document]
hectorcast-db: Tests the same behaviour in a different struct. The struct is different in that `noSendFieldChild` has only one anonymous field (vs 2 in the compositeChild). Go works a bit different when there is only one anonymous field. It will surface ForceSendFields to the containing struct. So I added a test to ensure everything is still correct.

marshal/types_test.go
```diff
@@ -0,0 +1,129 @@
...
+	var i1 interface{}
+	var i2 interface{}
+	err := json.Unmarshal([]byte(json1), &i1)
+	assert.NoError(t, err, "error while unmarshalling")
+	err = json.Unmarshal([]byte(json2), &i2)
+	assert.NoError(t, err, "error while unmarshalling")
+	if !reflect.DeepEqual(i1, i2) {
+		message := fmt.Sprintf("result doesn't match expected string:\nwant :%v\ngot: %v", i1, i2)
+		assert.Fail(t, message)
+	}
```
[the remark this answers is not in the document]
hectorcast-db: It didn't at some point. But it does now. Updated

marshal/unmarshal.go
```diff
@@ -0,0 +1,120 @@
...
+// the struct contains it. Only anotates basic types in the ForceSendFields.
+func Unmarshal(data []byte, v any) error {
+	if len(data) == 0 {
+		return nil
+	}
+	// create a new element if the pointer is nul
+	if v == nil {
+		objectType := reflect.ValueOf(v).Type()
+		v = reflect.New(objectType).Interface()
+	}
```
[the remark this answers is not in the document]
hectorcast-db: This was necessary at some point, but  not anymore. Removing.

marshal/types_test.go
hectorcast-db: Changed

marshal/marshal.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"golang.org/x/exp/maps"
+)
+
+const forceSendFieldName = "ForceSendFields"
+
+// Marshal returns a JSON encoding of the given object. Included fields:
+// - non-empty value
+// - a basic type whose field's name is present in forceSendFields
+// Interfaces and pointers are not serialized even if their field name appears in ForceSendFields
```
mgyucht, line 16:
```suggestion
// Nil interfaces, nil pointers, and nil or empty maps and slices are not serialized even if their field name appears in ForceSendFields
```

PR #623, opened 2023-09-19
Fixed code generation of primitive types

tanmay-db:
## Changes

It is not possible to generate primitive type variables currently. This change fixes that. Once done for Go SDK will be followed up for other SDKs.

## Tests

Ran: `deco openapi generate-sdk go --openapi-spec 09a7fa63d9ae243e5407941f200960ca14d48b07` and SDK generation is successful. 
- [ ] `make test` passing
- [ ] `make fmt` applied


service/ml/model_registry_usage_test.go
```diff
@@ -58,20 +58,20 @@ func ExampleModelRegistryAPI_CreateComment_modelVersionComments() {
 
 }
 
-func ExampleModelRegistryAPI_CreateModel_models() {
```
tanmay-db, line 61: These changes are unrelated but have come up few times before. These should be harmless.

pietern (approved):
The generated changes all stem from non-determinism in the generator.

cc @mgyucht

mgyucht:
Yeah, probably something like map iteration order randomized by go runtime yada yada yada... we should fix that.

PR #635, opened 2023-10-02
Add support for template libraries

mgyucht:
## Changes
This PR adds support in the codegen utility for template libraries. These are templates that are reused across multiple stages of code generation. For example, in the Java SDK, we need to compute the type of a field/parameter/return value multiple times in different phases of the SDK. This PR allows us to define these templates a single time and reuse them across multiple files. 

## Tests
Added to the Java SDK and regenerated the SDK.

### DIFF
--- a/openapi/roll/tool_test.go
+++ b/openapi/roll/tool_test.go
@@ -60,7 +60,7 @@ func TestRegenerateExamples(t *testing.T) {
 	target := "../.."
 	pass := render.NewPass(target, s.ServicesExamples(), map[string]string{
 		".codegen/examples_test.go.tmpl": "service/{{.Package}}/{{.SnakeName}}_usage_test.go",
-	})
+	}, []string{})
 	err = pass.Run(ctx)
 	assert.NoError(t, err)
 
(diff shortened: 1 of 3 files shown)


openapi/roll/tool_test.go
```diff
@@ -60,7 +60,7 @@ func TestRegenerateExamples(t *testing.T) {
 	target := "../.."
 	pass := render.NewPass(target, s.ServicesExamples(), map[string]string{
 		".codegen/examples_test.go.tmpl": "service/{{.Package}}/{{.SnakeName}}_usage_test.go",
-	})
+	}, []string{})
```
tanmay-db, line 63: Unrelated to the changes but do we plan to have the above test or we probably should remove it?
`"universe/bazel-bin/openapi/all-internal.json"`

`````
func TestOptimize(t *testing.T) {
	t.Skip()
	ctx := context.Background()
	home, _ := os.UserHomeDir()
	batch, err := code.NewFromFile(ctx, filepath.Join(home,
		"universe/bazel-bin/openapi/all-internal.json"))
	require.NoError(t, err)
	s, err := NewSuite("../../internal")
	require.NoError(t, err)
	err = s.OptimizeWithApiSpec(batch)
	require.NoError(t, err)
}
`````

mgyucht: That test is useful for debugging issues with example optimization. A developer can temporarily remove the skip at the start to easily run the example optimization.

PR #644, opened 2023-10-10
Added log statement if OIDC response is non 200

tanmay-db:
## Changes

Right now we throw the error string: `"databricks OAuth is not supported for this host"`  without information on response status.

## Tests

NA
- [ ] `make test` passing
- [ ] `make fmt` applied


pietern (approved):
It would also be good to get the first 100s of bytes from the response body, if it is set.

mgyucht (approved):
This seems good to me. To address @pietern's concern, we can add logging at the HTTP client level by implementing our own RoundTripper that logs requests and responses at debug or trace level. We may want to unify that with how it is done in client.go today too.

PR #659, opened 2023-10-17
Enforce running `az login --service-principal` on nightly runs

nfx:
This PR fixes the bug in the integration test, where the integration test environment is prematurely cleaned up before getting the relevant fixture values. This led to this test being permanently skipped and never run in the test environments.


pietern (approved):
Great catch!

PR #661, opened 2023-10-18
Added CanSetRequiredFieldsFromJson method for OpenAPI generator

andrewnester:
## Changes
Added CanSetRequiredFieldsFromJson method for OpenAPI generator

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- ~[ ] relevant integration tests applied~

### DIFF
--- a/openapi/code/method.go
+++ b/openapi/code/method.go
@@ -179,6 +179,30 @@ func (m *Method) IsJsonOnly() bool {
 	return m.Operation.JsonOnly
 }
 
+func (m *Method) CanSetRequiredFieldsFromJson() bool {
+	// cmethod supports only JSON input
+	if m.IsJsonOnly() {
+		return true
+	}
+
+	if m.Request == nil {
+		return false
+	}
+
+	// if not all required fields are primitive, then fields should be provided in JSON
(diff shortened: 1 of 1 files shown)


openapi/code/method.go
```diff
@@ -179,6 +179,30 @@ func (m *Method) IsJsonOnly() bool {
 	return m.Operation.JsonOnly
 }
 
+func (m *Method) CanSetRequiredFieldsFromJson() bool {
+	// cmethod supports only JSON input
```
pietern, line 183: cmethod typo?

openapi/code/method.go
```diff
@@ -179,6 +179,30 @@ func (m *Method) IsJsonOnly() bool {
 	return m.Operation.JsonOnly
 }
 
+func (m *Method) CanSetRequiredFieldsFromJson() bool {
```
pietern, line 182: Could you add a comment to very clearly state what we expect to hold here?

All branches below in isolation I understand, but in aggregate can't tell what the implication is of this condition being either true or false.

pietern (approved):

PR #665, opened 2023-10-19
Add telemetry for CI/CD platform to useragent

shreyas-goenka:
## Changes
This PR adds CI/CD provider infromation to the HTTP user agent. This is useful to track how often and from which CI/CD platforms our tools are being run from. 
 
## Tests
Unit tests and manually

For manual testing I ran a simple script in Go that list's all jobs in a workspace and inspected the user agent string.
**Case** **github-actions:**
command: 
```
GITHUB_ACTIONS=true go run main.go
```

User Agent:
```
User-Agent: unknown/0.0.0 databricks-sdk-go/0.23.0 go/1.21.0 os/darwin sdk-feature/pagination auth/pat cicd/github
```

**Case** **google-cloud-build:**
command:
```
PROJECT_ID=1 BUILD_ID=2 PROJECT_NUMBER=3 LOCATION=4  go run main.go
```

User agent:
```
User-Agent: unknown/0.0.0 databricks-sdk-go/0.23.0 go/1.21.0 os/darwin sdk-feature/pagination auth/pat cicd/google-cloud-build
```


useragent/cicd.go
```diff
@@ -0,0 +1,57 @@
...
+	{"gitlab", []envVar{{"GITLAB_CI", "true"}}},
+	{"jenkins", []envVar{{"JENKINS_URL", ""}}},
+	{"azure-devops", []envVar{{"TF_BUILD", "True"}}},
+	{"circle", []envVar{{"CIRCLECI", "true"}}},
+	{"travis", []envVar{{"TRAVIS", "true"}}},
+	{"bitbucket", []envVar{{"BITBUCKET_BUILD_NUMBER", ""}}},
+	{"google-cloud-build", []envVar{{"PROJECT_ID", ""}, {"BUILD_ID", ""}, {"PROJECT_NUMBER", ""}, {"LOCATION", ""}}},
+	{"aws-code-build", []envVar{{"CODEBUILD_BUILD_ARN", ""}}},
+	{"tf-cloud", []envVar{{"TFC_RUN_ID", ""}}},
+}
```
tanmay-db, line 32: We should have a method that returns the mutable variable since it's easy for some other method (down the line) to change the values.

I think we should abstract out the list to another file 
Have a method that returns this list instead of modifying this mutable variable.

client/client.go
```diff
@@ -389,6 +389,19 @@ func (c *DatabricksClient) addAuthHeaderToUserAgent(r *http.Request) error {
 	return nil
 }
 
+func (c *DatabricksClient) addCiCdProviderToUserAgent(r *http.Request) error {
```
tanmay-db, line 392: nit: decouple detection with addition.

useragent/cicd_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCiCdProviderDetect(t *testing.T) {
```
tanmay-db, line 11: Great test coverage!

useragent/cicd.go
shreyas-goenka: done

useragent/cicd_test.go
shreyas-goenka: Thanks!

useragent/cicd_test.go
```diff
@@ -0,0 +1,143 @@
...
+	t.Setenv("APPLE", "123")
+	t.Setenv("CHERRY", "000")
+	assert.False(t, cicdProvider.detect(), "should not detect when not all env vars are set")
+
+	// Set the rest of the env vars.
+	t.Setenv("BANANA", "456")
+	assert.True(t, cicdProvider.detect(), "should detect when all env vars are set")
+}
+
+func cleanupEnv(t *testing.T) {
```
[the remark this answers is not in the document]
shreyas-goenka: done

useragent/cicd.go
```diff
@@ -0,0 +1,57 @@
...
+
+// detect returns true if all env vars are set and have expected values.
+func (p cicdProvider) detect() bool {
+	for _, envVar := range p.envVars {
+		v, ok := os.LookupEnv(envVar.name)
```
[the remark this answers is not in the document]
shreyas-goenka: I'd recommend keeping the existing approach, because performance for both would be identical and this is easier to read.

Underneath the hood, maintaining a map of string to string is (almost) exactly what the `syscall` package does. Source: https://cs.opensource.google/go/go/+/master:src/syscall/env_unix.go;l=78;drc=57b4df2188388d190fa4f5385f8052c008f9c7a5

client/client.go
shreyas-goenka: Done

useragent/cicd.go
```diff
@@ -0,0 +1,72 @@
+package useragent
+
+import (
+	"os"
+	"sync"
+)
+
+type envVar struct {
+	name          string
+	expectedValue string
```
mgyucht, line 10: Nit: can we just comment here that if expectedValue is empty string, we only check whether the environment variable is defined, and if nonempty, we check that the value is equal to the provided value?

mgyucht (approved):
LGTM! One very small nit.

PR #682, opened 2023-11-07
Introduce more specific exceptions, like `errors.Is(err, databricks.ErrNotFound)`, `errors.Is(err, databricks.ErrAlreadyExists)`, `errors.Is(err, databricks.ErrBadRequest)`, `errors.Is(err, databricks.ErrPermissionDenied)`, `errors.Is(err, databricks.ErrInternal)`, and others

nfx:
## Changes
This PR enables ergonomic error handling for all Databricks API Errors, following the same principles as https://github.com/databricks/databricks-sdk-py/pull/376

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


.codegen/error_mapping.go.tmpl
```diff
@@ -0,0 +1,39 @@
...
+
+{{$builtinErrors := dict 
+    "BAD_REQUEST" "os.ErrInvalid"
+    "DEADLINE_EXCEEDED" "os.ErrDeadlineExceeded"
+    "CANCELLED" "context.Canceled"
+    "NOT_FOUND" "os.ErrNotExist"
+    "UNAUTHENTICATED" "os.ErrPermission"
+    "PERMISSION_DENIED" "os.ErrPermission"
+    "RESOURCE_CONFLICT" "os.ErrExist"
+}}
```
mgyucht, line 19: I think we should leave these out. If a client wants to adapt Databricks errors to another error library, they should wrap those errors themselves. For instance, there is ErrExist in os and in io/fs, why only extend the os version?

.codegen/error_mapping.go.tmpl
```diff
@@ -0,0 +1,39 @@
+// Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
+
+package apierr
+
+import (
+	"errors"
+    "os"
+    "context"
```
mgyucht, line 8:
```suggestion
	"os"
	"context"
```

mgyucht (approved):

PR #691, opened 2023-11-14
Ensure all generated names are safe from all languages

mgyucht:
## Changes
Currently, it is allowed to add fields to our API definition that conflict with keywords in one of the supported languages for our SDK, breaking the SDK, as the generated code is no longer valid code in that SDK's language. This PR safely escapes names by adding `_` to any name that conflicts with a generated name in one of our supported languages. 

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/openapi/code/named.go
+++ b/openapi/code/named.go
@@ -9,10 +9,46 @@ import (
 )
 
 var reservedWords = []string{
+	// Go reserved keywords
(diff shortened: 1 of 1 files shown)


mgyucht:
Not taking this approach.

PR #693, opened 2023-11-16
Paginate all SCIM list requests in the SDK

xinjiezhen-db:
## Changes
This PR incorporates two hard-coded changes for the SCIM API in the Python SDK:

startIndex starts at 1 for SCIM APIs, not 0. However, the existing .Pagination.Increment controls both the start index as well as whether the pagination is per-page or per-resource. Later, we should replace this extension with two independent OpenAPI options: `one_indexed` (defaulting to `false`) and `pagination_basis` (defaulting to resource but can be overridden to page).
If users don't specify a limit, the SDK will include a hard-coded limit of 100 resources per request. We could add this to the OpenAPI spec as an option `default_limit`, which is useful for any non-paginated APIs that later expose pagination options and allow the SDK to gracefully support those. However, we don't want to encourage folks to use this pattern: all new list APIs are required to be paginated from the start.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


mgyucht:
Continuing with this change in https://github.com/databricks/databricks-sdk-go/pull/717

PR #697, opened 2023-11-17
Allow injection of HTTP transport to enable HTTP replayer pattern

tcesnik-veza:
## Changes
Allowing to inject HTTP transport of a client through configuration, to enable use cases involving HTTP replayers: https://github.com/google/go-replayers.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -105,6 +105,8 @@ type Config struct {
 	// Number of seconds to keep retrying HTTP requests. Default is 300 (5 minutes)
 	RetryTimeoutSeconds int `name:"retry_timeout_seconds" auth:"-"`
 
+	HTTPTransport *http.Transport
+
 	Loaders []Loader
 
 	// marker for configuration resolving
--- a/config/config.go
+++ b/config/config.go
@@ -105,7 +105,7 @@ type Config struct {
 	// Number of seconds to keep retrying HTTP requests. Default is 300 (5 minutes)
 	RetryTimeoutSeconds int `name:"retry_timeout_seconds" auth:"-"`
 
-    // HTTPTransport can be overriden for unit testing and together with tooling like https://github.com/google/go-replayers
+	// HTTPTransport can be overriden for unit testing and together with tooling like https://github.com/google/go-replayers
 	HTTPTransport http.RoundTripper
 
 	Loaders []Loader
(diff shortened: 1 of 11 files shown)


config/config.go
```diff
@@ -105,6 +105,8 @@ type Config struct {
 	// Number of seconds to keep retrying HTTP requests. Default is 300 (5 minutes)
 	RetryTimeoutSeconds int `name:"retry_timeout_seconds" auth:"-"`
 
+	HTTPTransport *http.Transport
```
nfx, line 108:
```suggestion
	HTTPTransport http.RoundTripper
```

`*http.Transport` is too concrete in this place. can't do that.

nfx (changes requested):

config/config.go
nfx: @tcesnik-veza please also add a unit test for this change.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/google/s2a-go v0.1.7 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.3.2 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/stretchr/objx v0.5.0 // indirect
```
nfx, line 25: please implement the feature without any external dependencies. you don't need `mock.Mock` - you can just implement a dummy type that has a `RoundTrip(request *http.Request) (*http.Response, error)` just for unit tests.

nfx (changes requested):

go.mod
tcesnik-veza: Ok I can change this, although please note that the mock library (`github.com/stretchr/testify v1.8.4`) is already a part of the project, but for some reason its indirect dependency was not yet included. Let me know how would you like to proceed.

nfx: simplest mock is 

```
type hc func(r *http.Request) (*http.Response, error)

func (cb hc) RoundTrip(r *http.Request) (*http.Response, error) {
	return cb(r)
}
```

nfx (approved):

mgyucht (approved):

mgyucht:
@tcesnik-veza thank you for raising this! Can you please setup commit signing using a GPG key and make sure all your commits are signed? See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits for guidelines.

Issue #566, opened 2023-07-31
[FEATURE] Support for Azure AD Workload Identities


agchang:
**Problem Statement**
We are trying to deploy an Azure Databricks instance with a **workload identity**, but ran into issues when contacting the Azure Instance Metadata Service  

**Proposed Solution**

**Additional Context**
We are trying to deploy an Azure Databricks instance with a **workload identity**, but ran into issues when contacting the Azure Instance Metadata Service,

nfx:
Could you link to protocol description? We don’t really want to depend on any external liberties, as much as possible. Simply because of transitive dependency tree size.

agchang:
https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview#how-it-works

NathanNZ:
This functionality could help considerably when building secure GitHub pipelines. (https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure)

While reducing transitive dependencies is a good general practice, I believe the security benefits and velocity of using cloud identity specific libraries to support would be worth considering. From a customer-centric viewpoint it would allow your customers to take advantage of best practice security practices without having to wait months or years for cloud vendor specific functionality to be implemented. From the development side it would drastically reduce the amount of custom identity logic you'd need to maintain yourself, and the ability to provide features faster when they are released upstream.

Issue #495, opened 2023-06-13
Add ARM_USE_OIDC auth type


tanmay-db:
https://nedinthecloud.com/2022/06/08/using-oidc-authentication-with-the-azurerm-backend/
https://colinsalmcorner.com/using-oidc-with-terraform-in-github-actions/

loureirodnl:
Hi @tanmay-db
I'm not sure to understand if it's possible to authentify the databricks terraform provider using oidc like we do with terraform azurerm provider ?
I read the links you mentionned (and [issue](https://github.com/databricks/terraform-provider-databricks/issues/1627#issuecomment-1381836630)) but it's not clear for me. I'm using Terraform Cloud to deploy my resources on Azure.

```
provider "azurerm" {
  features {}

  tenant_id       = var.tenant_id
  subscription_id = var.subscription_id
}

provider "azuread" {
  tenant_id     = var.tenant_id
}

provider "databricks" {
  host                        = azurerm_databricks_workspace.databricks.workspace_url
  azure_workspace_resource_id = azurerm_databricks_workspace.databricks.id
}
```

I would expect the databricks provider to connect to databricks using the same OIDC worflow than azurerm but it's not happening.

To enable azurerm OIDC auth workflow I just had to create federated credentials to trust Terraform Cloud in Azure for the service principal and then add the following env variables to Terraform Cloud workspace:
TFC_AZURE_PROVIDER_AUTH=true
TFC_AZURE_RUN_CLIENT_ID=<CLIENT_ID>

What should I do to enable OIDC for Databricks provider ?

Thank you very much

audunsolemdal:
@loureirodnl  I dont use terraform cloud, but it is possible for regular TF for sure, it is difficult to get it to work with current options. Essentially this issue is here to get this working more slimply like in the AzureRM provider with the env var `ARM_USE_OIDC: true`

In my github actions I essentially run this:

```yaml
jobs:
  x:
   (..)
   steps:
    - name: FetchBearerToken
      shell: bash
      id: bearer
      run: |
        DATABRICKS_AAD_TOKEN=`az account get-access-token --resource 2ff814a6-3304-4ab8-85cb-cd0e6f879c1d | jq -r '.accessToken'`

        echo "[DEFAULT]
        host = ${{ inputs.workspace }}
        token = $DATABRICKS_AAD_TOKEN
        jobs-api-version = 2.1" > ~/.databrickscfg

        cat ~/.databrickscfg

[...]

Issue #321, opened 2023-03-03
Retry token refresh for Azure SPN in case of `Temporarily throttled, too many requests`


nfx:
Symptom: `inner token: token error: {"error":"invalid_request","error_description":"Temporarily throttled, too many requests"}`

Follow-up to https://github.com/databricks/terraform-provider-databricks/issues/2042#issuecomment-1452528306

cc @vmazobhc

tanmay-db:
technically, this is the same as https://github.com/databricks/databricks-sdk-go/issues/256 - we need retries

PR #699, opened 2023-11-18
Decouple HTTP retries and error mapping mechanics from `DatabricksClient` into `httpclient.ApiClient`

nfx:
## Changes

This PR decouples resiliency of HTTP client (rate limiting, retries, timeouts, debug logging) from `client.DatabricksClient` and moves it to a dedicated class, allowing to later use the same robust mechanism to refresh AAD tokens.

What's changed in `httpclient.ApiClient` compared from `client.DatabricksClient` (originated in Databricks Terraform Provider back in the early 2020):
- added native integration with `golang.org/x/oauth2` package for both using token sources in the request as well as injecting a custom `*http.Client` the context for token refreshes.
- error mapper is now a pluggable function, default implementation available
- error retry checking is now pluggable function, default implementation checks for HTTP 429 and 504
- transient error messages are now configurable per client instance
- HTTP request visitors are now configurable per client instance
- Explicit and coupled `Do(ctx context.Context, method, path string, headers map[string]string, request, response any, visitors ...func(*http.Request) error) error` became concise and flexible `func (c *ApiClient) Do(ctx context.Context, method, path string, opts ...DoOption) error`. Headers can now be specified with `httpclient.WithHeaders(headers)`, JSON deserialisation can now be specified with `httpclient.WithUnmarshal(&entity)`.
- added `httpclient.WithCaptureHeader(key, &value)` to assign headers to a string pointer. This is required to get a proper integration with Azure Container Registry Managed Identity passwordless setup.
- body can be specified as `httpclient.WithBody(any)`, that internally sets a special data field. Technically, this could have been a request visitor, but we have special logic for "debug bytes" logging, which is out of scope of this PR change.
- removed flawed trailing double-forward-slash trimming (`//`), as it was breaking requests containing full URLs.
- Once we merge the https://github.com/databricks/databricks-sdk-go/pull/697, we can properly unit-test the implementations of Azure Client Secret and MSI flows

This is required for downstream tasks like:
- https://github.com/databricks/databricks-sdk-go/issues/240
- https://github.com/databricks/databricks-sdk-go/issues/566
- https://github.com/databricks/databricks-sdk-go/issues/495
- proper solution to https://github.com/databricks/databricks-sdk-go/issues/321 instead of hack in https://github.com/databricks/databricks-sdk-go/commit/d5e57f7cf6e9e65b2e54ccbb391fc0cfed849f06
- 
## Tests

- [x] Only one test was failing in TF provider, fixed in https://github.com/databricks/terraform-provider-databricks/pull/2957


httpclient/api_client.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+	message := err.Error()
+	// Handle transient errors for retries
+	for _, substring := range c.config.TransientErrors {
+		if strings.Contains(message, substring) {
+			logger.Debugf(ctx, "Attempting retry because of %#v", substring)
+			return true
+		}
+	}
+	// some API's recommend retries on HTTP 500, but we'll add that later
```
nkuln, line 184: where in this PR that maps HTTP error code to retryable/not retryable?
(although generally I guess it's accepted that 500 means no retries.. must be some important legacy API)

httpclient/api_client.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+func (c *ApiClient) fromResponse(r *http.Response) (responseBody, error) {
+	if r == nil {
+		return responseBody{}, fmt.Errorf("nil response")
+	}
+	if r.Request == nil {
+		return responseBody{}, fmt.Errorf("nil request")
+	}
+	streamResponse := r.Request.Header.Get("Accept") != "application/json" && r.Header.Get("Content-Type") != "application/json"
```
nkuln, line 148: maybe add a comment on why having these headers means the response is completed?

nkuln (approved):
(drive-by) Just did a quick skim through. Although looks like most of the corner case handlings (that I have questions about) are moved from a different package that is time-tested in TF provider so that should be good.

config/azure.go
```diff
@@ -81,23 +80,18 @@ func (c *Config) azureEnsureWorkspaceUrl(ctx context.Context, ahr azureHostResol
...
-	if err != nil {
-		return fmt.Errorf("cannot read: %w", err)
-	}
 	var workspaceMetadata struct {
 		Properties struct {
 			WorkspaceURL string `json:"workspaceUrl"`
 		} `json:"properties"`
 	}
-	err = json.Unmarshal(raw, &workspaceMetadata)
+	requestURL := env.ResourceManagerEndpoint + c.AzureResourceID + "?api-version=2018-04-01"
```
tanmay-db, line 88: It might be good to abstract out `"?api-version=2018-04-01"` and use a variable here instead since this could be used at other places in future.

config/azure.go
```diff
@@ -81,23 +80,18 @@ func (c *Config) azureEnsureWorkspaceUrl(ctx context.Context, ahr azureHostResol
...
+	err = httpclient.DefaultClient.Do(ctx, "GET", requestURL,
+		httpclient.WithUnmarshal(&workspaceMetadata),
+		httpclient.WithTokenSource(management),
+	)
 	if err != nil {
-		return fmt.Errorf("cannot unmarshal: %w", err)
+		return fmt.Errorf("resolve workspace: %w", err)
```
tanmay-db, line 94: Since we are throwing a different error will this break tests (if any) that checks for this substring. Also should the error be `cannot resolve workspace` instead?

httpclient/api_client.go
```diff
@@ -0,0 +1,386 @@
...
+			sb.WriteString(fmt.Sprintf(" (Error: %s)", err))
+		}
+	} else {
+		sb.WriteString(fmt.Sprintf("Error: %s", err))
+	}
+	sb.WriteString("\n")
+	if len(responseBody) > 0 {
+		sb.WriteString(c.redactedDump("< ", responseBody))
+	}
+	logger.Debugf(ctx, sb.String()) // lgtm [go/log-injection] lgtm [go/clear-text-logging]
```
tanmay-db, line 313: we should remove go links since this is public repo.

httpclient/api_client_test.go
```diff
@@ -0,0 +1,589 @@
...
+				Request:    r,
+				Body:       io.NopCloser(strings.NewReader("")),
+			}, nil
+		}),
+	})
+	err := client.Do(context.Background(), "POST", "/a", WithData(customReader{}))
+	require.ErrorContains(t, err, "cannot reset reader of type httpclient.customReader")
+}
+
+func TestRetryGetRequest(t *testing.T) {
```
tanmay-db, line 558: Nice test coverage!

httpclient/api_client.go
tanmay-db, line 148: We also need to add support for response headers but it could be done in separate PR since it's not immediately related to the changes in this PR and is more of a Feature Request.

tanmay-db (approved):
LGTM

httpclient/api_client.go
nfx: > We also need to add support for response headers

@tanmay-db it's there - `WithCaptureHeader()`

httpclient/api_client.go
nfx: @tanmay-db those are not go links - those are linter rules

config/azure.go
nfx: @tanmay-db not in this PR, unfortunately. there's `makeRequestBody`, that has the logic for query string & other things, but it has to be decoupled some day later.

httpclient/request.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+func WithVisitor(visitor func(r *http.Request) error) DoOption {
+	return DoOption{
+		in: visitor,
+	}
+}
+
+func WithData(body any) DoOption {
```
mgyucht, line 118: Suggestion: WithRequestData?

mgyucht (approved):
Overall LGTM, thanks for doing this! Couple nits and suggestions.

httpclient/api_client.go
```diff
@@ -0,0 +1,386 @@
...
+		}
+		// merge client-wide and request-specific visitors
+		visitors = append(visitors, o.in)
+	}
+	var requestBody any
+	for _, o := range opts {
+		if o.body == nil {
+			continue
+		}
+		requestBody = o.body
```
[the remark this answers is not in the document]
nfx: i'd rather mark `WithRequestData` as experimental - i suppose that there's going to be some refactoring to split JSON/query string serialization and make them request visitors internally.

httpclient/api_client.go
nfx: > WithCaptureHeader and WithData

i'll rename `WithHeader` to `WithRequestHeader` and `WithCaptureHeader` to `WithResponseHeader`

config/azure.go
nfx: TODO

nfx:
testing TF provider with https://github.com/databricks/databricks-sdk-go/pull/699/commits/c3188f42c1913f1f1fc30d8bac82fc9944418809:

```
 go get github.com/databricks/databricks-sdk-go@c3188f42c1913f1f1fc30d8bac82fc9944418809
go: upgraded github.com/databricks/databricks-sdk-go v0.25.0 => v0.25.1-0.20231124124904-c3188f42c191
go: upgraded google.golang.org/api v0.150.0 => v0.151.0
➜  terraform-provider-databricks git:(test/go-sdk-http-client) ✗ go mod tidy
go: downloading github.com/jhump/protoreflect v1.15.1
go: downloading github.com/bufbuild/protocompile v0.4.0
➜  terraform-provider-databricks git:(test/go-sdk-http-client) ✗ make test                                                                              
✓ Filling vendor folder with library code ...
✓ Linting source code with https://staticcheck.io/ ...
```

PR #705, opened 2023-11-23
Only delete git credentials created by the current test run

mgyucht:
## Changes
If `TestAccGitCredentials` runs twice in parallel, the credentials created by one test can be deleted by another run. This can happen if nightly tests run simultaneously in two Github actions. This PR changes this test to only delete credentials created by the test case itself.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/internal/git_credentials_test.go
+++ b/internal/git_credentials_test.go
@@ -11,13 +11,6 @@ import (
 func TestAccGitCredentials(t *testing.T) {
 	ctx, w := workspaceTest(t)
 
-	list, err := w.GitCredentials.ListAll(ctx)
-	require.NoError(t, err)
-	for _, v := range list {
-		err = w.GitCredentials.DeleteByCredentialId(ctx, v.CredentialId)
-		require.NoError(t, err)
-	}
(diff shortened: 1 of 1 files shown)


mgyucht:
https://docs.databricks.com/api/workspace/gitcredentials/create says that you can only have one credential per user. Yuck! If we want proper isolation, we'd have to create a new user for the test run...

internal/git_credentials_test.go
```diff
@@ -29,6 +22,10 @@ func TestAccGitCredentials(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	list, err := w.GitCredentials.ListAll(ctx)
+	require.NoError(t, err)
+	assert.True(t, len(list) >= 1)
```
tanmay-db, line 27: nit: instead of this, would it be better to check for the git provider and username we set above?

tanmay-db (approved):

mgyucht:
This is outmoded by integration test distributed locking.

PR #708, opened 2023-11-24
Port `qa.HTTPFixtures` to faster transport-level stubs

nfx:
This PR replicates the similar behavior for HTTP fixtures that required starting new HTTP server for every test, providing the similar usage patterns, but better integrated with the decoupled http client.

Start testing APIs with the following snippet 
```
client := httpclient.NewApiClient(httpclient.ClientConfig{
  Transport: fixtures.SliceTransport{},
})
```
and follow the error messages to something like this:
```
	type op struct {
		Foo string `json:"foo"`
	}
	client := httpclient.NewApiClient(httpclient.ClientConfig{
		Transport: fixtures.SliceTransport{
			{
				Method:   "POST",
				Resource: "/foo",
				ExpectedRequest: op{
					Foo: "ba..r",
				},
			},
		},
	})
	ctx := context.Background()
	err := client.Do(ctx, "POST", "/foo", httpclient.WithRequestData(map[string]any{
		"foo": "bar",
	}))
	assert.EqualError(t, err, `Post "/foo": expected request: want {
  "foo": "ba..r"
}, got {
  "foo": "bar"
}`, err.Error())
```

Usage example for stability improvements:

https://github.com/databricks/databricks-sdk-go/blob/feat/token-refreshers/config/auth_azure_msi_test.go#L20-L40


hectorcast-db (approved):

httpclient/api_client.go
```diff
@@ -171,8 +171,15 @@ func (c *ApiClient) isRetriable(ctx context.Context, err error) bool {
...
+	var skipRetryOnIO bool
+	// depend on the HTTP fixture interface to prevent any coupling
+	if skippable, ok := c.httpClient.Transport.(interface {
+		SkipRetryOnIO() bool
+	}); ok {
+		skipRetryOnIO = skippable.SkipRetryOnIO()
+	}
 	_, isIO := err.(*url.Error)
-	if isIO {
+	if isIO && !skipRetryOnIO {
```
mgyucht, line 182: Where is this used/needed? I didn't see a reference to it in the linked file.

httpclient/fixtures/map_transport.go
```diff
@@ -0,0 +1,36 @@
...
+	expectedRequest, err := bodyStub(req)
+	if err != nil {
+		return nil, fmt.Errorf("marshal: %w", err)
+	}
+	// whitespace in this string is very important for unit tests
+	stub := fmt.Sprintf(`"%s": {
+		%sResponse: XXX {
+			// fill in specific fields...
+		},
+	},`, key, expectedRequest)
```
mgyucht, line 34: Should these be less indented? Otherwise all lines aside from the first start with an extra tab.

httpclient/fixtures/slice_transport.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var Failures = SliceTransport{
+	{
+		MatchAny:     true,
+		ReuseRequest: true,
+		Status:       418,
+		Response:     `{"error_code": "NONSENSE", "message": "I'm a teapot"}`,
+	},
+}
```
mgyucht, line 15: What is this? If it is unused, we don't need to expose it; users can define their own failures per unit test.

httpclient/fixtures/slice_transport.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil, fmt.Errorf("marshal: %w", err)
+	}
+	// whitespace in this string is very important for unit tests
+	stub := fmt.Sprintf(`{
+		Method:   "%s",
+		Resource: "%s",
+		%sResponse: XXX {
+			// fill in specific fields...
+		},
+	},`, req.Method, resource, expectedRequest)
```
mgyucht, line 49: Ditto, should this be less indented?

httpclient/fixtures/slice_transport_test.go
```diff
@@ -0,0 +1,164 @@
...
+			Foo: "bar",
+		},
+		// ExpectedRequest: map[string]interface {}{"foo":"bar"},
+		Response: XXX {
+			// fill in specific fields...
+		},
+	},`)
+}
+
+func TestPostErrorStubFilled(t *testing.T) {
```
mgyucht, line 118: No error in this case?

mgyucht (commented):
Couple comments, otherwise looking good to me.

httpclient/api_client.go
nfx: if we don't do this - we keep retrying for the default 5 minutes.

httpclient/fixtures/map_transport.go
nfx: tests are having leading tabs

httpclient/fixtures/slice_transport.go
nfx: there are ~200 usages of this in TF unit tests

httpclient/fixtures/slice_transport_test.go
mgyucht: As in, why is the test case name "PostErrorStubFilled"? There's no error, so the test case name shouldn't include "Error" right?

PR #700, opened 2023-11-20
Introduce `DatabricksEnvironment` and fix Azure MSI auth from ACR, where IMDS doesn't give host environment information

nfx:
## Changes
This PR allows determining Azure Environment from a Databricks account or workspace hostname, removing the need for a separate configuration/environment variable and complexities related to Azure MSI from within ACR.

Similar functionality in Python SDK: 
- https://github.com/databricks/databricks-sdk-py/pull/390

Stacked on top of:
- https://github.com/databricks/databricks-sdk-go/pull/699

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- [ ] relevant integration tests applied

PR #709, opened 2023-11-24
Migrate Azure MSI token source to `httpclient` and add 90% test coverage

nfx:
This PR improves the stability for Azure MSI authentication by adopting the httpclient transport.

Needs two PRs merged first:
- https://github.com/databricks/databricks-sdk-go/pull/708
- https://github.com/databricks/databricks-sdk-go/pull/700


nfx:
go 1.19 doesn't support multiple errors unwrapping
image
image

config/environments.go
```diff
@@ -86,6 +86,12 @@ func (c *Config) Environment() DatabricksEnvironment {
 			if v.azureEnvironment.Name != azureEnv {
 				continue
 			}
+			if strings.HasPrefix(v.dnsZone, ".dev") || strings.HasPrefix(v.dnsZone, ".staging") {
```
mgyucht, line 89: Correct, this actually failed in our nightlies so we can just remove this as I added it in before the v0.26 release.

config/config.go
```diff
@@ -121,6 +124,8 @@ type Config struct {
 	// marker for configuration resolving
 	resolved bool
 
+	refreshClient *httpclient.ApiClient
```
mgyucht, line 127: Please add a doccomment for this field

httpclient/api_client.go
```diff
@@ -61,14 +61,6 @@ func (cfg ClientConfig) httpTransport() http.RoundTripper {
 	}
 }
 
-var DefaultClient = NewApiClient(ClientConfig{
```
mgyucht, line 64: Thanks for removing this!

config/azure.go
```diff
@@ -2,13 +2,109 @@ package config
...
+func (e *tokenError) Error() string {
+	return e.message
+}
+
+func (e *tokenError) Unwrap() error {
+	sdkErr, ok := apierr.ByStatusCode(e.err.StatusCode)
+	if ok {
+		// this is how we distinguish between bad requests and permission denies
+		return sdkErr
+	}
```
mgyucht, line 30: Am I understanding correctly that this maps responses from AAD/ARM onto our error codes? Do we depend on this somewhere?

config/azure.go
```diff
@@ -2,13 +2,109 @@ package config
...
+
+func (c *Config) hostOrEmpty(endpoint string) string {
+	parsedURL, err := url.Parse(endpoint)
+	if err != nil {
+		return ""
+	}
+	return parsedURL.Host
+}
+
+type azureActiveDirectoryErrorResponse struct {
```
mgyucht, line 58: Can you link to docs for reference where this structure and the corresponding one for ARM responses are defined?

config/azure.go
```diff
@@ -2,13 +2,109 @@ package config
...
+
+func (c *Config) mapAzureActiveDirectoryError(defaultErr *httpclient.HttpError) error {
+	var aadError azureActiveDirectoryErrorResponse
+	err := json.Unmarshal([]byte(defaultErr.Message), &aadError)
+	if err != nil {
+		return defaultErr
+	}
+	// remove rather explicit error description, as we're adding a link
+	// in the error rendering interface
+	msg, _, ok := strings.Cut(aadError.ErrorDescription, ". Trace ID")
```
mgyucht, line 74: What information is being removed here? If there is debugging information that could be relevant to users, we may want to include it anyways?

httpclient/errors.go
```diff
@@ -40,12 +42,12 @@ func DefaultErrorMapper(ctx context.Context, resp *http.Response, body io.ReadCl
...
-	switch some := err.(type) {
-	case *HttpError:
-		if some.StatusCode == 429 {
+	var httpError *HttpError
+	if errors.As(err, &httpError) {
+		if httpError.StatusCode == 429 {
 			return true
 		}
-		if some.StatusCode == 504 {
+		if httpError.StatusCode == 504 {
```
mgyucht, line 50: QQ, not related to this change: isn't this supposed to be 503?

mgyucht (commented):
Mostly looks good, a couple comments/suggestions/questions.

config/environments.go
nfx: i think we should keep the impl for internal users

config/azure.go
nfx: that's correct - if we set a wrong AAD tenant ID, client code would expect to handle `apierr.ErrBadRequest` - no matter if it's our API or AAD's API, same for `apierr.ErrUnauthorized`.

config/azure.go
nfx: i didn't get it from docs - I've made a couple of failing requests through integration tests, looked at debug output and made the mapping. i don't even know if/where these are documented

config/azure.go
nfx: Trace ID / Correlation ID. If user is doing debug logs - it'll appear there anyway (in pretty-printed JSON). The most useful portion of info to resolve the issue is in the ErrorLink, which we appent later.

httpclient/errors.go
nfx: 504 is gateway timeout - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504, which is retriable. but we also need to retry on 503, yes.

mgyucht (approved):
After offline discussion, LGTM, please add docs for that one parameter and feel free to merge.

nfx:
tested as `github.com/databricks/databricks-sdk-go v0.26.2-0.20231201114025-d7a8895217fe` and confirm that this PR is working as expected

PR #710, opened 2023-11-25
Detect `Accept` header in `httpclient.WithResponseUnmarshal`

nfx:
Without doing so, debug logs for HTTP fixtures are not showing bodies.

when calling APIs (AAD, ARM, ACR, ...) via `httpclient.ApiClient`, it's quite redundant to set the `Accept` header every time. This PR removes the need for such redundancy.

### DIFF
--- a/httpclient/response.go
+++ b/httpclient/response.go
@@ -61,11 +61,14 @@ func WithResponseUnmarshal(response any) DoOption {
 			}
 			switch response.(type) {
 			case *bytes.Buffer, *io.ReadCloser, *[]byte:
-				r.Header.Set("Accept", "application/octet-stream")
+				// don't send Accept header for raw types, even though we have
+				// openapi/code/method.go:440#IsResponseByteStream() setting the
+				// Accept header explicitly.
+				return nil
 			default:
 				r.Header.Set("Accept", "application/json")
+				return nil
 			}
-			return nil
 		},
 		out: func(body *responseBody) error {
 			if response == nil {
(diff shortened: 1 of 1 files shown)


mgyucht (changes requested):
Can you set these headers in the fixture library itself? Like here: https://github.com/databricks/databricks-sdk-go/pull/708/files#diff-f294cebe7a06cea919481c37577fde1761dde31ea0a752aa7d8577c3381e004fR65

nfx:
@mgyucht that's the point of this PR - when calling APIs (AAD, ARM, ACR, ...) via `httpclient.ApiClient`, it's quite redundant to set the `Accept` header every time. This PR removes the need for such redundancy.

httpclient/response.go
```diff
@@ -55,6 +55,18 @@ func WithResponseHeader(key string, value *string) DoOption {
 
 func WithResponseUnmarshal(response any) DoOption {
 	return DoOption{
+		in: func(r *http.Request) error {
+			if r.Header.Get("Accept") != "" {
+				return nil
+			}
+			switch response.(type) {
+			case *bytes.Buffer, *io.ReadCloser, *[]byte:
+				r.Header.Set("Accept", "application/octet-stream")
```
mgyucht, line 64: Ok, let's just do nothing in this case then. Again, I would like to minimize coupling between Go types and request behavior where it isn't implied.

mgyucht (commented):
Let's narrow this down to only setting this header for structs (it's kind of a misnomer for the other types, because it doesn't actually unmarshal the response in that case). Please also add a doccomment explaining the new behavior here.

httpclient/response.go
nfx: fine...

mgyucht (approved):

PR #711, opened 2023-11-25
Detect `Content-Type` header in `newRequestBody` for `httpclient`

nfx:
This PR adds implicit `Content-Type` header detection to reduce boilerplate for middle-layer HTTP client code, because we already have 100% certainty about sending `application/json`, when we call `json.Marshal`.

### DIFF
--- a/httpclient/request.go
+++ b/httpclient/request.go
@@ -34,21 +34,18 @@ func newRequestBody(data any) (requestBody, error) {
 	switch v := data.(type) {
 	case io.Reader:
 		return requestBody{
-			Reader:      v,
-			ContentType: "application/octet-stream",
-			DebugBytes:  []byte("<io.Reader>"),
+			Reader:     v,
+			DebugBytes: []byte("<io.Reader>"),
 		}, nil
 	case string:
 		return requestBody{
-			Reader:      strings.NewReader(v),
-			ContentType: "application/octet-stream",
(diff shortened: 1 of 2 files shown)


mgyucht (commented):
While it does reduce boilerplate, it increases complexity by adding implicit default behavior. The boilerplate reduction is a single parameter to the Do function. I would prefer we keep this interface simple and explicit where possible to optimize for ease of understanding.

nfx:
@mgyucht without https://github.com/databricks/databricks-sdk-go/pull/710 & https://github.com/databricks/databricks-sdk-go/pull/711, we don't see correct debug logs for https://github.com/databricks/databricks-sdk-go/pull/709 and the likes.

httpclient/request.go
```diff
@@ -25,35 +25,40 @@ import (
...
 }
 
 func newRequestBody(data any) (requestBody, error) {
 	switch v := data.(type) {
 	case io.Reader:
 		return requestBody{
-			Reader:     v,
-			DebugBytes: []byte("<io.Reader>"),
+			Reader:      v,
+			ContentType: "application/octet-stream",
```
mgyucht, line 38: Let's not set ContentType for anything else then. Agreed on JSON, but we can default to no header for other request bodies.

mgyucht (approved):

PR #712, opened 2023-11-27
Handle double slash in Files API

mgyucht:
## Changes
Copy the URI-cleaning logic from the Python & Java SDKs to the Go SDK removing empty path segments from requests to the Files API. This is more conservative than what was done before, but it brings consistency between the Go SDK and the other SDKs.

## Tests
Unit test for this case.
Files API test passes.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/databricks/databricks-sdk-go/apierr"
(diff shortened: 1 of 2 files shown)


pietern (approved):
Did you pin-point where this regression was introduced?

client/client.go
```diff
@@ -108,6 +109,11 @@ func (c *DatabricksClient) Do(ctx context.Context, method, path string,
 	opts = append(opts, httpclient.WithRequestHeaders(headers))
 	opts = append(opts, httpclient.WithRequestData(request))
 	opts = append(opts, httpclient.WithResponseUnmarshal(response))
+	// Remove extra `/` from path for files API.
```
nfx, line 112: @mgyucht it's better applied in the `RequestVisitors` slice, as it'll be more forward-compatible. 

I wonder if we can trim+add a slash from end of URL and remove the slash from the beginning of the path? it'll be more logical and fit more cases

mgyucht: This is at its core a big in the files API spec. I don't want to support this in general and will fix it soon, after which I'll remove it. So consider this change to be temporary until the next OpenAPI bump.

nfx (approved):

nfx (changes requested):
This PR is still a workaround, that doesn't prevent errors on the side of OpenAPI spec. Let's do the change on code generator:

- https://github.com/databricks/databricks-sdk-go/pull/713

client/client.go
nfx: let's better do a more proactive change https://github.com/databricks/databricks-sdk-go/pull/713

PR #713, opened 2023-11-27
Trim method path suffix of redundant trailing slashes

nfx:
## Changes
Fundamentally prevent regressions like

```
PUT /api/2.0/fs/files//Volumes/main/x/y/z
> [non-JSON document of 11 bytes]. <io.Reader>
< HTTP/2.0 400 Bad Request
< {
<   "error_code": "BAD_REQUEST",
<   "message": "Invalid path"
< }
```


mgyucht (commented):
Regenerating the SDK with this change does not change the generated code. The issue is that users can specify an initial `/` in this parameter, and it is documented as `The absolute path of the file or directory in DBFS.`, where absolute paths begin with a `/`. However, if they do that, the path contains two consecutive `/`s.

PR #724, opened 2023-11-30
[WIP] allow empty string in enum set methods

mgyucht:
## Changes
WIP PR with the idea to support empty string in `Set()` methods for enums. This allows simpler setting of optional enum values, bypassing validation for optional fields in other structures.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/.codegen/model.go.tmpl
+++ b/.codegen/model.go.tmpl
@@ -45,7 +45,7 @@ func (f *{{.PascalName}}) String() string {
 // Set raw string value and validate it against allowed values
 func (f *{{.PascalName}}) Set(v string) error {
     switch v {
-    case {{range $i, $e := .Enum }}{{if $i}}, {{end}}`{{.Content}}`{{end}}:
+    case ``{{range $i, $e := .Enum }}, `{{.Content}}`{{end}}:
         *f = {{.PascalName}}(v)
         return nil
     default:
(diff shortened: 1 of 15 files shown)


.codegen/model.go.tmpl
```diff
@@ -45,7 +45,7 @@ func (f *{{.PascalName}}) String() string {
 // Set raw string value and validate it against allowed values
 func (f *{{.PascalName}}) Set(v string) error {
     switch v {
-    case {{range $i, $e := .Enum }}{{if $i}}, {{end}}`{{.Content}}`{{end}}:
+    case ``{{range $i, $e := .Enum }}, `{{.Content}}`{{end}}:
```
nfx, line 48: why do we want to allow this?.. can't it be fixed on the platform?..

nfx (changes requested):

.codegen/model.go.tmpl
mgyucht: Yeah, I marked it WIP because I'm not sure whether we actually want this or not.

The idea is that optional enum fields are represented as empty strings in structures. When you know the value you want to set an enum field to, you can simply assign it. However, you can't just say `req.EnumField.Set(x)` without checking if `x` is empty. However, an enum may be optional in some contexts and required in others, because optional or not is not a type-level decision, it is a field-level one. So here I propose adding it just in the `Set()` method for all enums to support optional enum fields, basically that a user can use empty string in the `Set()` method to indicate that the field shouldn't be sent in the request without introducing an empty string enum value (which we definitely don't want).

PR #728, opened 2023-12-04
Use per-config refresh context for OAuth tokens

nfx:
## Changes
This PR uses per-config refresh context for all oauth2 tokens, so that it's separate from the context coming in the first request (as before)

## Tests
- [x] `make test` passing
- [x] `make fmt` applied
- [x] azure env
- [x] gcp env

### DIFF
--- a/httpclient/response.go
+++ b/httpclient/response.go
@@ -38,6 +38,8 @@ func newResponseBody(data any, header http.Header, statusCode int, status string
 			ReadCloser: io.NopCloser(bytes.NewReader(v)),
 			DebugBytes: v,
 			Header:     header,
+			StatusCode: statusCode,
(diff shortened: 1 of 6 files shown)


httpclient/response.go
```diff
@@ -38,6 +38,8 @@ func newResponseBody(data any, header http.Header, statusCode int, status string
 			ReadCloser: io.NopCloser(bytes.NewReader(v)),
 			DebugBytes: v,
 			Header:     header,
+			StatusCode: statusCode,
+			Status:     status,
```
mgyucht, line 42: cc @hectorcast-db who was going to fix this.

mgyucht (approved):

PR #730, opened 2023-12-04
Added 100% unit test coverage for `oauth-m2m`

nfx:
This PR adds coverage of client-secret oauth for databricks on aws

### DIFF
--- a/config/auth_m2m.go
+++ b/config/auth_m2m.go
@@ -2,15 +2,14 @@ package config
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
 
+	"github.com/databricks/databricks-sdk-go/httpclient"
 	"github.com/databricks/databricks-sdk-go/logger"
 )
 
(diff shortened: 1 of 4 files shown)


config/auth_m2m.go
```diff
@@ -55,26 +54,11 @@ func oidcEndpoints(ctx context.Context, cfg *Config) (*oauthAuthorizationServer,
...
-	if err != nil {
-		return nil, fmt.Errorf("read .well-known: %w", err)
-	}
 	var oauthEndpoints oauthAuthorizationServer
-	err = json.Unmarshal(raw, &oauthEndpoints)
+	err := cfg.refreshClient.Do(ctx, "GET", oidc,
+		httpclient.WithResponseUnmarshal(&oauthEndpoints))
 	if err != nil {
-		return nil, fmt.Errorf("parse .well-known: %w", err)
+		return nil, errOAuthNotSupported
```
tanmay-db, line 61: Just want to check, is there a reason we want to be less descriptive on type of error message we show to user? 

Before we use to show empty body / response status / err  but now we will only show `databricks OAuth is not supported for this host`.

tanmay-db: While `parse .well-known:` doesn't make sense to be thrown on user end. It would be good to throw the `err` along with error message that OAuth is not supported. Stamping to unblock.

tanmay-db (approved):

PR #733, opened 2023-12-08
Retry on Status Code 503

mgyucht:
## Changes
The SDK currently retries requests on 429 and some specific transient errors. This PR adds retries on 503 so that more transient errors, such as "No webapps are available handle your request. Please try again later.", are retried.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


apierr/errors.go
```diff
@@ -104,6 +104,11 @@ func (apiError *APIError) IsRetriable(ctx context.Context) bool {
 	if apiError.IsTooManyRequests() {
 		return true
 	}
+	// Retry when the API is unavailable. This includes "No webapps are available
+	// to handle your request. Please try again later." from the API gateway.
+	if apiError.StatusCode == http.StatusServiceUnavailable {
+		return true
+	}
```
pietern, line 111: This could be too broad. Could we make it explicit through the transient error messages?

fjakobs: Generally retrying on 5xx is the right thing to do

pietern: Not for all of them. E.g. 504 Gateway Timeout for a POST must not automatically retry. If the unavailability means the request may or may not have been seen by a backend it's tricky. And we don't differentiate by HTTP method here.

shreyas-goenka (approved):

apierr/errors.go
mgyucht: I don't think retrying on 504 (Gateway Timeout) is safe, as that request could have been handled by the backend service. However, 503 (Service Unavailable) should be safe, since that implies that the request never was processed by a backend service.

pietern: Good point. Mozilla agrees: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After.

pietern (approved):

apierr/errors.go
mgyucht: Yup, that was my reference too :)

PR #736, opened 2023-12-09
Consolidate usage of named sorting

nfx:
## Changes

`github.com/databricks/databricks-sdk-go/httpclient/fixtures` stub generator uses Named.PascalName() method to come up with the best possible field name for generated copy-pastable stubs, though, when this library is attempted to be used together with github.com/spf13/viper, we immediately get the following error related to a change in golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e as:

```
.../entity.go:185:32: type func(a *Field, b *Field) bool of func(a, b *Field) bool {…} does not match inferred type
...                        func(a *Field, b *Field) int for func(a E, b E) int,
```
because github.com/spf13/viper v0.17+ transitively depends on golang.org/x/exp v0.0.0-20230905200255-921286631fa9

This PR keeps the sorting behavior but removes dependency on code that changed in the experimental slices package

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/openapi/code/entity.go
+++ b/openapi/code/entity.go
@@ -2,9 +2,9 @@ package code
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/databricks/databricks-sdk-go/openapi"
-	"golang.org/x/exp/slices"
 )
 
 // Field of a Type (Entity)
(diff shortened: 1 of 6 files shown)


nfx:
`go get github.com/databricks/databricks-sdk-go@610f7ae29ac996d91c3a642655651d1cfc5234b9` solves the issue

hectorcast-db (approved):

mgyucht (approved):

PR #739, opened 2023-12-12
Skip auth, config loading, and rate limits when fixture transport is present in the config

nfx:
## Changes
This PR simplifies downstream testing by speeding up the test runs and removing the redundant need to configure auth

At the moment, we have to supply _some_ auth config for every test:

```go
&databricks.Config{
  Host:               "ABC",
  Token:              "...",
  RateLimitPerSecond: 1000, // TODO: use no-op rate limiter in case of fixture transports to get 200x test speedup
  HTTPTransport:      fixtures.MappingTransport{ ... },
}
```

## Tests
- [ ] `make test` passing
- [ ] `make fmt` applied


hectorcast-db (approved):

nfx:
verified to work in https://github.com/databrickslabs/watchdog/commit/fb4951e075993b716c103108eca4b42bed0ee16c, running one integration test suite

nfx:
integration tests passed

PR #740, opened 2023-12-12
Extract API interfaces for all services and generate mock clients

andrewnester:
## Changes
Now SDK extracts an interfaces for all service APIs, generates mocks for them using mockery and provides a way to create an instance of `mocks.NewWorkspaceClient` which can be used for testing purposes

Usage outside of SDK as follows

```
	w := mocks.NewWorkspaceClient(t, &config.Config{})
	mockW := mocks.GetMockWorkspaceAPI(w)
	mockW.On(
		"GetStatusByPath",
		context.Background(),
		"/Shared/Notebook/notebook",
	).Return(&workspace.ObjectInfo{
		ObjectType: "NOTEBOOK",
		Language:   "PYTHON",
	}, nil)

	mockW.On(
		"Download",
		context.Background(),
		"/Shared/Notebook/notebook",
	).Return(io.NopCloser(strings.NewReader("Hello from notebook")), nil)
	
	doTasksOnWorkspace(context.Background(), w)
```

## Tests

- [x] `make test` passing
- [x] `make fmt` applied
- ~[ ] relevant integration tests applied~

Fixes https://github.com/databricks/cli/issues/113


pietern (commented):
To support better mocking across all services, is it an option to add the `Client()` method to all `XyzService` interfaces and implementations? Then you can keep using `WithImpl` and `Impl()` and have similar (the same?) behavior.

.codegen/interface.go.tmpl
```diff
@@ -19,6 +19,9 @@ type {{.PascalName}}Service interface {
 	{{- end}}
     {{.PascalName}}(ctx context.Context{{if .Request}}, request {{.Request.PascalName}}{{end}}) {{ template "response-type" . }}
 	{{end -}}
+
```
pietern, line 22: Nit: the `-}}` removes trailing whitespace, so it trims the newline in between.

Using just `{{ end }}` might work.

.codegen/interface.go.tmpl
```diff
@@ -19,6 +19,9 @@ type {{.PascalName}}Service interface {
...
+	// Returns an instance of DatabricksClient
```
pietern, line 23: I think the `[]` means godoc creates links.

```suggestion
	// Returns an instance of [client.DatabricksClient].
```

pietern (commented):
We had a brief in-person discussion about mocking the clients in the SDK. It would be ideal to include a separate package in the SDK that produces a fully mocked workspace/account client to use with https://github.com/stretchr/testify#mock-package. We could either use https://github.com/vektra/mockery or the codegen templates in this repository to generate these mocks.

.codegen/impl.go.tmpl
```diff
@@ -18,6 +18,10 @@ type {{.CamelName}}Impl struct {
 	client *client.DatabricksClient
 }
 
+func (a *{{.CamelName}}Impl) Client() client.DatabricksClientInterface {
+	return a.client
```
nfx, line 22: we don't need to do this anymore - there's HTTPTransport mocking - see so DatabricksClient doesn't really have to be mocked - we can mock out the HTTP transport - https://github.com/databrickslabs/watchdog/blob/main/scan/scan_test.go#L29-L72

and if you need direct files - see https://github.com/databrickslabs/watchdog/blob/main/scan/notebooks/content_scanner_test.go#L46-L48

nfx (changes requested):
this PR is not required, because we already can do transport mocking with fixtures.

see
- https://github.com/databrickslabs/watchdog/blob/main/scan/scan_test.go#L29-L72
- https://github.com/databrickslabs/watchdog/blob/main/scan/notebooks/content_scanner_test.go#L46-L48

client/client.go
```diff
@@ -78,6 +78,12 @@ func New(cfg *config.Config) (*DatabricksClient, error) {
 	}, nil
 }
 
+type DatabricksClientInterface interface {
+	Do(ctx context.Context, method, path string,
+		headers map[string]string, request, response any,
+		visitors ...func(*http.Request) error) error
+}
+
```
nfx, line 86: this is not the right interface to expose - if we do a change, then it has to be 

```
type ApiClientDoer interface {
   Do(ctx context.Context, method, path string, opts ...httpclient.DoOption) error
}
```

https://github.com/databricks/databricks-sdk-go/blob/main/httpclient/api_client.go#L100-L101

nfx (changes requested):

.codegen/api.go.tmpl
```diff
@@ -12,7 +12,92 @@ import (
...
+		"Clusters"
+		"CommandExecution"
+		"Dbfs"
+		"Libraries"
+		"StatementExecution"
+		"Workspace"
+}}
 {{range .Services}}
+
+type {{.PascalName}}APIInterface interface {
```
pietern, line 26: IMO we should call this just `Interface`, or even drop the suffix and use just `{{.PascalName}}` for brevity.

.codegen/mock_workspace_client.go.tmpl
```diff
@@ -0,0 +1,33 @@
+// Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
+
+package mocks
+
+import (
+    "github.com/databricks/databricks-sdk-go"
+	{{range .Packages}}
+	"github.com/databricks/databricks-sdk-go/mocks/service/{{.Name}}"{{end}}
```
pietern, line 8: Mix of tabs and spaces (fixed by `gofmt` but still).

mocks/mock_workspace_client.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+}
+
+func GetMockAccountAccessControlProxyAPI(c *databricks.WorkspaceClient) *iam.MockAccountAccessControlProxyAPIInterface {
+	api, ok := c.AccountAccessControlProxy.(*iam.MockAccountAccessControlProxyAPIInterface)
+	if !ok {
+		panic("Invalid API interface type")
+	}
+	return api
+}
```
pietern, line 105: No newlines between functions.

.codegen/api.go.tmpl
```diff
@@ -12,7 +12,92 @@ import (
...
+	{{if (in $extendedInterface .PascalName) }}
+		{{.CamelName}}APIUtilities
+	{{end -}}
+
+	// WithImpl could be used to override low-level API implementations for unit 
+	// testing purposes with [github.com/golang/mock] or other mocking frameworks.
+	WithImpl(impl {{.PascalName}}Service) {{.PascalName}}APIInterface
+
+	// Impl returns low-level {{.PascalName}} API implementation
+	Impl() {{.PascalName}}Service
```
pietern, line 36: Since these were useful only for mocking, we could consider removing them from the interface.

Also see https://github.com/databricks/databricks-sdk-go/tree/main/examples/mocking.

.codegen.json
```diff
@@ -11,7 +11,8 @@
     ".codegen/accounts.go.tmpl": "account_client.go",
     ".codegen/service-pkg.go.tmpl": "service/pkg.go",
     ".codegen/error_mapping.go.tmpl": "apierr/error_mapping.go",
-    ".codegen/error_alias.go.tmpl": "error_alias.go"
+    ".codegen/error_alias.go.tmpl": "error_alias.go",
+    ".codegen/mock_workspace_client.go.tmpl": "mocks/mock_workspace_client.go"
```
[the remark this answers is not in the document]
nfx: +1

.codegen/api.go.tmpl
nfx: +1

nfx (changes requested):

.codegen/api.go.tmpl
andrewnester: Just using {{.PascalName}} unfortunately conflicts with some other names defined, for example `WorkspaceConf`. Will keep it with `Interface` name

.codegen/api.go.tmpl
```diff
@@ -12,7 +12,92 @@ import (
...
+	{{end}}
+
+	{{if $hasWaiter}}
+	// Calls [{{.Service.Name}}APIInterface.{{.PascalName}}] and waits to reach {{range $i, $e := .Wait.Success}}{{if $i}} or {{end}}{{.Content}}{{end}} state
+	// 
+	// You can override the default timeout of {{.Wait.Timeout}} minutes by calling adding 
+	// retries.Timeout[{{.Wait.Poll.Response.PascalName}}](60*time.Minute) functional option.
+	//
+	// Deprecated: use [{{.Service.Name}}APIInterface.{{.PascalName}}].Get() or [{{.Service.Name}}APIInterface.{{.Wait.PascalName}}]
+	{{.PascalName}}AndWait(ctx context.Context{{if .Request}}, {{.Request.CamelName}} {{.Request.PascalName}}{{end}}, options ...retries.Option[{{.Wait.Poll.Response.PascalName}}]) (*{{.Wait.Poll.Response.PascalName}}, error)
```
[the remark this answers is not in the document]
andrewnester: This would be a breaking change in SDK, so I would prefer handling removing deprecated methods in a separate change

.codegen/api.go.tmpl
pietern: Thanks for checking!

experimental/mocks/mock_account_client.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/databricks/databricks-sdk-go/experimental/mocks/service/provisioning"
+	"github.com/databricks/databricks-sdk-go/experimental/mocks/service/settings"
+)
+
+// NewAccountClient creates new mocked version of Databricks SDK client for Accounts
+// which can be used for testing.
+func NewAccountClient(t interface {
+	mock.TestingT
+	Cleanup(func())
+}, cfg *config.Config) *databricks.AccountClient {
```
pietern, line 23: Given that the function signature is different from `databricks.NewAccountClient`, can we get rid of the `cfg` argument entirely? If users need to set it for some reason they can poke at the struct directly, and it keeps tests a little bit tighter to not have to worry about the `cfg` instance.

.gitattributes
```diff
@@ -68,55 +39,29 @@ service/pipelines/api.go linguist-generated=true
...
-service/sql/statement_execution_usage_test.go linguist-generated=true
-service/sql/warehouses_usage_test.go linguist-generated=true
 service/workspace/api.go linguist-generated=true
-service/workspace/git_credentials_usage_test.go linguist-generated=true
 service/workspace/impl.go linguist-generated=true
 service/workspace/interface.go linguist-generated=true
 service/workspace/model.go linguist-generated=true
-service/workspace/repos_usage_test.go linguist-generated=true
-service/workspace/secrets_usage_test.go linguist-generated=true
-service/workspace/workspace_usage_test.go linguist-generated=true
```
pietern, line 121: Do you know why the test ignores are gone?

If not, the change can be reverted to not mess with it, and we can deal with it elsewhere.

service/pkg.go
```diff
@@ -178,6 +178,7 @@
 package service
 
 import (
+	_ "github.com/databricks/databricks-sdk-go/experimental/mocks" // for mock support
```
pietern, line 181: Do we need this for docs?

Also, can you double check that this include doesn't bloat binary size of e.g. tests?

pietern (commented):
LGTM

Please also add a test case under `experimental/mocks` to confirm that mocking the account client and workspace client actually works. This can also serve as an example, but mostly makes sure that everything is hooked up correctly.

service/pkg.go
andrewnester: CLI when using SDK without mocks
```
andrew.nester@HFW9Y94129 cli % ls -lh ./cli                                                    
-rwxr-xr-x@ 1 andrew.nester  staff    34M Dec 27 12:50 ./cli
```

CLI when using SDK with mocks and using mocked client in tests
```
andrew.nester@HFW9Y94129 cli % ls -lh ./cli
-rwxr-xr-x@ 1 andrew.nester  staff    34M Dec 27 12:54 ./cli
```

.gitattributes
andrewnester: @pietern it was because SDK was regenerated without fix in openapi/roll code from this PR, fixed now

pietern (commented):
Regarding the code size; I originally commented in `service/pkg.go` but noticed that the mock dependencies are already pulled in via the main `databricks` package for the client constructor functions. That's why there is no code size difference.

I put this to the test and initialized a fresh module that makes only a single call and looked at the `go.mod` differences and binary size (with and without this PR).

With this PR enabled, I get the following `go.mod` diff:

```diff
diff --git a/go.mod b/go.mod
index ea260a4..4f84002 100644
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require github.com/databricks/databricks-sdk-go v0.28.1
 require (
        cloud.google.com/go/compute v1.23.3 // indirect
        cloud.google.com/go/compute/metadata v0.2.3 // indirect
+       github.com/davecgh/go-spew v1.1.1 // indirect
        github.com/felixge/httpsnoop v1.0.4 // indirect
        github.com/go-logr/logr v1.3.0 // indirect
        github.com/go-logr/stdr v1.2.2 // indirect
@@ -15,6 +16,9 @@ require (
        github.com/google/go-querystring v1.1.0 // indirect
        github.com/google/s2a-go v0.1.7 // indirect
        github.com/googleapis/enterprise-certificate-proxy v0.3.2 // indirect
+       github.com/pmezard/go-difflib v1.0.0 // indirect
+       github.com/stretchr/objx v0.5.0 // indirect
+       github.com/stretchr/testify v1.8.4 // indirect
        go.opencensus.io v0.24.0 // indirect
        go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.46.1 // indirect
        go.opentelemetry.io/otel v1.21.0 // indirect
@@ -34,6 +38,7 @@ require (
        google.golang.org/grpc v1.59.0 // indirect
        google.golang.org/protobuf v1.31.0 // indirect
        gopkg.in/ini.v1 v1.67.0 // indirect
+       gopkg.in/yaml.v3 v3.0.1 // indirect
 )
```

The code size of the resulting binary is 23M vs 20M.

I propose we put the constructors under `experimental/mocks` as well.

andrewnester:
@pietern moved the constructors under `experimental/mocks`

pietern (approved):

.mockery.yaml
```diff
@@ -0,0 +1,10 @@
+with-expecter: true
+filename: "{{.InterfaceName | snakecase}}_mock.go"
```
mgyucht, line 2: nit: should we do `mock_{{.InterfaceName | snakecase}}.go` to match the actual mock name?

experimental/mocks/mocks_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func TestMockWorkspaceClient(t *testing.T) {
+	ctx := context.Background()
+	w := NewMockWorkspaceClient(t)
+	mockClusters := w.GetMockClustersAPI()
+	mockClusters.EXPECT().ListAll(
+		ctx,
+		mock.AnythingOfType("compute.ListClustersRequest"),
+	).Return(
+		[]compute.ClusterDetails{
```
mgyucht, line 22: Nice!

.codegen/api.go.tmpl
```diff
@@ -12,7 +12,94 @@ import (
...
+	{{.Shortcut.PascalName}}(ctx context.Context{{range .Shortcut.Params}}, {{.CamelName}} {{template "type" .Entity}}{{end}}) {{ template "response-type" . }}
+	
+	{{end}}{{if and .Shortcut $hasWaiter}}
+	{{.Shortcut.PascalName}}AndWait(ctx context.Context{{range .Shortcut.Params}}, {{.CamelName}} {{template "type" .Entity}}{{end}}, options ...retries.Option[{{.Wait.Poll.Response.PascalName}}]) (*{{.Wait.Poll.Response.PascalName}}, error)
+	
+	{{end}}
+{{end -}}
+}
+
 func New{{.PascalName}}(client *client.DatabricksClient) *{{.PascalName}}API {
```
mgyucht, line 103: Should we also remove these types/constructors from the public interface at some point?

mgyucht (commented):
Mostly small comments, but overall this seems like a great change.

.codegen/mock_workspace_client.go.tmpl
```diff
@@ -0,0 +1,40 @@
...
+// NewMockWorkspaceClient creates new mocked version of Databricks SDK client for Workspaces
+// which can be used for testing.
+func NewMockWorkspaceClient(t interface {
+	mock.TestingT
+	Cleanup(func())
+}) (*MockWorkspaceClient) {
+	return &MockWorkspaceClient{
+		WorkspaceClient: &databricks.WorkspaceClient{
+			Config: nil,
+			{{range .Services}}{{if not .IsAccounts}}
```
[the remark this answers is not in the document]
andrewnester: It doesn't look as correct syntax with range + pipeline syntax as you specified

```
Error: Generate: generate code: batch: all: exec mock_workspace_client.go.tmpl: template: mock_workspace_client.go.tmpl:24:23: executing "mock_workspace_client.go.tmpl" at <not>: wrong number of args for not: want 1 got 2
```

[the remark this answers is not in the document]
mgyucht: My mistake, leave as is.

.codegen/mock_account_client.go.tmpl
```diff
@@ -0,0 +1,40 @@
...
+package mocks
+
+import (
+	"github.com/databricks/databricks-sdk-go"
+	{{range .Packages}}
+	"github.com/databricks/databricks-sdk-go/experimental/mocks/service/{{.Name}}"{{end}}
+)
+
+type MockAccountClient struct {
+	AccountClient *databricks.AccountClient
```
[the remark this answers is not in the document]
andrewnester: > That way you can use MockAccountClient exactly how you use AccountClient, without having to get the inner field storing the client.

I don't think it's possible to substitute structs like this in Go lang even if one embeds the other. 
See https://go.dev/play/p/I57OV6izp6u

```
package main

import "fmt"

type A struct {
	name string
}

type B struct {
	A
}

func acceptsA(a *A) {
	fmt.Println(a.name)
}

func main() {
	b := B{A: A{name: "b"}}
	acceptsA(&b)
}

```

Will fail with `cannot use &b (value of type *B) as *A value in argument to acceptsA` hence the having a field for Workspace/AccountClient

.codegen/api.go.tmpl
andrewnester: @mgyucht why should we?

mgyucht: This would correspond to us removing the Impl()/WithImpl() methods from the public interface. We would keep them in the SDK for sure, but they don't need to be public anymore.

mgyucht (approved):
Thanks!

PR #744, opened 2023-12-15
Improve debuggability of unexpected errors

mgyucht:
## Changes
Issues like https://github.com/databricks/terraform-provider-databricks/issues/3031 are difficult for users to debug when Databricks APIs respond with unexpected response bodies. This is often a bug in either the API (returning a non-JSON error message, for example) or from the caller (e.g. calling update but not specifying the ID of the resource to update), though it could also be due to a bug in the SDK itself. 

This PR improves the debuggability of errors returned from the SDK due to malformed responses or other failures to process the response by including the request log in the error, along with helpful text that points the user to report the issue at the Go SDK so it can be triaged and fixed.

The returned error's `Error()` method will produce a string like this:

unexpected error handling request: invalid character '<' looking for beginning of value. This is likely a bug in the Databricks SDK for Go or the underlying REST API. Please report this issue with the following debugging information to the SDK issue tracker at https://github.com/databricks/databricks-sdk-go/issues. Request log:
```
GET /a
> * Host: 
> * Accept: application/json
> * Authorization: REDACTED
> * User-Agent: unknown/0.0.0 databricks-sdk-go/0.28.0 go/1.21.3 os/darwin auth/pat
< HTTP/2.0 Internal Server Error
< <html><body>hello</body></html>
```

Note that the markdown ticks are present in the request log itself for ease of copy-pasting into Github's new issue UI.

## Tests
- [x] New unit tests cover RoundTripStringer.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -259,19 +259,46 @@ func TestNonJSONResponseIncludedInError(t *testing.T) {
 	}
 	cases := []testCase{
 		{
-			statusCode:   400,
-			status:       "Bad Request",
-			errorMessage: "unexpected error handling request: invalid character '<' looking for beginning of value. This is likely a bug in the Databricks SDK for Go or the underlying REST API. Please report this issue with the following debugging information to the SDK issue tracker at https://github.com/databricks/databricks-sdk-go/issues. Request log:\nGET /a\n> * Host: \n> * Accept: application/json\n> * Authorization: Bearer token\n> * User-Agent: unknown/0.0.0 databricks-sdk-go/0.27.0 go/1.21.3 os/darwin auth/pat\n<  Bad Request\n< <html><body>hello</body></html>\n",
+			statusCode: 400,
+			status:     "Bad Request",
+			errorMessage: `unexpected error handling request: invalid character '<' looking for beginning of value. This is likely a bug in the Databricks SDK for Go or the underlying REST API. Please report this issue with the following debugging information to the SDK issue tracker at https://github.com/databricks/databricks-sdk-go/issues. Request log:
+` + "```" + `
+GET /a
+> * Host: 
+> * Accept: application/json
+> * Authorization: Bearer token
+> * User-Agent: unknown/0.0.0 databricks-sdk-go/0.27.0 go/1.21.3 os/darwin auth/pat
+< HTTP/2.0 Bad Request
+< <html><body>hello</body></html>
+` + "```",
 		},
 		{
-			statusCode:   500,
-			status:       "Internal Server Error",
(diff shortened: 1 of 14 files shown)


client/client_test.go
```diff
@@ -250,3 +250,54 @@ func TestDoRemovesDoubleSlashesFromFilesAPI(t *testing.T) {
...
+	type testCase struct {
+		statusCode   int
+		status       string
+		errorMessage string
+	}
+	cases := []testCase{
+		{
+			statusCode:   400,
+			status:       "Bad Request",
+			errorMessage: "unexpected response from server (Bad Request): invalid character '<' looking for beginning of value (original: <html><body>hello</body></html>)",
```
pietern, line 264: When we're changing this anyway, this is a good opportunity to remove the boilerplate if this is indeed an HTML body. We can check if the error is equal to `json.SyntaxError`, and if so, just dump the HTML and not the string `invalid character...`.

client/client_test.go
```diff
@@ -250,3 +250,54 @@ func TestDoRemovesDoubleSlashesFromFilesAPI(t *testing.T) {
...
+	type testCase struct {
+		statusCode   int
+		status       string
+		errorMessage string
+	}
+	cases := []testCase{
+		{
+			statusCode:   400,
+			status:       "Bad Request",
+			errorMessage: "unexpected error handling request: invalid character '<' looking for beginning of value. This is likely a bug in the Databricks SDK for Go or the underlying REST API. Please report this issue with the following debugging information to the SDK issue tracker at https://github.com/databricks/databricks-sdk-go/issues. Request log:\nGET /a\n> * Host: \n> * Accept: application/json\n> * Authorization: Bearer token\n> * User-Agent: unknown/0.0.0 databricks-sdk-go/0.27.0 go/1.21.3 os/darwin auth/pat\n<  Bad Request\n< <html><body>hello</body></html>\n",
```
mgyucht, line 264: Make this markdown-formatted so it can be copy-pasted into a Github issue.

logger/httplog/round_trip_stringer.go
```diff
@@ -0,0 +1,86 @@
...
+		if len(r.ResponseBody) == 0 && r.Err != nil {
+			sb.WriteString(fmt.Sprintf(" (Error: %s)", r.Err))
+		}
+	} else {
+		sb.WriteString(fmt.Sprintf("Error: %s", r.Err))
+	}
+	sb.WriteString("\n")
+	if r.DebugHeaders {
+		r.writeHeaders(&sb, "< ", r.Response.Header)
+	}
```
mgyucht, line 68: Add separator here (empty line)

logger/httplog/round_trip_stringer.go
```diff
@@ -0,0 +1,86 @@
...
+		sb.WriteString("?")
+		q, _ := url.QueryUnescape(request.URL.RawQuery)
+		sb.WriteString(q)
+	}
+	sb.WriteString("\n")
+	if r.DebugHeaders {
+		sb.WriteString("> * Host: ")
+		sb.WriteString(escapeNewLines(request.Host))
+		sb.WriteString("\n")
+		r.writeHeaders(&sb, "> ", request.Header)
```
mgyucht, line 48: Add newline after headers

hectorcast-db (approved):

PR #754, opened 2024-01-03
Add debug authorization header flag

mgyucht:
## Changes
DebugHeaders by default prints all headers, including the sensitive Authorization header. To allow users to print debug logs with headers but without including this sensitive header, we introduce a new flag, DebugAuthorizationHeader, which is disabled by default for users. This way, enabling debug headers does not automatically leak the token used to authenticate to our REST API.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -26,12 +26,13 @@ type RequestVisitor func(*http.Request) error
 type ClientConfig struct {
 	Visitors []RequestVisitor
 
-	RetryTimeout       time.Duration
-	HTTPTimeout        time.Duration
-	InsecureSkipVerify bool
-	DebugHeaders       bool
-	DebugTruncateBytes int
-	RateLimitPerSecond int
+	RetryTimeout             time.Duration
+	HTTPTimeout              time.Duration
+	InsecureSkipVerify       bool
+	DebugHeaders             bool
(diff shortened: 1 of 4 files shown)


httpclient/api_client.go
```diff
@@ -257,8 +258,8 @@ func (c *ApiClient) recordRequestLog(
 		RequestBody:              requestBody,
 		ResponseBody:             responseBody,
 		DebugHeaders:             c.config.DebugHeaders,
+		DebugAuthorizationHeader: c.config.DebugAuthorizationHeader,
```
nfx: you're missing the `X-Databricks-Azure-SP-Management-Token` and `X-Databricks-GCP-SA-Access-Token` headers in https://github.com/databricks/databricks-sdk-go/blob/main/logger/httplog/round_trip_stringer.go#L33-L35

nfx (changes requested):

httpclient/api_client.go
mgyucht: I think you reviewed a stale version but I added this in the last commit

tanmay-db (approved):
Since by default we prevent printing of token, I think this makes user experience more safe. LGTM

PR #755, opened 2024-01-04
Update SDK to OpenAPI spec + Fix tests

tanmay-db:
## Changes

SDK generation and fixing tests by passing a request when calling ListAll method

## Tests

Successfully generated + Integration tests passed except 2 tests in aws-prod-acct due to `Credentials limit exceeded`
- [ ] `make test` passing
- [ ] `make fmt` applied


internal/catalog_test.go
```diff
@@ -212,7 +212,7 @@ func TestUcAccStorageCredentialsOnAws(t *testing.T) {
 	require.NoError(t, err)
 	assert.NotEqual(t, "", byName.Id)
 
-	all, err := w.StorageCredentials.ListAll(ctx)
+	all, err := w.StorageCredentials.ListAll(ctx, catalog.ListStorageCredentialsRequest{})
```
tanmay-db, line 215: [Discussion] 

This is a breaking change (as golang doesn't support optional parameters) so we will have to mention this explicitly in our doc or revert the change for not introducing this. 

We could do this change since Go SDK isn't GA yet but it won't be a good user experience i.e sending empty request.

Other non breaking alternatives are: 
* Make ListAll as variadic on `ListExternalLocationsRequest`. Not sure if this is supported in SDK generator + if there are other different types of parameters in future for this method, then there might be more difficulties in maintenance. 
````
func ListAll(ctx context.Context, requests ...ListExternalLocationsRequest) ([]ExternalLocationInfo, error) {
}
````

* Have a wrapper over ListAll -- but this is something we would have to maintain externally. 

Both of these are something that we could do to not have breakage. This type of problem is specific to golang and could arise for other methods as well so we would need a solution that will work for other such cases as well without extra overhead. 

What do you think? @miles @hectorcast-db

tanmay-db: + @pietern for visibility as CLI release is blocked on GoSDK release.

tanmay-db: If we decided to not revert the open api change: Java SDK: https://github.com/databricks/databricks-sdk-java/pull/204. There are some issues with Python SDK -- will take a look at them tomorrow.

hectorcast-db: Right now, since it's not GA we could go forward. List APIs should have pagination, so this is a good change.
But indeed, in general, adding an optional parameter to an API should not be a breaking change.

hectorcast-db: + @andrewnester Blocking issue for the release

mgyucht (commented):
Need to make sure .gitattributes is not regressing here.

mgyucht (approved):

PR #756, opened 2024-01-05
Handle empty response for errors

mgyucht:
## Changes
Credential deletion fails with 409 and no response body if the credential is in use with a running workspace. This leads to some race conditions because workspace deletion is asynchronous, so attempting to delete a credential while a workspace is running will ultimately fail. This PR tolerates empty error bodies. Note that `Unwrap()` for APIError will delegate this to the appropriate status code-based error, so users can still use `errors.Is` to examine errors with no body.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/apierr/errors.go
+++ b/apierr/errors.go
@@ -168,6 +168,11 @@ func GetAPIError(ctx context.Context, resp common.ResponseWrapper) error {
 }
 
 func parseErrorFromResponse(resp *http.Response, requestBody, responseBody []byte) *APIError {
+	if len(responseBody) == 0 {
+		return &APIError{
+			StatusCode: resp.StatusCode,
+		}
+	}
 	// try to read in nicely formatted API error response
 	var errorBody APIErrorBody
 	err := json.Unmarshal(responseBody, &errorBody)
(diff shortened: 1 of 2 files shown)


apierr/errors.go
```diff
@@ -168,6 +168,11 @@ func GetAPIError(ctx context.Context, resp common.ResponseWrapper) error {
 }
 
 func parseErrorFromResponse(resp *http.Response, requestBody, responseBody []byte) *APIError {
+	if len(responseBody) == 0 {
```
tanmay-db, line 171: > This leads to some race conditions because workspace deletion is asynchronous, so attempting to delete a credential while a workspace is running will ultimately fail. 

Just to confirm, with this change the credential deletion will still fail (as workspace is in use) so it doesn't prevent the race condition however we return before unmarshalling empty response body right?

tanmay-db (approved):

apierr/errors.go
mgyucht: Correct. I'm going to make another follow-up PR to address the race condition.

PR #757, opened 2024-01-05
Add utility to retry on specific errors

mgyucht:
## Changes
The SDK may return errors that users want to retry. For instance, a user that makes many SCIM operations in parallel and some fail with a Conflict may want to retry those failed requests. To simplify this process for users, this PR introduces a Retrier abstraction that allows users to configure the timeout for retries, a set of errors that should be retried, and/or a custom callback that can check whether a particular error should be retried. Example from this PR:

```go
	t.Cleanup(func() {
		err := retries.New[struct{}](retries.OnErrors(apierr.ErrResourceConflict)).Wait(ctx, func(ctx context.Context) error {
			return a.Credentials.DeleteByCredentialsId(ctx, role.CredentialsId)
		})
		require.NoError(t, err)
	})
```

I have used this to retry credential deletion in TestMwsAccWorkspaces.

## Tests
- [x] Lots of unit tests.

### DIFF
--- a/retries/retries.go
+++ b/retries/retries.go
@@ -91,22 +91,26 @@ func (et *ErrTimedOut) Unwrap() error {
 	return et.err
 }
 
+// RetryOption is a function that sets part of the retry configuration for a retrier.
 type RetryOption func(*RetryConfig)
 
+// RetryConfig is the configuration for a retrier.
 type RetryConfig struct {
-	timeout    time.Duration
-	shouldHalt func(error) bool
+	timeout     time.Duration
+	shouldRetry func(error) bool
 }
 
+// WithTimeout sets the timeout for the retrier.
 func WithTimeout(timeout time.Duration) RetryOption {
 	return func(rc *RetryConfig) {
 		rc.timeout = timeout
 	}
 }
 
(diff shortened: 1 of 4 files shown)


retries/retries.go
```diff
@@ -88,55 +95,122 @@ func (et *ErrTimedOut) Unwrap() error {
...
+func WithRetryFunc(halt func(error) bool) RetryOption {
+	return func(rc *RetryConfig) {
+		rc.shouldRetry = halt
+	}
+}
+
+// Retrier is a struct that can retry an operation until it succeeds or the timeout is reached.
+// The empty struct indicates that the retrier should run for 20 minutes and retry on any non-nil error.
+//
+// Example:
```
nfx, line 161: can you add a compilable example of this, so that it appears in docs properly? https://pkg.go.dev/github.com/databricks/databricks-sdk-go@v0.28.0/service/compute#example-ClusterPoliciesAPI.Create-ClusterPolicies

nfx (approved):
lgtm for cross-compiler.

hectorcast-db (approved):

PR #762, opened 2024-01-09
Fix TestAccUsageDownload by increasing HTTP timeout

shreyas-goenka:
## Changes
The test consistently timed out. The API for the billing download API states it can take up to multiple minutes, so I increased the HTTP timeout from 1 minute -> 5 minutes.

## Tests
The test now passes.


shreyas-goenka:
looking at the test failure

PR #765, opened 2024-01-11
Move openapi generator to deco tool

mgyucht:
## Changes
To simplify changes to the code generator, as well as to enable us to take additional dependencies in the code generator without adding to the Go SDK, we're moving the OpenAPI codegen components to the internal `deco` tool. We may later open-source an updated version of this if/when we make our OpenAPI specification public.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


openapi/code/named.go
```diff
@@ -1,347 +0,0 @@
...
-	"map", "struct", "chan", "else", "goto", "switch", "const", "fallthrough",
-	"if", "range", "type", "continue", "for", "import", "return", "var",
-	"append", "bool", "byte", "iota", "len", "make", "new", "package",
-}
-
-// Named holds common methods for identifying and describing things
-type Named struct {
-	Name        string
-	Description string
-}
```
nfx, line 22: I rely on this in a couple of tools. This change is too big. Please setup https://github.com/google/copybara to sync code as the stop-gap measure.

nfx (changes requested):
I think https://github.com/google/copybara is a better option overall.

PR #772, opened 2024-01-17
Support unmarshalling top-level ignored fields

mgyucht:
## Changes
The CLI allows users to pass a `--json` flag to provide all parameters needed for executing the request. When a request includes fields annotated with `json:"-"`, those fields are not deserialized. As a result, requests like
```
databricks account workspace-assignment update --json "{\"principal_id\":123, \"permissions\": [\"USER\"],\"workspace_id \":456}"
```
fail because the request made is
```
PUT /api/2.0/accounts/<account-id>/workspaces/0/permissionassignments/principals/0
```

To address this, we extend the unmarshaller to support unmarshalling top-level fields annotated with `json:"-"`. I'm not sure if it is possible to support unmarshalling _all_ fields with this annotation, as we call `json.Marshal` recursively. On the other hand, it isn't necessary to support at this time. The default behavior follows the `encoding/json` behavior of not including ignored fields.

## Tests
- [x] Added unit tests for UnmarshalCustom.

### DIFF
--- a/marshal/reflect_cache.go
+++ b/marshal/reflect_cache.go
@@ -4,6 +4,8 @@ import (
 	"reflect"
 	"strings"
 	"sync"
+
+	"github.com/databricks/databricks-sdk-go/openapi/code"
 )
 
 var mutexType sync.Mutex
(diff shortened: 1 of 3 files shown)


hectorcast-db (approved):
Are those messages automatically generated? For generated messages, the custom marshaller is only used if there is at least one `omitempty` field. But from the way it's done you seem to be aware of it.

If so, you will need to update the templates for CLI generation and probably add a helper function here in the Go SDK.

mgyucht:
We may not need this after all. Instead we may handle this by accepting required path parameters on the command line. Will reopen if that approach doesn't work.

PR #776, opened 2024-01-23
Support locking in integration tests

mgyucht:
### Locking Interface
Integration tests run in parallel, both within a single action runner as well as across different runs of an action (e.g. multiple people running integration tests in parallel). Some tests modify certain resources and need exclusive access to those resources during the duration of the test. As a result, when two tests that modify a shared resource run concurrently, they fail.

This PR introduces an integration test locking framework for the Go SDK & tools built on the Go SDK. In general, for Databricks integration tests, usage will look like the example in `git_credentials_test.go`:
```go

func acquireGitCredentialsLock(ctx context.Context, t *testing.T, w *databricks.WorkspaceClient) {
	me, err := w.CurrentUser.Me(ctx)
	require.NoError(t, err)
	lockable := lock.GitCredentials{
		WorkspaceHost: w.Config.Host,
		Username:      me.UserName,
	}
	_, err = lock.Acquire(ctx, lockable, lock.InTest(t))
	require.NoError(t, err)
}

func TestAccGitCredentials(t *testing.T) {
	ctx, w := workspaceTest(t)

	// skip-next-line-roll
	acquireGitCredentialsLock(ctx, t, w)
	...
}
```

Calling `lock.Acquire()` will attempt to lock a resource using a lock backend, returning a lock. When used in tests, unlocking the lock will be registered automatically as a test cleanup, so you only need to check that no error occurred while acquiring the lock. By default, the lock will be held for 60 seconds, and locks will be created in a workspace for which the serialized SDK config can be found in `LOCK_CLIENT_CONFIG` and the path to the lock directory in `LOCK_DIRECTORY`.

A lockable resource is any resource implementing the `Lockable` interface, which provides the lock ID to use. I've added several implementations of `Lockable` for cases that I know about, but any string or structure containing only string/int fields can be used as a Lockable through `lock.NewLockableImpl()`. If a string is provided, the lock ID is the string itself. If a struct is provided, the lock ID is the struct name followed by the names of each field & the value of each field, separated by semicolon. For example: `testStruct:A=hello;B=1`.

### OpenAPI Code Generator
Additionally, I've added a directive `// skip-next-line-roll` in the integration test parser. This makes it possible to skip the next statement from the codegen parsing. This should make it easier to add other code to integration tests without needing to update the OpenAPI test compiler. Just add the comment right before the statement you want to skip from example generation, e.g.
```go
func TestMwsAccAbc(t *testing.T) {
	...
	// skip-next-line-roll
	myCustomFunc()
	....
}
```

Finally, I've made a small bugfix in the comment parsing, which currently sets `hint` to the last comment before the current node, even if the current node is not immediately preceded by a comment.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


qa/lock/azure_backend.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Create an containerClient object that wraps the container's URL and a default pipeline.
+	u := fmt.Sprintf("%s/%s%s", endpoint, lockId, token)
+	blobClient, err := blockblob.NewClientWithNoCredential(u, nil)
+	if err != nil {
+		return err
+	}
+	b.blobClient = blobClient
+	fmt.Printf("Created a container client object with URL %s\n", blobClient.URL())
```
hectorcast-db, line 39: Will this print the token?

qa/lock/azure_backend.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return err
+	}
+	b.leaseClient = blobLeaseClient
+	fmt.Printf("Created a blob lease client object with lease ID %s\n", leaseId)
+
+	// Now acquire a lease on the container.
+	// You can choose to pass an empty string for proposed ID so that the service automatically assigns one for you.
+	leaseDuration := int32(math.Min(60, duration.Seconds()))
+	acquireLeaseResponse, err := blobLeaseClient.AcquireLease(context.TODO(), leaseDuration, nil)
```
hectorcast-db, line 64: Is TODO on purpose? Maybe some note on what kind of context we want here.

qa/lock/azure_backend.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Successfully uploaded contents %s:\n%s\n", contents, b.blobClient.URL())
+	return nil
+}
+
+func (b *azureBackend) RenewLock(ctx context.Context, leaseId string) error {
+	_, err := b.leaseClient.RenewLease(ctx, nil)
```
hectorcast-db, line 85: Don't we use the leaseId?

qa/lock/azure_backend.go
```diff
@@ -0,0 +1,107 @@
...
+func (b *azureBackend) AcquireLock(ctx context.Context, leaseId string, contents []byte, duration time.Duration) error {
+	blobLeaseClient, err := lease.NewBlobClient(b.blobClient, &lease.BlobClientOptions{LeaseID: &leaseId})
+	if err != nil {
+		return err
+	}
+	b.leaseClient = blobLeaseClient
+	fmt.Printf("Created a blob lease client object with lease ID %s\n", leaseId)
+
+	// Now acquire a lease on the container.
+	// You can choose to pass an empty string for proposed ID so that the service automatically assigns one for you.
```
hectorcast-db, line 62: Maybe add the comment to the AcquireLock function, since the leaseID is passed there as a parameter

qa/lock/azure_backend.go
hectorcast-db: Oh, it's part of the leaseClient in this particular implementation. Maybe add a comment.

openapi/roll/tool.go
```diff
@@ -353,6 +353,28 @@ func (s *Suite) expectExamples(file *ast.File) {
 	}
 }
 
+func (s *Suite) getCommentPrecedingStatementMap(fn *ast.FuncDecl, file *ast.File) map[int]string {
```
mgyucht, line 356: add comment here

qa/lock/core/lockable.go
```diff
@@ -0,0 +1,5 @@
+package core
+
+type Lockable interface {
```
mgyucht, line 3: comment

qa/lock/core/state.go
```diff
@@ -0,0 +1,66 @@
...
+	Lockable        string
+	LeaseId         string
+	IsInDebug       bool
+	User            string `env:"USER"`
+	GithubActions   string `env:"GITHUB_ACTIONS"`
+	GithubRef       string `env:"GITHUB_REF"`
+	GithubRunId     string `env:"GITHUB_RUN_ID"`
+	GithubRunNumber string `env:"GITHUB_RUN_NUMBER"`
+	GithubSha       string `env:"GITHUB_SHA"`
+	GithubWorkflow  string `env:"GITHUB_WORKFLOW"`
```
pietern, line 25: Is the repository encoded in one of these?

qa/lock/core/state.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+}
+
+func (l *LockState) IsValid() bool {
+	return time.Now().Before(l.Expiry)
+}
+
+func (l *LockState) Extend(duration time.Duration) {
+	l.Expiry = time.Now().Add(duration)
```
pietern, line 49: Shouldn't this also persist the extended expiry time?

qa/lock/databricks/databricks_backend.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (d *Backend) uploadLockState(ctx context.Context, contents *core.LockState, overwrite bool) error {
+	buf, err := json.Marshal(contents)
+	if err != nil {
+		return fmt.Errorf("failed to marshal contents: %w", err)
+	}
+
+	opts := []workspace.UploadOption{
+		workspace.UploadLanguage(workspace.LanguagePython),
```
pietern, line 93: Can we omit the language setting here?

qa/lock/core/state.go
pietern: Nevermind, this is done elsewhere.

Could the lock state itself be unexported?

qa/lock/lockables_test.go
```diff
@@ -0,0 +1,77 @@
+package lock
+
+import (
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/qa/lock/internal"
+)
+
+func TestGenerateBlobName(t *testing.T) {
```
pietern, line 9: This test looks like it should live in `internal` as well (only uses `GenerateBlobName`).

qa/lock/lock.go
```diff
@@ -0,0 +1,135 @@
...
+// When the lock is no longer needed, the caller should call the Unlock() method
+// on the returned lock. This will release the lock and return any errors that
+// occurred while releasing the lock. If a *testing.T is provided as a
+// LockOption, the Unlock() method will be called automatically when the test
+// completes.
+//
+// By default, the lock will be acquired for 1 minute. This can be changed by
+// passing the WithLeaseDuration() LockOption.
+//
+// By default, the lock will be acquired using the azureBackend. This can be
```
pietern, line 27: Default is different.

qa/lock/databricks/databricks_backend_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/databricks/databricks-sdk-go/apierr"
+	"github.com/databricks/databricks-sdk-go/experimental/mocks"
+	"github.com/databricks/databricks-sdk-go/qa/lock/core"
+	"github.com/databricks/databricks-sdk-go/service/workspace"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestAcquireLock_NoExistingLock(t *testing.T) {
```
mgyucht, line 17: Add test that upload failing causes error

qa/lock/lock.go
```diff
@@ -0,0 +1,135 @@
...
+	errs := make(chan error, 1)
+
+	// Renew the lock every 45 seconds until the lock is released.
+	go func() {
+		duration := opts.Backend.RefreshDuration()
+		timer := time.NewTimer(duration)
+		testTimer := time.NewTimer(opts.LeaseDuration)
+		defer close(errs)
+		for {
+			timer.Reset(duration)
```
pietern, line 83: Can this be a `time.Ticker`?

pietern (commented):
(incomplete review)

qa/lock/lock.go
```diff
@@ -0,0 +1,135 @@
...
+		time.Sleep(10 * time.Second)
+	}
+
+	// Channel to signal when the lock should be released
+	done := make(chan struct{})
+
+	// Channel to communicate errors renewing the lease or deleting the blob
+	errs := make(chan error, 1)
+
+	// Renew the lock every 45 seconds until the lock is released.
```
mgyucht, line 76: fix comment here and in this file.

qa/lock/lockables_test.go
```diff
@@ -0,0 +1,77 @@
+package lock
+
+import (
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/qa/lock/internal"
+)
+
```
mgyucht, line 8: move to internal

hectorcast-db (approved):

qa/lock/databricks/databricks_backend.go
mgyucht: Nope, otherwise the API thinks this is a ZIP file (probably DBC archive).

qa/lock/core/state.go
mgyucht: Discussed offline, I'll add doccomments here.

PR #788, opened 2024-01-26
Retry update of catalog in test to avoid flakiness

hectorcast-db:
## Changes
Retry update of catalog in test to avoid flakiness

## Tests
- [ ] `make test` passing
- [X] `make fmt` applied

### DIFF
--- a/internal/catalog_test.go
+++ b/internal/catalog_test.go
@@ -369,13 +369,15 @@ func TestUcAccCatalogWorkspaceBindings(t *testing.T) {
 	require.NoError(t, err)
 
 	thisWorkspaceID := MustParseInt64(GetEnvOrSkipTest(t, "THIS_WORKSPACE_ID"))
-	err = retries.New[struct{}](retries.OnErrors(apierr.ErrDeadlineExceeded), retries.WithTimeout(1*time.Minute)).Wait(ctx, func(ctx context.Context) error {
+	updateFunction := func(ctx context.Context) error {
 		_, err := w.WorkspaceBindings.Update(ctx, catalog.UpdateWorkspaceBindings{
 			Name:             created.Name,
 			AssignWorkspaces: []int64{thisWorkspaceID},
 		})
 		return err
-	})
+	}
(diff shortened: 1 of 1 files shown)


internal/catalog_test.go
```diff
@@ -365,12 +369,17 @@ func TestUcAccCatalogWorkspaceBindings(t *testing.T) {
...
+	err = retries.New[struct{}](retries.OnErrors(apierr.ErrDeadlineExceeded), retries.WithTimeout(1*time.Minute)).Wait(ctx, func(ctx context.Context) error {
+		_, err := w.WorkspaceBindings.Update(ctx, catalog.UpdateWorkspaceBindings{
+			Name:             created.Name,
+			AssignWorkspaces: []int64{thisWorkspaceID},
+		})
+		return err
 	})
 	require.NoError(t, err)
 
+	require.NoError(t, err)
```
tanmay-db, line 381: this is probably not required right? We have the same on line 379 above. Stamping to unblock since rest looks good.

internal/catalog_test.go
```diff
@@ -365,12 +369,17 @@ func TestUcAccCatalogWorkspaceBindings(t *testing.T) {
 	require.NoError(t, err)
 
 	thisWorkspaceID := MustParseInt64(GetEnvOrSkipTest(t, "THIS_WORKSPACE_ID"))
-	_, err = w.WorkspaceBindings.Update(ctx, catalog.UpdateWorkspaceBindings{
-		Name:             created.Name,
-		AssignWorkspaces: []int64{thisWorkspaceID},
+	err = retries.New[struct{}](retries.OnErrors(apierr.ErrDeadlineExceeded), retries.WithTimeout(1*time.Minute)).Wait(ctx, func(ctx context.Context) error {
```
tanmay-db, line 372: nit: We can abstract this out so later we can extend the list for other other types of errors easily: `retries.OnErrors(apierr.ErrDeadlineExceeded`

tanmay-db (approved):

PR #792, opened 2024-01-31
Support creating a new workspace client from an account client

mgyucht:
## Changes
This PR introduces `func (a *AccountClient) GetWorkspaceClient(w provisioning.Workspace) (*WorkspaceClient, error)` to make it easy to get a workspace client for a given workspace. The workspace client uses a copy of the config from the account client, reusing the underlying authenticator as well. This allows all workspace clients to use the same underlying token when authenticating with OAuth, decreasing the number of unnecessary token refreshes. Additionally, it allows workspace clients to work with U2M OAuth using the account-level token for customers with Unified Login.

Use-case: https://github.com/databricks/terraform-provider-databricks/pull/3188

## Tests
New example passes. 

- [ ] `make test` passing
- [ ] `make fmt` applied


account_functions.go
```diff
@@ -0,0 +1,21 @@
...
+// GetWorkspaceClient returns a WorkspaceClient for the given deployment name.
+// The deployment name for a workspace is present in the deploymentName field of
+// the Workspace struct, which can be obtained by calling w.Workspaces.Get()
+// or w.Workspaces.List().
+//
+// The config used for the workspace is identical to that used for the account,
+// except that the host is set to the workspace host, and the account ID is
+// not set.
+func (c *AccountClient) GetWorkspaceClient(w provisioning.Workspace) (*WorkspaceClient, error) {
+	host := c.Config.Environment().DeploymentURL(w.DeploymentName)
```
pietern, line 14: Can you comment when we have a `DeploymentName` and when we have to use the `AzureResourceID`?

config/config.go
```diff
@@ -148,6 +148,62 @@ type Config struct {
...
+		refreshClient: c.refreshClient,
+		refreshCtx:    c.refreshCtx,
+		// The config does not need to be re-resolved, as we reuse all attributes
+		// from the original config.
+		resolved:  c.resolved,
+		auth:      c.auth,
+		isTesting: c.isTesting,
+	}, nil
+}
+
```
pietern, line 206: Enumerating all properties is prone to mistakes. Could you add a test confirming that all exported properties are copied? Then it will fail if someone adds a property but fails to include it in this list.

account_functions.go
mgyucht: Because the deployment name is always present when fetching a workspace, we don't actually have any hard dependency on the AzureResourceID field being set for the core functionality of the Config. In fact, because we are copying the `auth` field, we don't have to preserve any of the fields that are used for configuration. I've simply implemented this because we can compute the AzureResourceId from the Workspace.

config/config.go
mgyucht: Instead of that, I'll use reflection to copy the exported properties so that fields are not missed by default.

pietern (approved):
If we expect this to only work with OAuth, should this be enforced?

I.e. if folks try to use this with username/password it will certainly not work.

PR #799, opened 2024-02-05
Added support to select Spark version with Photon

tanmay-db:
## Changes

The instance pool id we use to run cluster tests doesn't support Photon. This started happening recently. There doesn't seem to be any change in the sorting algorithm `sort.Sort(sparkVersionsType(versions))` to get photon supported cluster so change could be on the backend or someone could have done tf apply to get a difference instance pool id. I looked at the logs of passing test previous week but since tests were passing, there was no details on what instance pool id was.

Added support for Photon selection in spark versions

## Tests

Manually ran the tests and they passed.
- [ ] `make test` passing
- [ ] `make fmt` applied


service/compute/spark_version.go
```diff
@@ -21,6 +21,7 @@ type SparkVersionRequest struct {
 	GPU             bool   `json:"gpu,omitempty" tf:"optional,default:false"`
 	Scala           string `json:"scala,omitempty" tf:"optional,default:2.12"`
 	SparkVersion    string `json:"spark_version,omitempty" tf:"optional,default:"`
+	Photon          bool   `json:"photon,omitempty" tf:"optional,default:false"`
```
tanmay-db, line 24: Since default is false, we don't need to specify it in the request.

pietern (approved):

PR #801, opened 2024-02-07
Fix stack overflow on recursive schemas

hectorcast-db:
## Changes
Fix stack overflow on recursive schemas

## Tests
Tested by following these steps:
* Remove schema causing the issue from Universe
* Generate SDK using deco tool without the fix
* Generate SDK using deco tool with fix, and verify that there is no change
* Add back the schema causing the issue from Universe
* Generate SDK using deco tool with fix, and verify new structs manually
 
- [X] `make test` passing
- [X] `make fmt` applied
- [N/A] relevant integration tests applied

### DIFF
--- a/openapi/code/package.go
+++ b/openapi/code/package.go
@@ -98,7 +98,7 @@ func (pkg *Package) ImportedPackages() (res []string) {
 	return
 }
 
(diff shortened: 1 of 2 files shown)


mgyucht (approved):
Logic LGTM. Can you add doccomments to explain the parameter?

PR #806, opened 2024-02-08
SDK Generation + Fix backwards incompatible changes

tanmay-db:
## Changes

Update SDK to OpenAPI spec and fix backwards incompatible change

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


internal/catalog_test.go
```diff
@@ -285,8 +285,8 @@ func TestUcAccMetastores(t *testing.T) {
 	assert.NotEqual(t, created.MetastoreId, currentMetastore.MetastoreId)
 
 	_, err = w.Metastores.Update(ctx, catalog.UpdateMetastore{
-		Id:   created.MetastoreId,
-		Name: RandomName("go-sdk-updated"),
```
tanmay-db, line 289: Highlight

internal/connections_test.go
```diff
@@ -30,7 +30,6 @@ func TestAccConnections(t *testing.T) {
 	})
 
 	connUpdate, err := w.Connections.Update(ctx, catalog.UpdateConnection{
-		Name:    connCreate.Name,
```
tanmay-db, line 33: we don't need this

service/catalog/connections_usage_test.go
```diff
@@ -61,7 +61,6 @@ func ExampleConnectionsAPI_Get_connections() {
 	logger.Infof(ctx, "found %v", connCreate)
 
 	connUpdate, err := w.Connections.Update(ctx, catalog.UpdateConnection{
-		Name:    connCreate.Name,
```
tanmay-db, line 64: highlight

service/catalog/metastores_usage_test.go
```diff
@@ -226,8 +226,8 @@ func ExampleMetastoresAPI_Update_metastores() {
 	logger.Infof(ctx, "found %v", created)
 
 	_, err = w.Metastores.Update(ctx, catalog.UpdateMetastore{
-		Id:   created.MetastoreId,
-		Name: fmt.Sprintf("sdk-%x", time.Now().UnixNano()),
+		Id:      created.MetastoreId,
+		NewName: fmt.Sprintf("sdk-%x", time.Now().UnixNano()),
```
tanmay-db, line 230: highlight

internal/catalog_test.go
mgyucht: Fine. This is not necessary for Metastore (since the name is not the path parameter) but better for consistency.

internal/connections_test.go
mgyucht: Shouldn't we be removing `NameArg` instead of `Name`?

service/pipelines/api.go
```diff
@@ -147,19 +147,6 @@ type PipelinesInterface interface {
 	// List updates for an active pipeline.
 	ListUpdatesByPipelineId(ctx context.Context, pipelineId string) (*ListUpdatesResponse, error)
 
-	// Reset a pipeline.
-	//
-	// Resets a pipeline.
-	Reset(ctx context.Context, resetRequest ResetRequest) (*WaitGetPipelineRunning[struct{}], error)
```
mgyucht, line 153: Was this method removed?

internal/connections_test.go
mgyucht: I guess that change didn't make it through...

mgyucht (approved):
LGTM

PR #808, opened 2024-02-08
Add `CurrentWorkspaceID()` method on WorkspaceClient

mgyucht:
## Changes
There are times when it is especially useful to get the workspace ID for the current workspace client. Currently, the workspace ID for the current workspace is exposed as a header in the SCIM Me API call. We'll expose this through a `GetWorkspaceId()` method, caching the workspace ID for the lifetime of the client.

In the future, we may add a `meta` service for exposing information about the current account/workspace. At that point, we can migrate off of this somewhat hacky approach.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


internal/workspace_client_test.go
```diff
@@ -0,0 +1,15 @@
+package internal
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUcAccWorkspaceClient_CurrentWorkspaceId(t *testing.T) {
```
mgyucht, line 9: We use UC acceptance tests because it's the only environment where the current workspace ID is configured.

workspace_functions.go
```diff
@@ -0,0 +1,23 @@
...
+	"strconv"
+
+	"github.com/databricks/databricks-sdk-go/httpclient"
+)
+
+// CurrentWorkspaceId returns the workspace ID of the workspace that this client is
+// connected to.
+func (w *WorkspaceClient) CurrentWorkspaceId(ctx context.Context) (int64, error) {
+	var workspaceIdStr string
+	err := w.apiClient.Client.Do(ctx, "GET", "/api/2.0/preview/scim/v2/Me", httpclient.WithResponseHeader("X-Databricks-Org-Id", &workspaceIdStr))
```
pietern, line 14: Is this a stable interface?

workspace_functions.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+	"strconv"
+
+	"github.com/databricks/databricks-sdk-go/httpclient"
+)
+
+// CurrentWorkspaceId returns the workspace ID of the workspace that this client is
+// connected to.
+func (w *WorkspaceClient) CurrentWorkspaceId(ctx context.Context) (int64, error) {
```
pietern, line 12: Nit: for generated code we use `Id` but for manually written code we should stick to `ID` (see https://google.github.io/styleguide/go/decisions.html#initialisms).

client/client.go
```diff
@@ -111,7 +111,7 @@ func (c *DatabricksClient) Do(ctx context.Context, method, path string,
 	if strings.HasPrefix(path, "/api/2.0/fs/files//") {
 		path = strings.Replace(path, "/api/2.0/fs/files//", "/api/2.0/fs/files/", 1)
 	}
```
pietern, line 113: Unrelated: with the recent updates to the Files API spec, can we remove this?

client/client.go
```diff
@@ -86,7 +86,7 @@ func New(cfg *config.Config) (*DatabricksClient, error) {
 
 type DatabricksClient struct {
 	Config *config.Config
-	client *httpclient.ApiClient
+	Client *httpclient.ApiClient
```
pietern, line 89: Does exporting this come with unforeseen risks? We may want to include the rationale for exporting/unexporting this in a field comment (there must have been a reason to unexport it to begin with).

workspace_functions.go
mgyucht: I'm raising internally, will send you the links where I'm asking offline.

workspace_functions.go
mgyucht: Thanks for the tip!

client/client.go
mgyucht: Yes, we'll do that as a fast follow.

client/client.go
mgyucht: I was trying to figure out a convenient way to handle this without exporting the client. Let me take another stab. The thing is that there are some complex dependencies, as the API client doesn't depend on the `config` package, but these convenience constructors do. `config` doesn't depend on `client`, which is why these convenience constructors can live here.

client/client.go
```diff
@@ -2,85 +2,36 @@ package client
...
-	"time"
 
-	"github.com/databricks/databricks-sdk-go/apierr"
 	"github.com/databricks/databricks-sdk-go/config"
 	"github.com/databricks/databricks-sdk-go/httpclient"
-	"github.com/databricks/databricks-sdk-go/useragent"
 )
 
 func New(cfg *config.Config) (*DatabricksClient, error) {
-	if skippable, ok := cfg.HTTPTransport.(interface {
```
mgyucht, line 19: Moved to config/api_client.go.

pietern (approved):

PR #812, opened 2024-02-09
Generate fields for headers in responses

hectorcast-db:
## Changes
Support header fields  in responses.

## Tests
* Generated Go SDK. A new field was added:
```
	// The name of the served model that served the request. This is useful when
	// there are multiple models behind the same endpoint with traffic split.
	ServedModelName string `json:"-" url:"-" header:"served-model-name,omitempty"`
```
* Run integration tests in PR (only AWS due to issues running workflows in EMU)
* Run files_test.go test after generation to validate changes to Content type.

- [X] `make test` passing
- [X] `make fmt` applied

### DIFF
--- a/.codegen/impl.go.tmpl
+++ b/.codegen/impl.go.tmpl
@@ -40,11 +40,7 @@ func (a *{{.Service.CamelName}}Impl) {{.PascalName}}(ctx context.Context{{if .Re
 
 {{ define "response-var" -}}
 {{ if .Response }}
-  var {{ .Response.CamelName }} {{ if .IsResponseByteStream -}}
-    io.ReadCloser
-  {{- else -}}
-	{{ template "type" .Response }}
-  {{- end }}
+  var {{ .Response.CamelName }} {{ template "type" .Response }}
 {{- end }}
 {{- end }}
 
@@ -57,11 +53,7 @@ func (a *{{.Service.CamelName}}Impl) {{.PascalName}}(ctx context.Context{{if .Re
 
 {{ define "response" -}}
   {{ if .Response -}}
-  {{ if .IsResponseByteStream -}}
-    &{{ .Response.PascalName }}{ {{ .ResponseBodyField.PascalName }}: {{ .Response.CamelName }} }
-  {{- else -}}
-    {{ if not .Response.ArrayValue }}&{{end}}{{.Response.CamelName}}
-  {{- end }}, {{end}}err
+  {{ if not .Response.ArrayValue }}&{{end}}{{.Response.CamelName}}, {{end}}err
 {{- end }}
 
 {{ define "make-header" -}}
--- a/httpclient/response.go
+++ b/httpclient/response.go
@@ -47,9 +47,10 @@ func WithResponseUnmarshal(response any) DoOption {
 				return nil
 			}
 			injectHeaders(response, body)
-			// If the field contains a "Content" field of type bytes.Buffer, write the body over there.
-			set := tryInjectContent(response, body)
-			if set {
+			// If the field contains a "Content" field of type bytes.Buffer, write the body over there and return.
+			if field, ok := findContentsField(response, body); ok {
+				// If so, set the value
+				field.Set(reflect.ValueOf(body.ReadCloser))
 				return nil
 			}
 
@@ -84,11 +85,11 @@ func WithResponseUnmarshal(response any) DoOption {
 	}
 }
 
-func tryInjectContent(response any, body *common.ResponseWrapper) bool {
+func findContentsField(response any, body *common.ResponseWrapper) (*reflect.Value, bool) {
 	value := reflect.ValueOf(response)
 	value = reflect.Indirect(value)
 	if value.Kind() != reflect.Struct {
-		return false
+		return nil, false
 	}
 
 	// Check if there is a "Contents" of typo io.ReadCloser
@@ -96,11 +97,9 @@ func tryInjectContent(response any, body *common.ResponseWrapper) bool {
 	ioType := reflect.TypeOf((*io.ReadCloser)(nil)).Elem()
 	contentField := value.FieldByName("Contents")
 	if !contentField.IsValid() || !contentField.CanSet() || contentField.Type() != ioType {
-		return false
+		return nil, false
 	}
-	// If so, set the value
-	contentField.Set(reflect.ValueOf(body.ReadCloser))
-	return true
+	return &contentField, true
 }
 
 func injectHeaders(response any, body *common.ResponseWrapper) error {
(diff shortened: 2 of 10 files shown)


openapi/code/service.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/databricks/databricks-sdk-go/openapi"
 )
 
+// The following headers should not be added added to the generated structs
+var HIDDEN_HEADERS = map[string]bool{
```
mgyucht, line 12: nit: let's use `map[string]struct{}`, since the boolean value is meaningless.

openapi/code/service.go
```diff
@@ -202,9 +208,14 @@ func (svc *Service) newMethodEntity(op *openapi.Operation) (*Entity, openapi.Mim
 	return res, mimeType, bodyField
 }
 
-func (svc *Service) addParams(request *Entity, op *openapi.Operation, params []openapi.Parameter) {
+func (svc *Service) skipHeader(v openapi.Parameter, includeHeaders bool) bool {
+	hiddenHeader := HIDDEN_HEADERS[v.Name]
+	return v.In == "header" && (!includeHeaders || hiddenHeader)
+}
+
+func (svc *Service) addParams(request *Entity, op *openapi.Operation, params []openapi.Parameter, includeHeaders bool) {
```
mgyucht, line 216: Why do we need `includeHeaders`? Is `includeHeaders` only `false` on the request pathway? I imagine the `skipHeader` method is sufficient? What other headers are referenced in the spec?

mgyucht (commented):
A bit confused about `includeHeaders`, can you explain why that logic is needed? Otherwise LGTM.

.codegen/impl.go.tmpl
```diff
@@ -40,11 +40,7 @@ func (a *{{.Service.CamelName}}Impl) {{.PascalName}}(ctx context.Context{{if .Re
 
 {{ define "response-var" -}}
 {{ if .Response }}
-  var {{ .Response.CamelName }} {{ if .IsResponseByteStream -}}
-    io.ReadCloser
```
hectorcast-db, line 44: We are removing a special case for byte stream where we were passing just a field to the cli
```
func (a *billableUsageImpl) Download(ctx context.Context, request DownloadRequest) (*DownloadResponse, error) {
	var downloadResponse io.ReadCloser ->  var downloadResponse DownloadResponse
	path := fmt.Sprintf("/api/2.0/accounts/%v/usage/download", a.client.ConfiguredAccountID())
	headers := make(map[string]string)
	headers["Accept"] = "text/plain"
	err := a.client.Do(ctx, http.MethodGet, path, headers, request, &downloadResponse)
	return &DownloadResponse{Contents: downloadResponse}, err -> return downloadResponse
}
```

httpclient/response.go
```diff
@@ -71,3 +83,94 @@ func WithResponseUnmarshal(response any) DoOption {
...
+			value.Field(i).Set(reflect.ValueOf(headerValue))
+		case reflect.Int:
+			intValue, err := strconv.Atoi(headerValue)
+			if err != nil {
+				return fmt.Errorf("failed to parse header %s as int: %w", headerTag.name, err)
+			}
+			value.Field(i).Set(reflect.ValueOf(intValue))
+		default:
+			// Don't fail the request if we can't inject a header for backwards compatibility.
+			logger.Warnf(context.Background(), "unsupported header type %s for field %s", field.Type.Kind(), field.Name)
```
hectorcast-db, line 138: Can fail

httpclient/response.go
```diff
@@ -71,3 +84,92 @@ func WithResponseUnmarshal(response any) DoOption {
...
 }
+
+func findContentsField(response any, body *common.ResponseWrapper) (*reflect.Value, bool) {
+	value := reflect.ValueOf(response)
+	value = reflect.Indirect(value)
+	if value.Kind() != reflect.Struct {
+		return nil, false
+	}
+
+	// Check if there is a "Contents" of typo io.ReadCloser
```
tanmay-db, line 95: nit: typo -> type

httpclient/response.go
tanmay-db: What happens if there is a different type of header (other than int, string)? Should we fail earlier? If we don't fail and headers aren't injected properly (which are further used downstream) then this could lead to incorrect behaviour right?

hectorcast-db: Each option has downsides. Let's have a discussion here.

httpclient/response.go
```diff
@@ -71,3 +84,92 @@ func WithResponseUnmarshal(response any) DoOption {
...
+
+func injectHeaders(response any, body *common.ResponseWrapper) error {
+	value := reflect.ValueOf(response)
+	value = reflect.Indirect(value)
+	objectType := value.Type()
+
+	if objectType.Kind() != reflect.Struct {
+		return nil
+	}
+	headers := body.Response.Header
```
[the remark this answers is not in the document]
hectorcast-db: Header is not a pointer field, so it cannot be nil.

```
type Response struct {

	Header Header. <-- Cannot be nil

        Request *Request <-- This could be nil

```

httpclient/response.go
```diff
@@ -41,6 +46,14 @@ func WithResponseUnmarshal(response any) DoOption {
 			if response == nil {
 				return nil
 			}
+			injectHeaders(response, body)
```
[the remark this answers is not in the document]
hectorcast-db: Fixed

openapi/code/load_test.go
```diff
@@ -19,7 +19,7 @@ func TestBasic(t *testing.T) {
 
 	require.Len(t, batch.Packages(), 1)
 	require.Len(t, batch.Services(), 1)
-	require.Len(t, batch.Types(), 15)
+	require.Len(t, batch.Types(), 16)
```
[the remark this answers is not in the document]
hectorcast-db: The `/api/1.2/commands/cancel` post operation had an empty response body, so we were not generating a `CancelResponse` schema. Now we added headers to the response, so we are generating that schema.

openapi/code/service.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/databricks/databricks-sdk-go/openapi"
 )
 
+// The following headers should not be added added to the generated structs
```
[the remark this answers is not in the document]
hectorcast-db: Some headers are special. This one in particular, is for authentication. It's not a field that the user can/should add.It's something that the CLI/Auth flow will inject if necessary. Similarly, for requests (not supported yet), we won't want to expose the content-type header to the user.

I found the existance of this header when generating the spec for testing purses and @mgyucht explained me what it was and that we should not expose it.

tanmay-db (approved):
Synced offline, LGTM

PR #818, opened 2024-02-15
Add integration tests for files API

bonnetn:
## Changes
Adding integration tests for the Files API SDK methods.

## Tests
Those are the tests.

- [x] `make test` passing
- [x] `make fmt` applied


internal/files_test.go
```diff
@@ -72,9 +50,76 @@ func TestUcAccFilesAPI(t *testing.T) {
 	require.NoError(t, err)
 	contents, err := io.ReadAll(raw.Contents)
 	require.NoError(t, err)
+	defer raw.Contents.Close()
```
bonnetn, line 53: Note: It's a good practice to close the contents after reading, although not strictly necessary in this test.
For the sake of showing a good example to the users, let's have it.

hectorcast-db (approved):

PR #819, opened 2024-02-15
Add Int64 to header type injection

hectorcast-db:
## Changes
Add Int64 to header type injection

## Tests

- [X] `make test` passing
- [X] `make fmt` applied


edwardfeng-db (approved):
LGTM

PR #821, opened 2024-02-18
Differentiate between empty structures and components that can take on any value

mgyucht:
## Changes
Currently, the SDK doesn't distinguish between empty structures and components that can take any value. As a result, the SDK uses type `any` for fields that should really have a type of `struct{}` or `*struct{}` (or an equivalent type). This PR updates the OpenAPI generator code to distinguish between these two cases.

Additionally, we update the generation of SDK structures to include empty types. This includes some types that are not currently used in the SDK, mainly empty responses because the generated API calls for methods with an empty response just omits the response return value entirely. Later, we may want to include these by default; even though they are empty, it means that adding anything to an empty response is a breaking change for SDK users.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


openapi/code/package.go
```diff
@@ -299,9 +297,10 @@ func (pkg *Package) definedEntity(name string, s *openapi.Schema, processedEntit
 
 func (pkg *Package) define(entity *Entity) *Entity {
 	if entity.IsEmpty {
-		if slices.Contains(pkg.emptyTypes, &entity.Named) {
-			//panic(fmt.Sprintf("%s is already defined", entity.Name))
-			return entity
+		for _, e := range pkg.emptyTypes {
```
mgyucht, line 300: This is a bugfix, as the empty type identity is based on the type name, not the reference to the `Named` struct.

hectorcast-db (approved):
Looks good

PR #825, opened 2024-02-20
HTTP proxy example

mgyucht:
## Changes
A demonstration/example of how to use the Databricks WorkspaceClient with an HTTP and HTTPS proxy.

## Tests
Ran this on Mac and Windows.

- [ ] `make test` passing
- [ ] `make fmt` applied


examples/http-proxy/client/main.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	log.Printf("Constructing client...")
+	logger.DefaultLogger = &logger.SimpleLogger{
+		Level: logger.LevelDebug,
+	}
+	var tlsNextProto map[string]func(authority string, c *tls.Conn) http.RoundTripper
+	if os.Getenv("HTTPS_PROXY") != "" {
+		// Go's HTTP client only supports HTTP/1.1 proxies when using TLS. See
+		// https://github.com/golang/go/issues/26479 for more information.
+		tlsNextProto = make(map[string]func(authority string, c *tls.Conn) http.RoundTripper)
```
pietern, line 38: Please include "configuring this property to not be nil means disabling HTTP/2 on the HTTP transport".

examples/http-proxy/client/main.go
```diff
@@ -0,0 +1,53 @@
...
+depending on the proxy type you started.
+`)
+		os.Exit(1)
+	}
+	log.Printf("Constructing client...")
+	logger.DefaultLogger = &logger.SimpleLogger{
+		Level: logger.LevelDebug,
+	}
+	var tlsNextProto map[string]func(authority string, c *tls.Conn) http.RoundTripper
+	if os.Getenv("HTTPS_PROXY") != "" {
```
pietern, line 35: Why check if we already check that a proxy is configured earlier in the function?

examples/http-proxy/proxy/proxy.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	runCommands(cmds)
+}
+
+func removeCertFromSystemKeychain() {
+	log.Printf("Removing certificate from system keychain...")
+	allCmds := map[string][][]string{
+		"darwin":  {{"security", "remove-trusted-cert", "proxy.crt"}},
+		"linux":   {{"sudo", "rm", "/usr/local/share/ca-certificates/proxy.crt"}, {"sudo", "update-ca-certificates"}},
+		"windows": {{"certutil", "-delstore", "ROOT", "proxy.crt"}},
```
pietern, line 152: Could combine with a struct that includes both the add and remove commands.

examples/http-proxy/proxy/proxy.go
```diff
@@ -0,0 +1,235 @@
...
+	cmds, ok := allCmds[runtime.GOOS]
+	if !ok {
+		panic("unsupported OS")
+	}
+	runCommands(cmds)
+}
+
+// The following code is a simple HTTPS proxy server.
+func handleTunneling(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Handling request: %s", r.URL.String())
```
pietern, line 163: Can you include the method as well?

pietern (approved):
Cool!

examples/http-proxy/client/main.go
mgyucht: This is specifically only needed for HTTPS proxies, in which case the client is speaking HTTP/2 directly to the proxy. Go's HTTP client doesn't currently implement sending the CONNECT method correctly over HTTP/2.

The example demonstrates HTTP proxies as well. With an HTTP proxy, Go's HTTP client uses HTTP's CONNECT method over HTTP/1.1 to establish a TLS connection to a remote endpoint. That is where HTTP/2 is negotiated, and that is supported just fine.

mgyucht: Actually, I need to take this back. Stupid of me... you obviously can't proxy an HTTPS connection over HTTP, that would be a huge security problem. At least, you can't use HTTP_PROXY when the target endpoint is HTTPS...

PR #830, opened 2024-02-22
Do not transpile getWorkspaceClient

mgyucht:
## Changes
This test should not be transpiled into a Python example.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/internal/workspace_client_test.go
+++ b/internal/workspace_client_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestUcAccWorkspaceClient_CurrentWorkspaceId(t *testing.T) {
+func TestUcAccWorkspaceClient_CurrentWorkspaceId_NoTranspile(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/workspace_client_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestUcAccWorkspaceClient_CurrentWorkspaceId(t *testing.T) {
+func TestUcAccWorkspaceClient_CurrentWorkspaceId_NoTranspile(t *testing.T) {
```
tanmay-db, line 9: Curious, what's the reason for this? From the comment:
````
// Tests with NoTranspile suffix are too expensive to automatically
// translate, we just ignore them.
````
I guessing that making it transpile friendly would be more effort?

tanmay-db (approved):

internal/workspace_client_test.go
mgyucht: Yeah, basically that.

PR #837, opened 2024-02-28
Adaptive request timeouts

pietern:
## Changes

The net/http package supports only a fixed timeout.

This is problematic when streaming a large request or response body.

The approach proposed in this PR is to disable the net/http timeouts and use our own context for cancellation instead. This context is canceled only if there is no activity for the duration of 1 timeout period. As the request or response body is read, the deadline is reinitialized to equal 1 timeout period from the current timestamp. This means real timeouts will still time out in the same amount of time, but streaming requests will not time out as long as there is activity.

If a request times out, it used to return the following error:
> Get "/a?": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

Now it returns the following error:
> Get "/a?": request timed out after 10ms of inactivity

**Note:** the unit tests are time-based. Not great but I don't have a great alternative (that doesn't involve mocking time itself).

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -233,6 +233,7 @@ func (c *ApiClient) attempt(
 		// After this point, the request body has (probably) been consumed. handleError() must be called to reset it if
 		// possible.
 		if uerr, ok := err.(*url.Error); ok {
+			// If the timeout context has been canceled but the parent context hasn't, then the request has timed out.
 			if pctx.Err() == nil && uerr.Err == context.Canceled {
 				uerr.Err = fmt.Errorf("request timed out after %s of inactivity", c.config.HTTPTimeout)
 			}
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -200,10 +200,11 @@ func (c *ApiClient) attempt(
 			return c.failRequest(ctx, "failed in rate limiter", err)
 		}
 
-		// This custom context enables us to extend the request timeout
-		// while the request or response body is being read.
-		// It exists because the net/http package supports only a fixed timeout.
 		pctx := ctx
+
+		// This timeout context enables us to extend the request timeout
+		// while the request or response body is being read.
+		// It exists because the net/http package uses a fixed timeout regardless of payload size.
 		ctx, ticker := newTimeoutContext(pctx, c.config.HTTPTimeout)
 		request, err := http.NewRequestWithContext(ctx, method, requestURL, requestBody.Reader)
 		if err != nil {
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -223,7 +223,7 @@ func (c *ApiClient) attempt(
 		// If there is a request body, wrap it to extend the request timeout while it is being read.
 		// Note: we do not wrap the request body earlier, because [http.NewRequestWithContext] performs
 		// type probing on the body variable to determine the content length.
-		if request.Body != nil {
+		if request.Body != nil && request.Body != http.NoBody {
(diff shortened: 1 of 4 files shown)


pietern:
I confirmed that with a very tight timeout (5 seconds), the CLI no longer times out when working with huge files and Unity Catalog Volumes. Ad hoc debug logging confirms that the deadline keeps getting extended while the request/response body is being read. Requests succeed regardless of the network bandwidth as long as there is activity.

httpclient/timeout_context.go
```diff
@@ -0,0 +1,120 @@
...
+	"sync"
+	"time"
+)
+
+type timeoutContext struct {
+	ctx    context.Context
+	cancel context.CancelCauseFunc
+
+	// Timeout is constant.
+	// Deadline is updated when Mark function is called.
```
mgyucht, line 15:
```suggestion
	// Deadline is updated when Tick function is called.
```

httpclient/api_client.go
```diff
@@ -196,7 +199,12 @@ func (c *ApiClient) attempt(
 		if err != nil {
 			return c.failRequest(ctx, "failed in rate limiter", err)
 		}
-		request, err := http.NewRequestWithContext(ctx, method, requestURL, requestBody.Reader)
+
+		// This custom context enables us to extend the request timeout
+		// while the request or response body is being read.
+		// It exists because the net/http package supports only a fixed timeout.
+		ctx, ticker := newTimeoutContext(ctx, c.config.HTTPTimeout)
+		request, err := http.NewRequestWithContext(ctx, method, requestURL, tickingReader(ticker, requestBody.Reader))
```
mgyucht, line 207: Ok, dug through the guts of `net/http` and I believe the client uses a 32kb buffer when reading from the request body. So this would mean that if less than 32kb is written out over `HTTPTimeout` duration, the request could be broken. Given the default 30s timeout, the user needs to sustain more than 1066b/s. Sounds good to me.

httpclient/timeout_context.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"context"
+	"io"
+	"sync"
+	"time"
+)
+
+type timeoutContext struct {
+	ctx    context.Context
+	cancel context.CancelCauseFunc
```
mgyucht, line 12: We should aim to deprecate go 1.19 at some point. It is currently EOL for 5 months. Will do a bit of datascience to see how many users would be affected by this deprecation. In the meantime, can we use `WithCancel()`?

pietern:
Kicking off an integration test run now.

httpclient/api_client.go
```diff
@@ -211,16 +220,35 @@ func (c *ApiClient) attempt(
 		if request.Header.Get("Content-Type") == "" && requestBody.ContentType != "" {
 			request.Header.Set("Content-Type", requestBody.ContentType)
 		}
+		// If there is a request body, wrap it to extend the request timeout while it is being read.
+		// Note: we do not wrap the request body earlier, because [http.NewRequestWithContext] performs
+		// type probing on the body variable to determine the content length.
+		if request.Body != nil && request.Body != http.NoBody {
```
pietern, line 226: @mgyucht I needed to check for `http.NoBody` as well. The guts of `net/http` use this to signal an empty body if it is not nil... We clearly don't care about extending timeouts if there is no body, so safe to filter here.

mgyucht (approved):
LGTM. Thanks for the clear comments!

pietern:
Integration tests pass.

PR #838, opened 2024-02-29
Added Config.GetAuthDetails

andrewnester:
## Changes
Added Config.GetAuthDetails and use it in config debug string

Needed for https://github.com/databricks/cli/pull/1244
## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -63,16 +63,16 @@ type Config struct {
 	// in ~/.databrickscfg.
 	ConfigFile string `name:"config_file" env:"DATABRICKS_CONFIG_FILE"`
 
-	GoogleServiceAccount string `name:"google_service_account" env:"DATABRICKS_GOOGLE_SERVICE_ACCOUNT" auth:"google"`
-	GoogleCredentials    string `name:"google_credentials" env:"GOOGLE_CREDENTIALS" auth:"google,sensitive"`
+	GoogleServiceAccount string `name:"google_service_account" env:"DATABRICKS_GOOGLE_SERVICE_ACCOUNT" auth:"google" auth_types:"google-id"`
+	GoogleCredentials    string `name:"google_credentials" env:"GOOGLE_CREDENTIALS" auth:"google,sensitive" auth_types:"google-credentials"`
 
 	// Azure Resource Manager ID for Azure Databricks workspace, which is exhanged for a Host
-	AzureResourceID string `name:"azure_workspace_resource_id" env:"DATABRICKS_AZURE_RESOURCE_ID" auth:"azure-cli,azure-msi" auth_group:"azure"`
+	AzureResourceID string `name:"azure_workspace_resource_id" env:"DATABRICKS_AZURE_RESOURCE_ID" auth:"azure" auth_types:"azure-cli,azure-msi"`
 
-	AzureUseMSI       bool   `name:"azure_use_msi" env:"ARM_USE_MSI" auth:"azure-msi" auth_group:"azure"`
-	AzureClientSecret string `name:"azure_client_secret" env:"ARM_CLIENT_SECRET" auth:"azure-client-secret,sensitive" auth_group:"azure"`
-	AzureClientID     string `name:"azure_client_id" env:"ARM_CLIENT_ID" auth:"azure-client-secret,azure-msi" auth_group:"azure"`
-	AzureTenantID     string `name:"azure_tenant_id" env:"ARM_TENANT_ID" auth:"azure-cli,azure-client-secret" auth_group:"azure"`
+	AzureUseMSI       bool   `name:"azure_use_msi" env:"ARM_USE_MSI" auth:"azure" auth_types:"azure-msi"`
+	AzureClientSecret string `name:"azure_client_secret" env:"ARM_CLIENT_SECRET" auth:"azure,sensitive" auth_types:"azure-client-secret"`
+	AzureClientID     string `name:"azure_client_id" env:"ARM_CLIENT_ID" auth:"azure" auth_types:"azure-client-secret,azure-msi"`
+	AzureTenantID     string `name:"azure_tenant_id" env:"ARM_TENANT_ID" auth:"azure" auth_types:"azure-cli,azure-client-secret"`
 
 	// AzureEnvironment (PUBLIC, USGOVERNMENT, CHINA) has specific set of API endpoints. Starting from v0.26.0,
 	// the environment is determined based on the workspace hostname, if it's specified.
@@ -147,6 +147,9 @@ type Config struct {
 
 	// HTTP request interceptor, that assigns Authorization header
 	auth func(r *http.Request) error
+
+	// Keep track of the source of each attribute
+	attrSource map[string]*Source
 }
 
 // NewWithWorkspaceHost returns a new instance of the Config with the host set to
--- a/config/config.go
+++ b/config/config.go
@@ -67,12 +67,12 @@ type Config struct {
 	GoogleCredentials    string `name:"google_credentials" env:"GOOGLE_CREDENTIALS" auth:"google,sensitive"`
 
 	// Azure Resource Manager ID for Azure Databricks workspace, which is exhanged for a Host
-	AzureResourceID string `name:"azure_workspace_resource_id" env:"DATABRICKS_AZURE_RESOURCE_ID" auth:"azure"`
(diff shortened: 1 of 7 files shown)


pietern (commented):
I don't understand why the additional `auth_group` tag is needed. When I read the comment in the struct I kind of get it, but then I don't see where it has an effect in the rest of the PR. For example, in the `Validate` function the code falls back to using the `AuthGroup` field and the result is the same as before this PR. Am I missing anything?

config/config.go
```diff
@@ -67,12 +67,12 @@ type Config struct {
...
+	AzureResourceID string `name:"azure_workspace_resource_id" env:"DATABRICKS_AZURE_RESOURCE_ID" auth:"azure-cli,azure-msi" auth_group:"azure"`
 
-	AzureUseMSI       bool   `name:"azure_use_msi" env:"ARM_USE_MSI" auth:"azure"`
-	AzureClientSecret string `name:"azure_client_secret" env:"ARM_CLIENT_SECRET" auth:"azure,sensitive"`
-	AzureClientID     string `name:"azure_client_id" env:"ARM_CLIENT_ID" auth:"azure"`
-	AzureTenantID     string `name:"azure_tenant_id" env:"ARM_TENANT_ID" auth:"azure"`
+	AzureUseMSI       bool   `name:"azure_use_msi" env:"ARM_USE_MSI" auth:"azure-msi" auth_group:"azure"`
+	AzureClientSecret string `name:"azure_client_secret" env:"ARM_CLIENT_SECRET" auth:"azure-client-secret,sensitive" auth_group:"azure"`
+	AzureClientID     string `name:"azure_client_id" env:"ARM_CLIENT_ID" auth:"azure-client-secret,azure-msi" auth_group:"azure"`
+	AzureTenantID     string `name:"azure_tenant_id" env:"ARM_TENANT_ID" auth:"azure-cli,azure-client-secret" auth_group:"azure"`
```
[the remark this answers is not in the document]
andrewnester: I change `auth` tag to be a specific type of authentication in order to be able to describe which properties are used / not used for specific auth method. But since some of attributes are can be in more than 1 auth `validate` call will fail if they are used. So for this I introduce `auth_group` attribute to make sure that `validate` behaviour stay the same.

config/config_attribute.go
```diff
@@ -7,17 +7,45 @@ import (
...
+	Name    string
+	Kind    reflect.Kind
+	EnvVars []string
+	Auth    []string
+
+	// AuthGroup is used to group auth attributes.
+	// It is used later in validate step to check if there are more than 1 auth defined.
+	// This field is needed because some auth attributes can be used in multiple auth methods of the same group.
+	// But when we are validating we need to check if there are auth attribute from different groups.
+	// If AuthGroup is empty, we use Auth as a grouping field.
```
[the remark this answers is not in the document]
andrewnester: Good point, a good idea to capture this

config/config.go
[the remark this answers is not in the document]
pietern: Ah, I see. Is the intent to make it match the resolved `auth_type`?

If so, then the Google ones also need to be updated. It may be easier to keep `auth` as is and use a new tag to annotate the auth types (e.g. `auth_types`).

config/config_auth_details.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (a *AttrConfig) String() string {
+	s := fmt.Sprintf(`%s (from %s)`, a.Value, a.Source.String())
+	if a.AuthTypeMismatch {
+		s += " (not used by the current auth type)"
+	}
+	return s
+}
+
+func (c *Config) GetAuthDetails(showSensitive bool) AuthDetails {
```
pietern, line 79: Could we default this to never include sensitive values?

A bool argument can be deceptive. If we need a way to include the sensitive fields, I would rather have a separate function call that makes it clear it is including sensitive values verbatim.

config/config_attributes.go
```diff
@@ -81,12 +52,21 @@ func (a attributes) Name() string {
...
-		v := attr.ReadEnv()
+		var v string
+		for _, envName := range attr.EnvVars {
+			v = os.Getenv(envName)
+			if v == "" {
+				continue
+			}
+			cfg.SetAttrSource(attr, &Source{Type: SourceEnv, Name: envName})
+			break
+		}
```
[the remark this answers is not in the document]
andrewnester: Indeed possibly, the reason I didn't do it is not to make backward incompatible change

config/config_attributes.go
```diff
@@ -81,12 +52,21 @@ func (a attributes) Name() string {
 
 // Configure implements Loader interface for environment variables
 func (a attributes) Configure(cfg *Config) error {
-	for _, attr := range a {
+	for k := range a {
+		attr := &a[k]
```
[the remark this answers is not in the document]
andrewnester: It's used in CLI as well and overall better abstraction (I believe) than just checking if .Auth is empty

config/config_attributes.go
```diff
@@ -112,12 +118,18 @@ func (a attributes) ResolveFromStringMap(cfg *Config, m map[string]string) error
...
 }
 
 func (a attributes) ResolveFromAnyMap(cfg *Config, m map[string]interface{}) error {
-	for _, attr := range a {
+	return a.ResolveFromAnyMapWithSource(cfg, m, &Source{Type: SourceDynamicConfig})
+}
+
+func (a attributes) ResolveFromAnyMapWithSource(cfg *Config, m map[string]interface{}, source *Source) error {
+	for k := range a {
+		attr := &a[k]
```
pietern, line 132: Pointer references no longer needed.

config/config_attributes.go
```diff
@@ -99,7 +100,12 @@ func (a attributes) Configure(cfg *Config) error {
 }
 
 func (a attributes) ResolveFromStringMap(cfg *Config, m map[string]string) error {
-	for _, attr := range a {
+	return a.ResolveFromStringMapWithSource(cfg, m, &Source{Type: SourceDynamicConfig})
+}
+
+func (a attributes) ResolveFromStringMapWithSource(cfg *Config, m map[string]string, source *Source) error {
+	for k := range a {
+		attr := &a[k]
```
pietern, line 108: Pointer references no longer needed.

config/config.go
```diff
@@ -147,6 +147,9 @@ type Config struct {
 
 	// HTTP request interceptor, that assigns Authorization header
 	auth func(r *http.Request) error
+
+	// Keep track of the source of each attribute
+	attrSource map[string]*Source
```
pietern, line 152: The `*Source` elements of this map are never mutated, so they can (should?) be a regular value, not a ptr.

config/config_auth_details.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (c AuthConfiguration) String() string {
+	keys := make([]string, 0, len(c))
+	for k := range c {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+
+	var conf []string
+	for _, k := range keys {
```
pietern, line 48: Can we store this as a list in the natural order (same as they appear in `ConfigAttributes`)?

pietern:
@mgyucht Could you take a look at this as well?

config/config_attributes.go
```diff
@@ -86,11 +86,12 @@ func (a attributes) Configure(cfg *Config) error {
 			// don't overwtite a value previously set
 			continue
 		}
-		v := attr.ReadEnv()
+		v, envName := attr.ReadEnv()
 		if v == "" {
 			continue
 		}
 		err := attr.SetS(cfg, v)
+		cfg.SetAttrSource(&attr, Source{Type: SourceEnv, Name: envName})
```
pietern, line 94: Strictly speaking, this should only be set after the error check.

pietern (commented):
LGTM, thanks! Will defer to @mgyucht for the stamp.

config/config_auth_details.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return AuthDetails{
+		AuthType:      c.AuthType,
+		Host:          c.Host,
+		Configuration: attrSet,
+	}
+}
+
+func (c *Config) SetAttrSource(attr *ConfigAttribute, source Source) {
```
mgyucht, line 101: Does this need to be public?

mgyucht (approved):
Small comment, but otherwise LGTM. Thanks for doing this! We'll need to update the error logging to take advantage of this change.

config/config_auth_details.go
andrewnester: @mgyucht yes, it's being used in CLI to set a different attribute source which does not exist in SDK (`bundle` to be specifc)

PR #841, opened 2024-03-04
Include Go 1.22 in test matrix

pietern:


pietern:
FYI:
* 1.19 is EOL since Sep 6, 2023 (release of 1.21)
* 1.20 is EOL since Feb 6, 2024 (release of 1.22)

We should figure out if we follow this schedule or trail it by some fixed interval.

tanmay-db (approved):

PR #851, opened 2024-03-11
Add telemetry for SDK usage from DBR

shreyas-goenka:
## Changes
This PR adds telemetry through the user agent for DBR usage from the SDK. We use the same mechanism and key as the Python SDK to add this to the user agent (PR: https://github.com/databricks/databricks-sdk-py/pull/287).

## Tests
Using unit tests


useragent/user_agent.go
```diff
@@ -85,13 +85,27 @@ func (u info) String() string {
...
+	}
+	if !isAlphanum(value) && !isSemVer(value) {
+		return fmt.Errorf("expected user agent value for key %q to be alphanumeric or semver: %s", key, value)
+	}
+	return nil
+}
+
 // With always uses the latest value for a given alphanumeric key.
 // Panics if key or value don't satisfy alphanumeric or semver format.
 func (d data) With(key, value string) data {
```
nfx, line 107: why not a simple `if` statement in the `With` method?

nfx (changes requested):

useragent/user_agent.go
shreyas-goenka: It's cleaner this way IMO since "runtime" is the exception case.  We still want to add the key when it's "runtime", meaning the check ends up being something like:
```
 if key != "runtime" && !isAlphanum(value) && !isSemVer(value) {}
```

shreyas-goenka: Also better error messages and can unit test this way. With the panic stack trace it's difficult to tell where exactly the validation failed.

useragent/patterns.go
```diff
@@ -36,3 +36,11 @@ func matchAlphanumOrSemVer(s string) error {
 	}
 	return fmt.Errorf("invalid alphanumeric or semver string: %s", s)
 }
+
+func isAlphanum(s string) bool {
+	return regexpAlphanum.MatchString(s)
```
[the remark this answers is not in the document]
shreyas-goenka: done

config/api_client.go
```diff
@@ -64,6 +66,20 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
...
+				// Detect if the SDK is being run in a Databricks Runtime.
+				v, ok := os.LookupEnv("DATABRICKS_RUNTIME_VERSION")
+				if !ok {
+					return nil
+				}
+				// Replace any slashes and spaces with dashes, just in case.
+				v = strings.ReplaceAll(v, "/", "-")
+				v = strings.ReplaceAll(v, " ", "-")
+				// Add the detected Databricks Runtime version to the user agent
+				ctx := useragent.InContext(r.Context(), "runtime", v)
```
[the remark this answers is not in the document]
shreyas-goenka: We have to call it "runtime" in this case because the Python SDK already calls it runtime, and we don't want to diverge from that convention.
see: https://github.com/databricks/databricks-sdk-py/pull/287

The string constant part is done!

useragent/user_agent.go
```diff
@@ -79,19 +86,48 @@ type info struct {
 	Value string
 }
 
+// Sanitize the user agent value. This is useful when the value is not ensured to be
+// to be valid at compile time. Having this sanitization then ensures downstream
+// applications can correctly parse the full user agent header, by making sure
+// characters like '/' and ' ' are not present in the value.
+func Sanitize(s string) string {
+	allowList := func(r rune) rune {
+		if r == '.' || unicode.IsLetter(r) || unicode.IsDigit(r) {
```
pietern, line 95: This includes everything in the "letter" class, which includes modifiers, accents, and more that doesn't match the regexp for alphanumerical values.

To guarantee that sanitize produces a valid value it would need to use the same set of characters as the characters in the alphanumeric regexp and map all non-matching characters to the placeholder.

useragent/user_agent.go
```diff
@@ -79,19 +86,48 @@ type info struct {
...
 	return fmt.Sprintf("%s/%s", u.Key, u.Value)
 }
 
 type data []info
 
+// Validate the key value pair being set in the user agent. Error if invalid.
+func validate(key, value string) error {
+	// DBR versions as set in the `DATABRICKS_RUNTIME_VERSION` environment variable
+	// are not valid semver strings. Eg: 15.1 or client.0. Thus we allow arbitrary
+	// values for the runtime key.
```
pietern, line 114: We could broaden `regexpAlphanum` class to include `.`. That seems less brittle than the current approach.

Are there other examples of values that we need to allow?

useragent/user_agent.go
pietern: A refresh of the error messages sounds good, but special casing `runtime` does not.

useragent/user_agent_test.go
```diff
@@ -60,3 +60,22 @@ func TestFromContext_Custom(t *testing.T) {
...
+	assert.NoError(t, validate("runtime", "whatever#!@"))
+	assert.NoError(t, validate("abc", "123"))
+	assert.NoError(t, validate("abc", "1.1.1"))
+}
+
+func TestUserAgentNormalizeString(t *testing.T) {
+	assert.Equal(t, "abc123", Sanitize("abc123"))
+	assert.Equal(t, "1-2-3-4-5-6-7-8-", Sanitize("1@2#3?4,5/6!7 8 "))
+	assert.Equal(t, "1.2.3", Sanitize("1.2.3"))
+	assert.Equal(t, "client.0", Sanitize("client.0"))
```
pietern, line 80: Please test that sanitized values match `isAlphanum`, and include Unicode characters, emoji's, etc.

useragent/patterns_test.go
```diff
@@ -8,25 +8,54 @@ import (
...
+	assert.True(t, isAlphanum("foo-bar"))
+
 	assert.Error(t, matchAlphanum("foo bar"))
+	assert.False(t, isAlphanum("foo bar"))
+
 	assert.Error(t, matchAlphanum("foo/bar"))
+	assert.False(t, isAlphanum("foo/bar"))
 }
 
 func TestMatchAlphanumOrSemVer(t *testing.T) {
```
pietern, line 46: I don't think this function is used.

useragent/user_agent.go
```diff
@@ -79,19 +86,48 @@ type info struct {
...
+	if key == RuntimeKey {
+		return nil
+	}
+	if !isAlphanum(key) {
+		return fmt.Errorf("expected user agent key to be alphanumeric: %s", key)
```
pietern, line 119: Use `%q` for key so that it is quoted as well (it's not alphanumeric so it could be anything).

useragent/user_agent.go
```diff
@@ -79,19 +86,48 @@ type info struct {
...
+	}
+	if !isAlphanum(value) && !isSemVer(value) {
```
pietern, line 121: Did you mean to use `matchAlphanumOrSemVer`?

client/client_test.go
```diff
@@ -320,6 +320,64 @@ GET /a
...
 }
 
+func TestUserAgentForDBR(t *testing.T) {
+	goVersion := strings.TrimPrefix(runtime.Version(), "go")
+	cicdHeader := ""
+	if useragent.CiCdProvider() != "" {
+		cicdHeader = fmt.Sprintf(" cicd/%s", useragent.CiCdProvider())
+	}
+	t.Setenv("DATABRICKS_RUNTIME_VERSION", "15.1")
+	expectedUserAgent := `unknown/0.0.0 databricks-sdk-go/` + version.Version + ` go/` + goVersion + ` os/` + runtime.GOOS + ` auth/pat` + cicdHeader + ` runtime/15.1`
```
pietern, line 330: Can you check `assert.Contains(t, " runtime/15.1", userAgent)` instead?

The other details seem irrelevant to the test.

client/client_test.go
```diff
@@ -320,6 +320,64 @@ GET /a
...
+				Body:       io.NopCloser(strings.NewReader(`{}`)),
+				Request:    r,
+			}, nil
+		}),
+	})
+	require.NoError(t, err)
+	err = c.Do(context.Background(), "GET", "/a", nil, nil, nil)
+	assert.Equal(t, userAgent, expectedUserAgent)
+	require.NoError(t, err)
+}
```
pietern, line 379: This test is equivalent to the one above, except for the value of the env var.

useragent/user_agent.go
shreyas-goenka: Removed the exception for `runtime`

useragent/user_agent.go
shreyas-goenka: Good catch. Did some research and it seems UTF8 characters are not supported in general in HTTP headers. The recommended approach seems to be to stick to ASCII characters in the HTTP headers. 

I have changed the validation to allow alphanumeric, `-`, `_` AND `.` as valid characters.

useragent/user_agent.go
shreyas-goenka: Done. We don't need any other characters at the moment, but we can expand this set as and when needed.

useragent/user_agent_test.go
shreyas-goenka: Done

useragent/user_agent.go
shreyas-goenka: Done

useragent/user_agent.go
shreyas-goenka: Now we use a single function, `isValid`

client/client_test.go
shreyas-goenka: Done

client/client_test.go
shreyas-goenka: Good point. Added test casing.

useragent/patterns.go
```diff
@@ -14,25 +14,28 @@ const (
...
-		return nil
-	}
-	return fmt.Errorf("invalid alphanumeric string: %s", s)
+// Alphanumeric characters, hyphen, underscore, and period. This is the subset of
+// characters that we allow in user agent keys and values. This is to ensure that
+// downstream applications can correctly parse the full user agent header.
+//
+// NOTE: HTTP headers in general only work well with ASCII characters. see:
+// https://stackoverflow.com/questions/4400678/what-character-encoding-should-i-use-for-a-http-header
+var validChars = `0-9A-Za-z_\-\.`
```
pietern, line 30: Please initialize the regexps you wish to use as globals as well. This string (which can be a const) is not useful outside of regexps. It can also include the brackets because it is always a character class.

The regexp compile adds overhead for every call (multiple calls per request) that we can amortize by initializing it only once.

useragent/user_agent.go
```diff
@@ -99,13 +106,29 @@ func (u info) String() string {
 
 type data []info
 
+// Validate the key value pair being set in the user agent. Error if invalid.
+func validate(key, value string) error {
+	if !isValid(key) {
+		return fmt.Errorf("expected user agent key to be alphanumeric: %q", key)
+	}
+	if !isValid(value) {
```
pietern, line 114: This no longer checks for semver strings, which can already be in use.

I believe the semver string is a superset of the current alphanum class, so that can be an issue.

Also, I don't think you need the `isValid` function and can avoid the additional function definition.

useragent/user_agent_test.go
```diff
@@ -74,3 +74,56 @@ func TestFromContext_Custom(t *testing.T) {
...
+	assert.Equal(t, "foo-bar", Sanitize("foo bar"))
+
+	assert.False(t, isValid("foo/bar"))
+	assert.Equal(t, "foo-bar", Sanitize("foo/bar"))
+
+	assert.False(t, isValid("foo:)bar"))
+	assert.Equal(t, "foo--bar", Sanitize("foo:)bar"))
+
+	assert.False(t, isValid("foo😊bar"))
+	assert.Equal(t, "foo-bar", Sanitize("foo😊bar"))
```
pietern, line 128: This is hard to read; both the valid cases and the invalid but sanitizes cases can be a slice.

config/api_client.go
```diff
@@ -60,7 +61,19 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
...
+				*r = *r.WithContext(ctx) // replace request
+				return nil
+			},
+			func(r *http.Request) error {
+				// Detect if the SDK is being run in a Databricks Runtime.
+				v, ok := os.LookupEnv("DATABRICKS_RUNTIME_VERSION")
+				if !ok {
+					return nil
+				}
+				v = useragent.Sanitize(v)
```
pietern, line 74: Please also move this to `useragent/runtime.go` so we can cache the output just like we do for CI/CD above.

shreyas-goenka: done

useragent/user_agent_test.go
shreyas-goenka: done

useragent/user_agent.go
shreyas-goenka: done

useragent/patterns.go
shreyas-goenka: removed

useragent/user_agent.go
```diff
@@ -99,6 +106,22 @@ func (u info) String() string {
...
+//
+// Note: Semver strings are comprised of alphanumeric characters, hyphens, periods
+// and plus signs. This function will not remove these characters.
+// see:
+// 1. https://semver.org/#spec-item-9
+// 2. https://semver.org/#spec-item-10
+func Sanitize(s string) string {
+	return regexp.MustCompile(`[^0-9A-Za-z_\.\+-]`).ReplaceAllString(s, "-")
+}
+
```
pietern, line 124: Please move to `patterns.go` and keep the regexp definition close to the one defining the regexp for alphanum.

More likely they will drift otherwise.

The regexp can also be statically initialized instead of recompiled on every call (per my earlier comment).

useragent/runtime.go
```diff
@@ -0,0 +1,29 @@
...
+func Runtime() string {
+	if DisableRuntimeCaching {
+		return getRuntimeVersion()
+	}
+
+	runtimeOnce.Do(func() {
+		runtimeVersion = getRuntimeVersion()
+	})
+	return runtimeVersion
+}
```
pietern, line 29: Like the CI/CD test, this one can have a companion test where you verify that a few select runtime version strings are correctly returned/sanitized and are valid. Then the e2e test can just verify that it is set and contains something and not iterate over all options.

Also removes the need for exported symbol for `DisableRuntimeCaching`.

pietern (commented):
Please use newlines in all table tests. It is hard to read the cases that are covered otherwise.

useragent/user_agent.go
shreyas-goenka: done

useragent/runtime.go
shreyas-goenka: done

useragent/user_agent.go
shreyas-goenka: done

useragent/runtime.go
shreyas-goenka: The end to end test is failing because of the caching. Does it make sense to keep the exported flag?

useragent/user_agent_test.go
```diff
@@ -74,3 +74,30 @@ func TestFromContext_Custom(t *testing.T) {
 func TestDefaultsAreValid(t *testing.T) {
 	WithProduct(productName, productVersion)
 }
+
+func TestSanitize(t *testing.T) {
```
pietern, line 78: Nit: this belongs in `patterns_test.go`.

useragent/runtime.go
pietern: What do we do for the CI/CD one?

shreyas-goenka: We originally did had the e2e test for the CI/CD one. Also the tests are run from GitHub actions so another one of our tests as a side effect asserted on the CICD useragent. (`TestNonJSONResponseIncludedInError`)

shreyas-goenka: Done. Added a `ClearCache` method and also added an e2e test for the cicd user agent.

useragent/user_agent_test.go
shreyas-goenka: done

client/client_test.go
```diff
@@ -320,6 +321,99 @@ GET /a
...
+		},
+		"tf-cloud": {
+			"TFC_RUN_ID": "",
+		},
+	}
+
+	for ci, envVars := range ciToEnv {
+		t.Run(ci, func(t *testing.T) {
+			env.CleanupEnvironment(t)
+			useragent.ClearCache()
```
pietern, line 402: What's the rationale for a `t.Run` here but not in the other test?

client/client_test.go
```diff
@@ -320,6 +321,99 @@ GET /a
...
+	require.NoError(t, err)
+
+	err = c.Do(context.Background(), "GET", "/a", nil, nil, nil)
+	require.NoError(t, err)
+
+	return userAgent
+}
+
+func TestUserAgentForDBR(t *testing.T) {
+	for _, dbrVersion := range []string{"client.0", "client.1", "15.5", "15.5.0", "13.3"} {
```
pietern, line 350: Please include newline separators here as well for readability.

Btw, doesn't this duplicate the tests done in the `useragent` package?

useragent/runtime_test.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func TestUserAgentRuntime(t *testing.T) {
+	for _, tc := range []string{
+		"1.2.3",
+		"0.0.0-dev+2e014739024a",
+		"client.0",
+		"foo",
+		"15.0",
+		"13.3",
```
pietern, line 16: Can you include an emoji/invalid character here as well?

I think this test passes if you remove the `Sanitize` call from the function.

pietern (approved):

client/client_test.go
shreyas-goenka: This is more robust since it does not depend upon the implementation details of the validation logic and tests it end to end. In the other test we call the `match...` function manually to verify validation does not fail.

client/client_test.go
shreyas-goenka: We should have `t.Run` at both places. That makes for better error logs. Updated the PR.

client/client_test.go
pietern: That's fine, but could you then also include a garbage value to test that it doesn't panic/fail.

shreyas-goenka: Added a couple of cases that need sanitization.

useragent/runtime_test.go
shreyas-goenka: done

PR #864, opened 2024-03-22
Override INVALID_PARAMETER_VALUE on fetching non-existent job/cluster

mgyucht:
## Changes
Most services use `RESOURCE_DOES_NOT_EXIST` error code with 404 status code to indicate that a resource doesn't exist. However, for legacy reasons, Jobs and Clusters services use `INVALID_PARAMETER_VALUE` error code with 400 status code instead. This makes tools like Terraform and UCX difficult to maintain, as these services need different error handling logic. However, we can't change these behaviors as customers already depend on the raw HTTP response status & contents.

This PR corrects these errors in the SDK itself. SDK users can then do
```go
_, err := w.Jobs.Get(ctx, jobs.GetJobRequest{JobId: 12345})
if errors.Is(err, apierr.ErrResourceDoesNotExist) { ... }
```
just as you would expect from other resources.

Updated the README with more information about this as well.

## Tests
Added unit tests for error overrides.
Added 2 integration tests for Clusters and Jobs.

- [x] `make test` passing
- [x] `make fmt` applied


apierr/error_override.go
```diff
@@ -0,0 +1,83 @@
...
+	if e.errorCodeMatcher != nil && !e.errorCodeMatcher.MatchString(err.ErrorCode) {
+		return false
+	}
+	if e.messageMatcher != nil && !e.messageMatcher.MatchString(err.Message) {
+		return false
+	}
+	return true
+}
+
+var invalidParameterValue = "INVALID_PARAMETER_VALUE"
```
tanmay-db, line 54: If we aren't changing this, then it would be good to have this as a constant

apierr/error_override.go
```diff
@@ -0,0 +1,83 @@
...
+
+var allOverrides = []errorOverride{
```
tanmay-db, line 56: nit: Might be good to encapsulate this in a method so it's not accidentally modified.

apierr/error_override.go
```diff
@@ -0,0 +1,83 @@
...
+	{
+		debugName:        "Clusters InvalidParameterValue => ResourceDoesNotExist",
+		pathRegex:        regexp.MustCompile(`^/api/2\.0/clusters/get`),
+		verb:             "GET",
+		messageMatcher:   regexp.MustCompile("Cluster .* does not exist"),
+		errorCodeMatcher: regexp.MustCompile(invalidParameterValue),
```
tanmay-db, line 62: Should we have a `statusCodeMatcher`?

tanmay-db (approved):
LGTM

apierr/error_override.go
mgyucht: Good idea, will add.

tanmay-db (approved):

PR #869, opened 2024-03-26
Properly escape multi-segment path parameters

mgyucht:
## Changes
Currently, path parameters are directly interpolated into the request URL without escaping. This means that characters like `/`, `?` and `#` will not be percent-encoded and will affect the semantics of the URL, starting a new path segment, query parameters, or fragment, respectively. This means that it is impossible for users of the Files API to upload/download objects that contain `?` or `#` in their name. `/` is allowed in the path of the Files API, so it does not need to be escaped.

The Files API is currently marked with `x-databricks-multi-segment`, indicating that it should be permitted to have `/` characters but other characters need to be percent-encoded. This PR implements this.

This PR builds on #868, correcting the test names for UC files.

## Tests
- [x] Unit test for multi-segment path escaping behavior.
- [x] Updated integration test to use `#` and `?` symbols in the file name. This failed on `main` but succeeded on this PR.

### DIFF
--- a/httpclient/request.go
+++ b/httpclient/request.go
@@ -112,6 +112,7 @@ func makeQueryString(data interface{}) (string, error) {
 func EncodeMultiSegmentPathParameter(p string) string {
 	segments := strings.Split(p, "/")
 	b := strings.Builder{}
+	b.Grow(len(p))
 	for i, s := range segments {
 		if i > 0 {
 			b.WriteString("/")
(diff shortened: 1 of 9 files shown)


httpclient/request.go
```diff
@@ -109,6 +109,18 @@ func makeQueryString(data interface{}) (string, error) {
 	return "", fmt.Errorf("unsupported query string data: %#v", data)
 }
 
+func EncodeMultiSegmentPathParameter(p string) string {
+	segments := strings.Split(p, "/")
+	b := strings.Builder{}
```
pietern, line 114: This can be preallocated to at least `len(p)`.

hectorcast-db (approved):

PR #874, opened 2024-04-02
Support custom AuthVisitors

hectorcast-db:
## Changes
Support custom AuthVisitors

## Tests
- [X] `make test` passing
- [X] `make fmt` applied

### DIFF
--- a/config/api_client.go
+++ b/config/api_client.go
@@ -49,7 +49,11 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
 				return nil
 			},
 			func(r *http.Request) error {
-				ctx := useragent.InContext(r.Context(), useragent.AuthKey, c.AuthType)
+				authType := c.AuthType
+				if t := r.Context().Value(useragent.AuthKey); t != nil {
+					authType = t.(string)
+				}
+				ctx := useragent.InContext(r.Context(), useragent.AuthKey, authType)
 				*r = *r.WithContext(ctx) // replace request
 				return nil
 			},
--- a/config/api_client.go
+++ b/config/api_client.go
@@ -34,8 +34,8 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
 		DebugTruncateBytes: c.DebugTruncateBytes,
 		InsecureSkipVerify: c.InsecureSkipVerify,
 		Transport:          c.HTTPTransport,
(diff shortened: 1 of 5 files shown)


config/api_client.go
```diff
@@ -34,8 +34,8 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
 		DebugTruncateBytes: c.DebugTruncateBytes,
 		InsecureSkipVerify: c.InsecureSkipVerify,
 		Transport:          c.HTTPTransport,
+		AuthVisitor:        c.Authenticate,
 		Visitors: []httpclient.RequestVisitor{
-			c.Authenticate,
```
nfx, line 38: why duplication? AuthVisitor is still a visitor.

nfx (changes requested):

config/api_client.go
mgyucht: We will be overriding auth per method for direct-to-dataplane operations. These will not use the cached header factory, they'll use a new one that is injected on a per-method basis depending on the endpoint in question.

mgyucht (approved):

config/api_client.go
```diff
@@ -49,7 +49,11 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
 				return nil
 			},
 			func(r *http.Request) error {
-				ctx := useragent.InContext(r.Context(), useragent.AuthKey, c.AuthType)
+				authType := c.AuthType
+				if t := r.Context().Value(useragent.AuthKey); t != nil {
```
mgyucht, line 53: Does `go vet` show yellow squigglies for this? I think we need to define these context keys as separate types to prevent users from accidentally overriding them.

mgyucht (approved):
One comment, otherwise LGTM.

httpclient/api_client_test.go
```diff
@@ -595,3 +595,89 @@ func TestOAuth2Integration(t *testing.T) {
...
+				Request:    r,
+			}, nil
+		}),
+	})
+	err := c.Do(context.Background(), "GET", "/a/b", WithRequestData(map[string]any{}), authOption)
+	require.NoError(t, err)
+}
+
+func TestDefaultAuthVisitor(t *testing.T) {
+	defaultAuthVisitor := func(r *http.Request) error {
```
mgyucht, line 658: Can you test that the user agent is set correctly as well?

PR #882, opened 2024-04-05
Do not leak secondary authorization tokens in debug logs

mgyucht:
## Changes
Besides the `Authorization` token, there are other tokens included by the SDK in requests in Azure & GCP with headers `X-Databricks-Azure-SP-Management-Token` and `X-Databricks-GCP-SA-Access-Token`. These need to be redacted as well unless DebugAuthorizationHeaders is set.

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


nfx (approved):
lgtm

PR #886, opened 2024-04-09
Create a method to generate OAuth tokens

hectorcast-db:
## Changes
Create a method to generate OAuth tokens. This is required to later generate tokens to use in the DataPlane APIs. This PR introduces a breaking change by renaming `CredentialsProvider` -> `CredentialsStrategy` and changing the signature of the `Configure` method for such interface.

## Tests

- [X] Manual test. No API returns the `authorization_details` for an object. 
- [X] `make test` passing
- [X] `make fmt` applied


httpclient/oauth_token.go
```diff
@@ -0,0 +1,50 @@
...
+		"Content-Type": "application/x-www-form-urlencoded",
+	}
+	data := GetOAuthTokenRequest{
+		GrantType:            "urn:ietf:params:oauth:grant-type:jwt-bearer",
+		AuthorizationDetails: credentials.AuthorizationDetails(authorizationDetails),
+		Assertion:            token.AccessToken,
+		// By default, the tokens API does not refresh the token, so you can get a token that is about to expire.
+		// We keep a cache of the Tokens, and only request when a new one is needed
+		ExpiresIn:    3600,
+		RefreshToken: true,
```
hectorcast-db, line 36: It does not seem to work. Tokens are only refreshed when the old one is expired, which is an issue. Pending discussion with auth team.

httpclient/oauth_token.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/databricks/databricks-sdk-go/credentials"
+	"golang.org/x/oauth2"
+)
+
+type GetOAuthTokenRequest struct {
+	GrantType            string `url:"grant_type"`
+	AuthorizationDetails string `url:"authorization_details"`
+	Assertion            string `url:"assertion"`
+	ExpiresIn            int    `url:"expires_in,omitempty"`
```
hectorcast-db, line 15: Remove ExpiresIn and RefreshToken

httpclient/api_client.go
```diff
@@ -109,6 +109,7 @@ type DoOption struct {
 	in           RequestVisitor
 	out          func(body *common.ResponseWrapper) error
 	body         any
+	contentType  string
```
mgyucht, line 112: This duplicates information in the headers. I would just pull `Content-Type` header if set, defaulting to `application/json`.

config/config.go
```diff
@@ -11,20 +11,22 @@ import (
...
 // authentication credentials for Databricks REST APIs
-type CredentialsProvider interface {
+type CredentialsStrategy interface {
 	// Name returns human-addressable name of this credentials provider name
 	Name() string
 
-	// Configure creates HTTP Request Visitor or returns nil if a given credetials
-	// are not configured. It returns an error if credentials are misconfigured.
+	// Configure creates CredentialsProvider or returns nil if a given credetials
+	// strategy ßare not configured. It returns an error if credentials are misconfigured.
```
mgyucht, line 27:
```suggestion
	// Configure creates CredentialsProvider or returns nil if a given credentials
	// strategy are not configured. It returns an error if credentials are misconfigured.
```

credentials/oauth_credentials_provider.go
```diff
@@ -0,0 +1,32 @@
+package credentials
+
+import (
+	"net/http"
+
+	"golang.org/x/oauth2"
+)
+
+type OAuthCredentialsProvider interface {
+	SetHeaders(r *http.Request) error
```
mgyucht, line 10: In case we add more methods to CredentialsProvider, this should probably inherit from that interface.
```suggestion
	CredentialsProvider
```

credentials/oauth_token.go
```diff
@@ -0,0 +1,8 @@
+package credentials
+
+type OAuthToken struct {
```
mgyucht, line 3: For those less familiar with OAuth, let's add links to the RFC & document each field

credentials/credentials_provider.go
```diff
@@ -0,0 +1,23 @@
+package credentials
+
+import (
+	"net/http"
+)
+
+type CredentialsProvider interface {
+	SetHeaders(r *http.Request) error
```
mgyucht, line 8: Can you add doccomments here?

credentials/oauth_credentials_provider.go
```diff
@@ -0,0 +1,32 @@
+package credentials
+
+import (
+	"net/http"
+
+	"golang.org/x/oauth2"
+)
+
+type OAuthCredentialsProvider interface {
```
mgyucht, line 9: Can you add doccomments here?

config/config.go
```diff
@@ -11,20 +11,22 @@ import (
...
 	"github.com/databricks/databricks-sdk-go/logger"
+	"golang.org/x/oauth2"
 )
 
-// CredentialsProvider responsible for configuring static or refreshable
+// CredentialsStrategy responsible for configuring static or refreshable
 // authentication credentials for Databricks REST APIs
-type CredentialsProvider interface {
+type CredentialsStrategy interface {
 	// Name returns human-addressable name of this credentials provider name
```
mgyucht, line 23:
```suggestion
	// Name returns human-addressable name of this credentials strategy
```

config/auth_m2m.go
```diff
@@ -22,7 +22,7 @@ func (c M2mCredentials) Name() string {
 	return "oauth-m2m"
 }
 
-func (c M2mCredentials) Configure(ctx context.Context, cfg *Config) (func(*http.Request) error, error) {
+func (c M2mCredentials) Configure(ctx context.Context, cfg *Config) (credentials.CredentialsProvider, error) {
```
mgyucht, line 25: minor suggestion: I would return the most specific type that you can. If someone uses this directly and wants to be able to call the Token method, they'll need to match on the type first. We already know that this is an OAuthCredentialsProvider.
```suggestion
func (c M2mCredentials) Configure(ctx context.Context, cfg *Config) (credentials.OAuthCredentialsProvider, error) {
```

config/auth_gcp_google_id.go
```diff
@@ -44,7 +45,8 @@ func (c GoogleDefaultCredentials) Configure(ctx context.Context, cfg *Config) (f
 		return nil, err
 	}
 	logger.Infof(ctx, "Using Google Default Application Credentials for Accounts API")
-	return serviceToServiceVisitor(inner, platform, "X-Databricks-GCP-SA-Access-Token"), nil
+	visitor := serviceToServiceVisitor(inner, platform, "X-Databricks-GCP-SA-Access-Token")
+	return credentials.NewCredentialsProvider(visitor), nil
```
mgyucht, line 49: Can we not use inner.Token here for OAuth?

config/auth_azure_client_secret.go
```diff
@@ -52,5 +52,6 @@ func (c AzureClientSecretCredentials) Configure(ctx context.Context, cfg *Config
 	managementEndpoint := env.AzureServiceManagementEndpoint()
 	inner := azureReuseTokenSource(nil, c.tokenSourceFor(ctx, cfg, aadEndpoint, env.AzureApplicationID))
 	management := azureReuseTokenSource(nil, c.tokenSourceFor(ctx, cfg, aadEndpoint, managementEndpoint))
-	return azureVisitor(cfg, serviceToServiceVisitor(inner, management, xDatabricksAzureSpManagementToken)), nil
+	visitor := azureVisitor(cfg, serviceToServiceVisitor(inner, management, xDatabricksAzureSpManagementToken))
+	return credentials.NewCredentialsProvider(visitor), nil
```
mgyucht, line 56: This one should also probably support OAuth through the `inner` token source?

config/config.go
```diff
@@ -204,7 +206,25 @@ func (c *Config) Authenticate(r *http.Request) error {
...
+		return nil, err
+	}
+	err = c.authenticateIfNeeded()
+	if err != nil {
+		return nil, err
+	}
+	if h, ok := c.credentialsProvider.(credentials.OAuthCredentialsProvider); ok {
+		return h.Token()
+	} else {
+		return nil, fmt.Errorf("OAuth Token can only be retrieved for OauthCredentialsProvider")
```
mgyucht, line 226: Slightly more useful error message for users
```suggestion
		return nil, fmt.Errorf("OAuth Token not supported for current auth type %s", c.AuthType)
```

config/config.go
```diff
@@ -204,7 +206,25 @@ func (c *Config) Authenticate(r *http.Request) error {
 	if err != nil {
 		return err
 	}
-	return c.auth(r)
+	return c.credentialsProvider.SetHeaders(r)
+}
+
+// Authenticate returns an OAuth token for the current configuration.
+// It will return an error if the CredentialsStrategy does not support OAuth tokens.
+func (c *Config) GetToken() (*oauth2.Token, error) {
```
mgyucht, line 214: This is the method that will be used in the follow-up PR?

mgyucht (commented):
Looking good, some initial comments.

httpclient/request.go
```diff
@@ -64,6 +64,23 @@ func WithRequestData(body any) DoOption {
 	}
 }
 
+// WithUrlEncodedData takes either a struct instance, map, string, bytes, or io.Reader plus
+// a content type, and sends it either as query string for GET and DELETE calls, or as request body
+// for POST, PUT, and PATCH calls. The content type is set to "application/x-www-form-urlencoded"
+// and the body is encoded as a query string.
+//
+// Experimental: this method may eventually be split into more granular options.
+func WithUrlEncodedData(body any) DoOption {
```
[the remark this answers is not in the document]
hectorcast-db: Marshall is done before applying visitors to the headers. I can have this just set the contentType variable, and then using the existing WithRequestHeaders for the headers. 
Reference:
https://github.com/databricks/databricks-sdk-go/pull/886/files#diff-92f653de407b0c4c5447f7270f30191ea84914da49514bf2ada63c26d1cace7dR151
Line 151: extract contentType from visitor
Line 153: use the contentType for marshalling
Line 157: call to "c.perform", which will apply the visitors to the headers

httpclient/api_client.go
hectorcast-db: As explained above, the headers have not been set yet (i.e., visitors have not been applied) when marshalling the data.

config/auth_m2m.go
hectorcast-db: Then `M2mCredentials` does not fulfill the `CredentialsStrategy` interface.  Not sure if there is a way around it.

config/auth_gcp_google_id.go
hectorcast-db: I think I just missed that one. Updated.

config/config.go
hectorcast-db: This is already being used in `GetOAuthToken` method in the Workspace Client
https://github.com/databricks/databricks-sdk-go/pull/886/files#diff-75f383f34048f57e9041c3a70e938f38345a52ca6473a7790c7069d801b9bacfR980
We will use that one in the follow-up PR

mgyucht (approved):

PR #888, opened 2024-04-11
[DRAFT] Dataplane generation sample

hectorcast-db:


service/serving/impl.go
```diff
@@ -75,6 +76,46 @@ type servingEndpointsImpl struct {
...
+	auth_visitor := func(r *http.Request) error {
+		r.Header.Set("Authorization", fmt.Sprintf("Bearer %s", dataPlaneToken.AccessToken))
+		return nil
+	}
+	opts := []httpclient.DoOption{}
+	opts = append(opts, httpclient.WithAuthVisitor(auth_visitor))
+	opts = append(opts, httpclient.WithRequestHeaders(headers))
+	opts = append(opts, httpclient.WithRequestData(request))
+	opts = append(opts, httpclient.WithResponseUnmarshal(queryEndpointResponse))
+
```
hectorcast-db, line 113: Wrong path. Use endpoint_url and use it's host.

PR #890, opened 2024-04-12
Add retries for "worker env WorkerEnvId(workerenv-XXXXX) not found"

mgyucht:
## Changes
Due to the eventually consistent nature of worker environment creation as a part of workspace setup, certain API requests can fail when made right after workspace creation. We have some exception handling for these in the SDKs, but a new case has appeared: "worker env WorkerEnvId(workerenv-XXXXX) not found" (see https://github.com/databricks/terraform-provider-databricks/issues/3452).

This PR addresses this issue. Furthermore, it moves the transient error messages into autogeneration so that there is a single source of truth that applies to all SDKs.

One other small change: I removed running unit tests from the autogeneration flow. It removes a small amount of convenience (users need to run `make test` after regenerating the SDK), but it speeds up the devloop when iterating on code generation, and it allows the release flow to continue to make a PR before failing. This PR can be modified as needed to fix any test failures or compilation failures, as per usual.

## Tests
Added a unit test to cover this.

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/.codegen.json
+++ b/.codegen.json
@@ -25,8 +25,7 @@
   "toolchain": {
     "required": ["go"],
     "post_generate": [
-      "go run github.com/vektra/mockery/v2@b9df18e0f7b94f0bc11af3f379c8a9aea1e1e8da",
-      "go test -short ./..."
+      "go run github.com/vektra/mockery/v2@b9df18e0f7b94f0bc11af3f379c8a9aea1e1e8da"
     ]
   }
 }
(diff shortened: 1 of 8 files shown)


tanmay-db (approved):

jiropardo:
Hello @mgyucht @tanmay-db , does this take into account that XXXXX changes depending on the worker env ? I may have missed this in [#3452](https://github.com/databricks/terraform-provider-databricks/issues/3452)

PR #910, opened 2024-05-06
[Fix] Infer Azure tenant ID if not set

mgyucht:
## Changes
In order to use Azure U2M or M2M authentication with the Databricks SDK, users must request a token from the correct Entra ID instance, specifically, the same tenant as the one that the underlying workspace or account belongs to. Otherwise, Databricks will reject a user's requests. However, with Azure CLI auth, it is possible that a user is logged into multiple tenants at the same time. Currently, the SDK uses the subscription ID from the configured Azure Resource ID for the workspace when issuing the `az account get-access-token` command. However, when users don't specify the resource ID, the SDK simply fetches a token for the active subscription for the user. If the active subscription is in a different tenant than the workspace, users will see an error such as:

```
io.jsonwebtoken.IncorrectClaimException: Expected iss claim to be: https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/, but was: https://sts.windows.net/e3fe3f22-4b98-4c04-82cc-d8817d1b17da/
```

This PR modifies Azure CLI and Azure SP credential providers to attempt to load the tenant ID of the workspace if not provided before authenticating. Currently, there are no unauthenticated endpoints that the tenant ID can be directly fetched from. However, the tenant ID is indirectly exposed via the redirect URL used when logging into a workspace. In this PR, we fetch the tenant ID from this endpoint and configure it if not already set.

Here, we lazily fetch the tenant ID only in the auth methods that need it. This prevents us from making any unnecessary requests if these Azure credential providers are not needed.

## Tests

- [ ] `make test` passing
- [ ] `make fmt` applied


config/command.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+	"os/exec"
+	"strings"
+
+	"github.com/databricks/databricks-sdk-go/logger"
+)
+
+// Run a command and return the output.
+func runCommand(ctx context.Context, cmd string, args []string) ([]byte, error) {
+	logger.Debugf(ctx, "Running command: %s %v", cmd, strings.Join(args, " "))
```
tanmay-db, line 13: Might be good to allow passing debug level defaulting to DEBUG. 


`````
    logMessage := fmt.Sprintf("Running command: %s %v", cmd, strings.Join(args, " "))

    // logLevel is enum for log levels
    switch logLevel {
    case LogLevelDebug:
        logger.Debugf(ctx, logMessage)
    case LogLevelInfo:
        logger.Infof(ctx, logMessage)
    case LogLevelWarn:
        logger.Warnf(ctx, logMessage)
    case LogLevelError:
        logger.Errorf(ctx, logMessage)
    default:
        logger.Infof(ctx, "Unspecified log level for: %s", logMessage)
    }

`````

tanmay-db (approved):

PR #914, opened 2024-05-07
Add traceparent header to enable distributed tracing.

mgyucht:
## Changes
Currently, it is difficult to correlate requests made by a client with a request made by a backend service, especially if the same request is made multiple times. The REST API supports the [Trace Context standard](https://www.w3.org/TR/trace-context/) which defines a standard way of propagating tracing information through HTTP headers.

This PR implements a traceparent generator to construct new traceparent headers in accordance with this standard. These traceparents are attached to the headers of each individual request. The resulting header is visible when debug logs are enabled and DATABRICKS_DEBUG_HEADERS is set.

## Tests
Added a unit test to ensure that new traceparents are set for every request.

It's hard to test that a debug log actually contains this header without rewriting a lot of the matching logic for debug logs, but I did need to remove traceparent from those unit tests, suggesting that debug logging does include traceparent (like any other header).

Manually tested that the traceparent set by the client is visible to the server by checking our internal access logs.

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -99,7 +99,7 @@ type Config struct {
 	// Use at your own risk or for unit testing purposes.
 	InsecureSkipVerify bool `name:"skip_verify" auth:"-"`
 
-	// Number of seconds for HTTP timeout. Default is 300 (5 minutes).
+	// Number of seconds for HTTP timeout. Default is 60 (1 minute).
 	HTTPTimeoutSeconds int `name:"http_timeout_seconds" auth:"-"`
 
 	// Truncate JSON fields in JSON above this limit. Default is 96.
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/databricks/databricks-sdk-go/common"
+	"github.com/databricks/databricks-sdk-go/httpclient/traceparent"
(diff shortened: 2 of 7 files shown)


httpclient/api_client.go
```diff
@@ -236,6 +237,9 @@ func (c *ApiClient) attempt(
 				return c.failRequest(ctx, "failed during request visitor", err)
 			}
 		}
+		// Set traceparent for distributed tracing.
```
hectorcast-db, line 240: Users running the SDK in a server can set up their own Traceparent using visitors. This code correctly doesn't override this. But it is important that this line is applied after the visitors. I would add a note about this to avoid moving this during a refactor, as well as a test to ensure that we don't override traceparent set by the users.

hectorcast-db (approved):
Added a note, otherwise it looks good.

config/config.go
```diff
@@ -99,7 +99,7 @@ type Config struct {
 	// Use at your own risk or for unit testing purposes.
 	InsecureSkipVerify bool `name:"skip_verify" auth:"-"`
 
-	// Number of seconds for HTTP timeout. Default is 300 (5 minutes).
+	// Number of seconds for HTTP timeout. Default is 60 (1 minute).
```
mgyucht, line 102: This was a typo from before, the default has always been 60 seconds.

PR #919, opened 2024-05-13
Log cancelled and failed requests

mgyucht:
## Changes
Currently, we do not log requests that failed to complete or that failed when the SDK started processing the response. This means that requests that timeout due to a client-side timeout or canceled context are not logged, meaning that there is no visibility in the log of these kinds of requests.

This PR changes the request processing and error handling to ensure that all requests are logged, even those that are cancelled or fail due to other errors. The general approach is to process as much of the response as possible until encountering an error, log what we've retrieved, and then do any post-processing.

Some changes are needed in the roundtrip stringer to handle the case of a nil response.

## Tests
Unit tests verify that canceled requests or nil responses are logged.

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/apierr/errors.go
+++ b/apierr/errors.go
@@ -224,6 +224,7 @@ func parseUnknownError(resp *http.Response, requestBody, responseBody []byte, er
 
 func MakeUnexpectedError(resp *http.Response, err error, requestBody, responseBody []byte) error {
 	rts := httplog.RoundTripStringer{
+		Request:                  resp.Request,
 		Response:                 resp,
 		Err:                      err,
 		RequestBody:              requestBody,
--- a/apierr/errors.go
+++ b/apierr/errors.go
@@ -223,8 +223,12 @@ func parseUnknownError(resp *http.Response, requestBody, responseBody []byte, er
 }
 
 func MakeUnexpectedError(resp *http.Response, err error, requestBody, responseBody []byte) error {
+	var req *http.Request
+	if resp != nil {
+		req = resp.Request
+	}
 	rts := httplog.RoundTripStringer{
-		Request:                  resp.Request,
+		Request:                  req,
 		Response:                 resp,
 		Err:                      err,
 		RequestBody:              requestBody,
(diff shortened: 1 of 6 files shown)


apierr/errors.go
```diff
@@ -224,6 +224,7 @@ func parseUnknownError(resp *http.Response, requestBody, responseBody []byte, er
 
 func MakeUnexpectedError(resp *http.Response, err error, requestBody, responseBody []byte) error {
 	rts := httplog.RoundTripStringer{
+		Request:                  resp.Request,
```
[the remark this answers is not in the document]
tanmay-db: It might be good to add a check as Pieter mentioned it's exported, since we were using response in line 227 before without null check, it should be safe:
`Response: resp,`

logger/httplog/round_trip_stringer.go
```diff
@@ -69,15 +69,17 @@ func (r RoundTripStringer) String() string {
...
 		sb.WriteString(fmt.Sprintf("Error: %s", r.Err))
+		return sb.String()
+	}
+
+	sb.WriteString(fmt.Sprintf("%s %s", r.Response.Proto, r.Response.Status))
+	// Only display error on this line if the response body is empty or the
+	// client failed to read the response body.
+	// Otherwise the response body will include details about the error.
+	if len(r.ResponseBody) == 0 && r.Err != nil {
+		sb.WriteString(fmt.Sprintf(" (Error: %s)", r.Err))
```
tanmay-db, line 82: a question, when length of response body is > 0 (line 86), is it always true that the Err would be nil?

tanmay-db (approved):

logger/httplog/round_trip_stringer.go
mgyucht: Not necessarily. We call the ErrorMapper function that parses the response into one of the SDK errors depending on the result. If the API responds with any of our standard error codes, `r.Err` will be set to that.

This is just refactored from above, so there is no behavior change here.

PR #935, opened 2024-06-03
[discuss] Don't generate shortcuts for list functions

pietern:
## Changes

The `ListByZZZ` functions don't use the same pagination as the main functions.

By definition, the shortcut functions are generated only if there are required request parameters. Therefore, they should either return the same paginator, or not exist at all.

## Tests

n/a


service/compute/library_utilities.go
```diff
@@ -183,11 +183,19 @@ func (a *LibrariesAPI) Wait(ctx context.Context, wait Wait,
 		o(&i)
 	}
 	result, err := retries.Poll(ctx, i.Timeout, func() (*ClusterLibraryStatuses, *retries.Err) {
-		status, err := a.ClusterStatusByClusterId(ctx, wait.ClusterID)
```
andrewnester, line 186: Is this the only place where we explicitly use shortcut?

PR #936, opened 2024-06-03
Add DataPlane API Support

hectorcast-db:
## Changes
Add DataPlane API Support

## Tests
Manual test:
```
input := []string{}
input = append(input, "Hello, I'm a language model,")
_, err := w.ServingEndpointsDataPlane.Query(context.Background(), serving.QueryEndpointInput{
	Name:   <...>,
	Inputs: input,
})
require.NoError(t, err)
```

- [x] `make test` passing
- [x] `make fmt` applied


.codegen/workspaces.go.tmpl
```diff
@@ -14,25 +14,12 @@ type WorkspaceClient struct {
...
-	{{range .Services}}{{if and (not .IsAccounts) (not .HasParent) (not .IsDataPlane)}}
+	{{range .Services}}{{if and (not .IsAccounts) (not .HasParent) }}
 	{{.Comment "    // " 80}}
 	{{.Name}} {{.Package.Name}}.{{.Name}}Interface
 	{{end}}{{end}}
 }
 
-// Returns a new OAuth scoped to the authorization details provided.
-// It will return an error if the CredentialStrategy does not support OAuth tokens.
-//
```
hectorcast-db, line 25: We don't need this on the WorkspaceClient, but the Client itself.

httpclient/request.go
```diff
@@ -33,6 +33,17 @@ func WithRequestHeaders(headers map[string]string) DoOption {
...
+// WithToken uses the specified golang.org/x/oauth2 token on a request
+func WithToken(token *oauth2.Token) DoOption {
+	visitor := WithRequestVisitor(func(r *http.Request) error {
+		auth := fmt.Sprintf("%s %s", token.TokenType, token.AccessToken)
+		r.Header.Set("Authorization", auth)
+		return nil
+	})
+	visitor.isAuthOption = true
+	return visitor
+}
```
renaudhartert-db, line 45: Alternatively, we could do the following:

```go
func WithToken(token *oauth2.Token) DoOption {
  return WithTokenSource(oauth2.StaticTokenSource(token))
}
```

Or maybe completely remove the function and use `WithTokenSource(oauth2.StaticTokenSource(token))` directly instead of `WithToken(token)`?

openapi/code/method.go
```diff
@@ -46,9 +46,27 @@ type Method struct {
 	wait       *openapi.Wait
 	pagination *openapi.Pagination
 	Operation  *openapi.Operation
+	DataPlane  *openapi.DataPlane
 	shortcut   bool
 }
 
+// Returns the fields which contains the DataPlane info. Each field is nested in the previous one.
```
renaudhartert-db, line 53: [nit] Starts the comment with the name of the function:

```go
// DataPlaneInfoFields returns the fields ...
``` 

Reference: https://go.dev/wiki/CodeReviewComments#comment-sentences

Note: could you also do this for the other new functions in this PR (e.g. `HasDataPlaneAPI`)?

service/oauth2/data_plane.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if o.tokens == nil {
+		o.tokens = make(map[string]*oauth2.Token)
+	}
+	key := o.generateKey(method, params)
+	info, infoOk := o.infos[key]
+	token, tokenOk := o.tokens[key]
+	if infoOk && tokenOk && token.Valid() {
+		return info.EndpointUrl, token, nil
+	}
```
renaudhartert-db, line 26: I think we would obtain the same behavior if we were to remove that if right? Assuming that this is correct, doing so will allow us to move the variable instantiation closer to where they are used:

```go
	info, infoOK := o.infos[key]
	if !infoOK {
		// do things ...
	}

	token, tokenOK := o.tokens[key]
	if !tokenOK || !token.Valid() {
		// do other things ...
	}
```

This also have the advantage of better following the Go practice of keeping the "normal flow" at the lowest level of indentation (see https://go.dev/wiki/CodeReviewComments#indent-error-flow).

Alternatively, you could use a `switch` statement to make it clear that you have three distinct code paths:

```go
	token, tokenOK := o.tokens[key]
	info, infoOK := o.infos[key]

	switch {
	case !infoOK:
		// do things ...
	case !tokenOK || !token.Valid():
		// do other things ...
	default:
		// happy path
	}
```

That `switch` statement should produce the same compiled code than what you would obtain with the `ifs`.

renaudhartert-db (commented):
@hectorcast-db, I left a few Go-style suggestions — please have a look. Happy to discuss these offline.

httpclient/request.go
hectorcast-db: If you look at it closely, `WithTokenSource` does not set `visitor.isAuthOption = true`.
This is because `IsAuthOption` is a new flag. If should be added there, but it requires delicate testing.

But we can still simplify it like this: 

```
func WithToken(token *oauth2.Token) DoOption {
  visitor := WithTokenSource(oauth2.StaticTokenSource(token))
  visitor.isAuthOption = true
  return visitor
}
```

renaudhartert-db: Oh I missed that, thanks for pointing it out!

.codegen/impl.go.tmpl
```diff
@@ -13,22 +15,71 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	infoGetter := func() (*oauth2.DataPlaneInfo, error) {
+		response, err := a.controlPlane.{{.Service.DataPlaneInfoMethod.PascalName}}(ctx, getRequest)
+		if err != nil {
+			return nil, err
+		}
+		if response.{{(index .DataPlaneInfoFields 0).PascalName}} == nil {
+			return nil, errors.New("resource does not support direct Data Plane access")
```
mgyucht, line 46: When would this case happen? Is this a bug in the API definition? Or is it possible that some model serving endpoints support direct-to-dataplane access and others don't?

.codegen/impl.go.tmpl
```diff
@@ -13,22 +15,71 @@ import (
...
+	}
+	endpointUrl, dataPlaneToken, err := a.GetDataPlaneDetails("{{.PascalName}}", getParams, refresh, infoGetter) 
+	if err != nil {
+		return nil, err
+	}
+	{{ template "make-header" . }}
+	opts := []httpclient.DoOption{}
+	opts = append(opts, httpclient.WithRequestHeaders(headers))
+	{{- template "response-var" . }}
+	{{if .Response}}opts = append(opts, httpclient.WithRequestData(request)){{end}}
```
mgyucht, line 66: `if .Request`?

service/oauth2/data_plane.go
```diff
@@ -0,0 +1,50 @@
...
+// DataPlaneHelper is a helper struct to fetch, refresh and manage DataPlane details and tokens.
+type DataPlaneHelper struct {
+	infos  map[string]*DataPlaneInfo
+	tokens map[string]*oauth2.Token
+}
+
+// GetDataPlaneDetails returns the endpoint URL and token. It returns a cached token if it is valid,
+// otherwise it refreshes the token and returns the new token.
+func (o *DataPlaneHelper) GetDataPlaneDetails(method string, params []string, refresh func(*DataPlaneInfo) (*oauth2.Token, error), infoGetter func() (*DataPlaneInfo, error)) (string, *oauth2.Token, error) {
+	if o.infos == nil {
```
mgyucht, line 18: I think we need a mutex to protect access to these fields, since they are stateful and can be called concurrently. Might be a good candidate for a double check, since this is only done once the first time a dataplane API is called for a service:
```go
if o.infos == nil {
  lock()
  if o.infos == nil {
    initialize
  }
  unlock()
}
```

service/oauth2/data_plane.go
```diff
@@ -0,0 +1,50 @@
...
+		o.infos = make(map[string]*DataPlaneInfo)
+	}
+	if o.tokens == nil {
+		o.tokens = make(map[string]*oauth2.Token)
+	}
+	key := o.generateKey(method, params)
+	info, infoOk := o.infos[key]
+	if !infoOk {
```
mgyucht, line 26: Ditto, I think we may need a lock here as well?

service/oauth2/data_plane.go
```diff
@@ -0,0 +1,50 @@
...
+		newInfo, err := infoGetter()
+		if err != nil {
+			return "", nil, err
+		}
+		o.infos[key] = newInfo
+		info = newInfo
+	}
+	token, tokenOk := o.tokens[key]
```
mgyucht, line 34: And here too.

mgyucht (commented):
Looking really good! Some feedback but I think we're close to the final cut here.

.codegen/impl.go.tmpl
hectorcast-db: DataPlane access needs to be enabled on a resource level. If you call this method for a Model which is not "optimized" (their flag for DataPlane), this field won't be set (the endpointURL does not exists)

service/oauth2/data_plane.go
hectorcast-db: Done for the ones below. For the maps, I created a constructor method.

service/catalog/model.go
```diff
@@ -745,11 +745,6 @@ func (s CreateConnection) MarshalJSON() ([]byte, error) {
 	return marshal.Marshal(s)
 }
 
-type CreateEndpointRequest struct {
```
mgyucht: Are these deletions expected?

.codegen/api.go.tmpl
```diff
@@ -114,6 +114,7 @@ controlPlane *{{.ControlPlaneService.PascalName}}API,
 		impl: &{{.CamelName}}Impl{
 			client: client,
 			{{- if .IsDataPlane}}
+			DataPlaneService: oauth2.NewDataPlaneService(),
```
mgyucht, line 117: Do we need to export this?

mgyucht (approved):
Two small notes, otherwise LGTM!

.codegen/api.go.tmpl
hectorcast-db: Removed

PR #950, opened 2024-06-19
Add a credentials provider for Github Azure OIDC

renaudhartert-db:
_Note: this PR could not be merged because of its first unverified commit, please see PR #965._

## Changes

This PR adds a `CredentialsProvider` to authenticate with Azure from Github workflows. 

The code is inspired by a similar feature already implemented in the Python SDK. It works as follows:

1. Obtain an ID token from Azure leveraging the env variables `ACTIONS_ID_TOKEN_REQUEST_URL` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN` as [explained here](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers).
2. Exchange that ID token for an auth token.

## Tests

Added a test suite which covers all the added code paths. I've also confirmed in my own Github Action that the code is properly able to authenticate. 

Note: I'm not super happy with how errors are compared (i.e. using a prefix) which is a little brittle. A better approach would be to leverage `errors.As` or `errors.Is`. However, it is difficult to do that at the moment without adding ad hoc new error types. A longer term solution would probably involve standardizing the package around a set of clearly defined error types shared by all implementations of `CredentialsProvider` in `config`. That is out of the scope of this PR though.

- [x] `make test` passing
- [x] `make fmt` applied


config/auth_azure_github_oidc_test.go
```diff
@@ -0,0 +1,287 @@
...
+			}
+		})
+	}
+}
+
+func TestAzureGithubOIDCCredentials_Name(t *testing.T) {
+	c := AzureGithubOIDCCredentials{}
+	want := "github-oidc-azure"
+
+	if got := c.Name(); got != want {
```
mgyucht, line 284: Just curious, why not `assert.Equal(t, "github-oidc-azure", c.Name())`?

config/auth_azure_github_oidc_test.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	return strings.HasPrefix(err.Error(), prefix)
+}
+
+func TestAzureGithubOIDCCredentials(t *testing.T) {
+	testCases := []struct {
+		desc         string
+		cfg          *Config
+		wantHeaders  map[string]string
+		wantErrPefix *string
```
mgyucht, line 29: nit: `wantErrPrefix`?

mgyucht (approved):
Overall this seems good to me, great test cases as well! One small suggestion for logging when we can't successfully fetch the credential.

I think this is great for now. I'm picturing something in the future that there is a more generalized workflow of:
1. Discover the current environment the SDK is in (github actions, azure devops, etc).
2. Fetch the ID token according to the local environment (query the Github endpoint, somehow gather the ID token from Azure DevOps, etc.)
3. Exchange the ID token for an OAuth token (using Azure EntraID endpoint, using Databricks OAuth API, etc.)

Essentially, we can probably generalize this to support OIDC strategies in the future by decoupling the ID token fetching from the token exchange endpoint. That said, let's not let perfect be the enemy of good, and we can make that change when we do work to support Azure DevOps, for example.

config/auth_azure_github_oidc_test.go
renaudhartert-db: Oh, good catch!

config/auth_azure_github_oidc_test.go
renaudhartert-db: It's a good question. The short answer is that using `testify` is not idiomatic. As a result, one could argue that using `testify` actually goes against Databricks' Go style guidelines (see refs below). 

IMO, `testify` can be quite convenient in some use cases but it adds cognitive load for the developers and the readers that are not familiar with the library (i.e. it is one additional mini-language to learn). Adding that cognitive load doesn't seem to be worth the couple of lines it removes.

Some references:

- https://google.github.io/styleguide/go/decisions#assertion-libraries (not in Databricks' go style)
- https://go.dev/wiki/CodeReviewComments#useful-test-failures

config/auth_azure_github_oidc.go
```diff
@@ -0,0 +1,117 @@
...
+	if !cfg.IsAzure() || cfg.AzureClientID == "" || cfg.Host == "" || cfg.AzureTenantID == "" {
+		return nil, nil
+	}
+
+	idToken, err := requestIDToken(ctx, cfg)
+	if err != nil {
+		return nil, err
+	}
+	if idToken == "" {
+		return nil, nil
```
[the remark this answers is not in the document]
renaudhartert-db: Good idea, added some logging in the underlying function to provide more context.

renaudhartert-db:
Thanks for the review @mgyucht! Please have a look at my answers and feel free to re-open the discussions.

renaudhartert-db:
> I'm picturing something in the future that there is a more generalized workflow of:
1. Discover the current environment the SDK is in (github actions, azure devops, etc).
2. Fetch the ID token according to the local environment (query the Github endpoint, somehow gather the ID token from Azure DevOps, etc.)
3. Exchange the ID token for an OAuth token (using Azure EntraID endpoint, using Databricks OAuth API, etc.)

Agreed, this is also how I picture the evolution of this package.

PR #963, opened 2024-07-04
Added more error messages for retriable errors (timeouts, etc.)

alexott:
These are necessary to handle a problem with timeouts in permissions fetching

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/openapi/errors.go
+++ b/openapi/errors.go
@@ -75,4 +75,6 @@ var TransientErrorRegexes = []*regexp.Regexp{
 	regexp.MustCompile(`Unknown worker environment`),
 	regexp.MustCompile(`ClusterNotReadyException`),
 	regexp.MustCompile(`worker env .* not found`),
+	regexp.MustCompile(`Timed out after `),
+	regexp.MustCompile(`deadline exceeded`),
 }


apierr/error_overrides.go
```diff
@@ -33,4 +33,6 @@ var allTransientErrors = []*regexp.Regexp{
 	regexp.MustCompile(`Unknown worker environment`),
 	regexp.MustCompile(`ClusterNotReadyException`),
 	regexp.MustCompile(`worker env .* not found`),
+	regexp.MustCompile(`Timed out after `),
```
mgyucht, line 36: You need to add these to `openapi/errors.go`, otherwise they will be removed during autogeneration.

mgyucht (changes requested):
Change seems OK but this is an autogenerated file. You just need to update the source of truth.

apierr/error_overrides.go
alexott: 🤦 fixed...

openapi/errors.go
```diff
@@ -75,4 +75,6 @@ var TransientErrorRegexes = []*regexp.Regexp{
 	regexp.MustCompile(`Unknown worker environment`),
 	regexp.MustCompile(`ClusterNotReadyException`),
 	regexp.MustCompile(`worker env .* not found`),
+	regexp.MustCompile(`Timed out after `),
+	regexp.MustCompile(`deadline exceeded`),
```
mgyucht, line 79: Where does this one come from btw?

alexott: First one - from permissions as I remember, 2nd one - from SCIM as I remember

mgyucht (approved):

openapi/errors.go
```diff
@@ -75,4 +75,6 @@ var TransientErrorRegexes = []*regexp.Regexp{
 	regexp.MustCompile(`Unknown worker environment`),
 	regexp.MustCompile(`ClusterNotReadyException`),
 	regexp.MustCompile(`worker env .* not found`),
+	regexp.MustCompile(`Timed out after `),
```
tanmay-db, line 78: There is space after `after`, is it expected?

tanmay-db (approved):

openapi/errors.go
alexott: yes - to show that there is an argument after it

PR #984, opened 2024-07-16
[Fix] Avoid loading the response bodies twice in memory when parsing `bytes.Buffer`

renaudhartert-db:
## Changes

Prior to this PR, `WithResponseUnmarshal` was unmarshalling a response body to a `bytes.Buffer` by:

1. Reading all the response body in a slice of bytes; and
2. Writing that slice in the receiving `bytes.Buffer`.

Step 1 unnecessarily increases the memory footprint as the response body is essentially stored twice in memory during unmarshalling. This PR fixes that problem by loading the body directly in the buffer without passing by an intermediate container. 

## Tests

Added a few unit test to cover all unmarshalling types.

Memory benchmark:

```go
var longString = strings.Repeat("a", 1<<30) 

func BenchmarkWithResponseUnmarshal_byteBuffer(b *testing.B) {
	b.ReportAllocs()

	client := NewApiClient(ClientConfig{
		Transport: hc(func(r *http.Request) (*http.Response, error) {
			return &http.Response{
				StatusCode: 200,
				Body:       io.NopCloser(strings.NewReader(longString)),
				Request:    r,
			}, nil
		}),
	})

	for i := 0; i < b.N; i++ {
		var got bytes.Buffer
		client.Do(context.Background(), "GET", "/a", WithResponseUnmarshal(&got))
	}
}
```

Results:

```
Before:	6810268826 B/op	140 allocs/op
After:	4294972868 B/op	101 allocs/op
```

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/httpclient/response.go
+++ b/httpclient/response.go
@@ -84,6 +84,9 @@ func WithResponseUnmarshal(response any) DoOption {
 			if err != nil {
 				return fmt.Errorf("failed to read response body: %w", err)
 			}
+			if len(bodyBytes) == 0 {
+				return nil
+			}
 			if bs, ok := response.(*[]byte); ok {
 				*bs = bodyBytes
 				return nil
--- a/httpclient/response.go
+++ b/httpclient/response.go
@@ -32,6 +32,10 @@ func WithResponseHeader(key string, value *string) DoOption {
 //   - *[]byte,
 //   - a pointer to a struct with a Contents io.ReadCloser field,
 //   - a pointer to a struct representing a JSON object.
+//
+// If response is a pointer to a io.ReadCloser or a struct with a io.ReadCloser
+// field name "Contents", then the response io.ReadCloser is set to the value of
+// the body's reader without actually reading it.
 func WithResponseUnmarshal(response any) DoOption {
 	return DoOption{
 		in: func(r *http.Request) error {
@@ -58,22 +62,14 @@ func WithResponseUnmarshal(response any) DoOption {
 				return err
 			}
 
-			// If the response is a struct with a field "Contents" of type
-			// io.ReadCloser, set that field to the response body ReadCloser.
 			if field, ok := findContentsField(response); ok {
 				field.Set(reflect.ValueOf(body.ReadCloser))
 				return nil
 			}
-
-			// If the response is a ReadCloser, return it directly without
-			// reading the body.
-			if rc, ok := response.(*io.ReadCloser); ok {
-				*rc = body.ReadCloser
+			if reader, ok := response.(*io.ReadCloser); ok {
+				*reader = body.ReadCloser
 				return nil
 			}
-
-			// If the response is a bytes.Buffer, read the entire body directly
-			// into it.
 			if buffer, ok := response.(*bytes.Buffer); ok {
 				defer body.ReadCloser.Close()
 				_, err := buffer.ReadFrom(body.ReadCloser)
--- a/httpclient/response.go
+++ b/httpclient/response.go
@@ -25,6 +25,13 @@ func WithResponseHeader(key string, value *string) DoOption {
 	}
 }
 
+// WithResponseUnmarshal unmarshals the response body into response. The
+// supported response types are the following:
+//   - *bytes.Buffer,
+//   - *io.ReadCloser,
+//   - *[]byte,
+//   - a pointer to a struct with a Contents io.ReadCloser field,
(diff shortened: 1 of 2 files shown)


httpclient/response_test.go
```diff
@@ -1,29 +1,120 @@
...
+		Contents: io.NopCloser(strings.NewReader("foo bar")),
+	}
+
+	var got structWithContents
+	c := mockClient(make200Response("foo bar"))
+	gotErr := c.Do(context.Background(), "GET", "/a", WithResponseUnmarshal(&got))
+
+	if gotErr != nil {
+		t.Errorf("WithResponseUnmarshal(): want no error, got: %s", gotErr)
+	}
```
pietern, line 45: You can use the `require` and/or `assert` packages for a shorthand.

```suggestion
	gotErr := c.Do(context.Background(), "GET", "/a", WithResponseUnmarshal(&got))
	require.NoError(t, gotErr, "WithResponseUnmarshal")
```

pietern (commented):
The change LGTM, but were we writing into a `bytes.Buffer` anywhere?

httpclient/response_test.go
renaudhartert-db: I tend to avoid using libraries like `testify` because they are not idiomatic and are arguably against the Databricks' Go style guidelines (see refs below).

IMO, testify can be quite convenient in some use cases but it adds cognitive load for the developers and the readers that are not familiar with the library (i.e. it is one additional mini-language to learn). Adding that cognitive load doesn't seem to be worth the couple of lines it removes.

Some references:

- https://google.github.io/styleguide/go/decisions#assertion-libraries (not in Databricks' go style)
- https://go.dev/wiki/CodeReviewComments#useful-test-failures

renaudhartert-db:
> The change LGTM, but were we writing into a `bytes.Buffer` anywhere?

It doesn't seem that we do in the SDK. I looked into that issue as a follow-up to https://github.com/databricks/cli/issues/1575 assuming that there was some upstream usage but that might not be the case.

httpclient/response.go
```diff
@@ -25,6 +25,17 @@ func WithResponseHeader(key string, value *string) DoOption {
...
+// WithResponseUnmarshal unmarshals the response body into response. The
+// supported response types are the following:
+//   - *bytes.Buffer,
+//   - *io.ReadCloser,
+//   - *[]byte,
+//   - a pointer to a struct with a Contents io.ReadCloser field,
+//   - a pointer to a struct representing a JSON object.
+//
+// If response is a pointer to a io.ReadCloser or a struct with a io.ReadCloser
+// field name "Contents", then the response io.ReadCloser is set to the value of
```
mgyucht, line 37: The key thing to note here is that currently, our autogeneration _only_ includes a `Contents io.ReadCloser` field for APIs that return non-JSON responses. We don't attempt to parse these at all and simply return them to the caller.

mgyucht (approved):
One small suggestion on the test cases (which are very nice), otherwise LGTM! Thanks for this fix!

httpclient/response_test.go
```diff
@@ -10,20 +11,87 @@ import (
...
+		Contents: io.NopCloser(strings.NewReader("foo bar")),
+	}
+
+	var got structWithContents
+	gotErr := c.Do(context.Background(), "GET", "/a", WithResponseUnmarshal(&got))
+
+	require.NoError(t, gotErr)
+	wantBytes, _ := io.ReadAll(want.Contents)
+	gotBytes, _ := io.ReadAll(got.Contents)
+	require.Equal(t, wantBytes, gotBytes)
```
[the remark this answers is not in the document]
renaudhartert-db: Oh, that's funny: I actually used `require` because I thought it was the opposite and that `assert` will be stopping the test!

httpclient/response_test.go
renaudhartert-db: From offline discussion: we decided to keep using `testify` as it is the most common practice in the Databrick's SDK and CLI.

pietern (approved):

renaudhartert-db:
For the records: I would have loved to use a table tests here instead of having 5 different test functions. The reason I didn't do it are:

- Added complexity to either infer and init `got` from `want` (or declare `got` in the test case).
- I could not find an obvious and simple way to compare `got` and `want` that would have worked for all the tested types.

PR #985, opened 2024-07-16
[Internal] Use `assert` instead of `require`

renaudhartert-db:
## Changes

Replace all usages of `require` with `assert` in test files. The rationale is that `require` fails the test immediately while `assert` keeps going and prints all the failed checks. 

For reference https://google.github.io/styleguide/go/decisions#keep-going

- [x] `make test` passing
- [x] `make fmt` applied


client/client_test.go
```diff
@@ -221,11 +220,11 @@ func TestHttpTransport(t *testing.T) {
 		}),
 	}
 	client, err := New(cfg)
-	require.NoError(t, err)
+	assert.NoError(t, err)
```
pietern, line 223: Cases like this will cause a panic because `client` will be `nil`.

The point of require is to stop executing because you know the rest of the test won't work anyway.

renaudhartert-db: I see, I didn't consider that we would branch on the `testify` assertions — that was naive. Thanks for being diligent, I'll close the PR.

PR #986, opened 2024-07-16
[Internal] Trigger the `validate` workflow in the merge queue

renaudhartert-db:
## Changes

PRs in the merge queue are currently unable to be merged (and thus dequeued) because the `validate` workflow is not triggered. This PR fixes the problem by adding the missing triggering event.

image

See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group

## Tests

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/.github/workflows/message.yml
+++ b/.github/workflows/message.yml
@@ -3,6 +3,8 @@ name: Validate Commit Message
 on:
   pull_request:
     types: [opened, synchronize, edited]
+  merge_group:
+    types: [checks_requested]
 
 jobs:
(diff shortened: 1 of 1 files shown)


hectorcast-db (approved):

hectorcast-db:
The test fails in the merge queue, because there is no PR Title in the event. 
We should just skip the job itself if it's not a PR trigger
```
if: github.event_name != 'pull_request'
```

PR #990, opened 2024-07-18
[Internal] Use HTTP status text as message if the error response body is empty

pietern:
## Changes

Errors with an empty response body now show up as empty strings when serialized to a string.

## Tests

Modified existing unit test.


renaudhartert-db (approved):
LGTM

PR #991, opened 2024-07-18
[Internal] Update OpenAPI spec

hectorcast-db:
## Changes
Update OpenAPI spec

## Tests

- [X] `make test` passing
- [X] `make fmt` applied


internal/billing_test.go
```diff
@@ -81,62 +81,84 @@ func TestMwsAccLogDelivery(t *testing.T) {
 }
 
 func TestMwsAccBudgets(t *testing.T) {
+	// Currently, the API returns errors
```
hectorcast-db, line 84: Not only for the SDK. Even curl returns 400 when using Oauth tokens. Investigation ongoind.

mgyucht (approved):
LGTM. I think we are going to have to include some notes in the changelog for this release communicating the most relevant changes, e.g. renaming the Queries, QueryVisualizations, and Alerts services to ...Legacy.

PR #993, opened 2024-07-22
[Internal] Pin jobs APIs to 2.1 in SDKs

mgyucht:
## Changes
Pin certain endpoints in jobs API to use API 2.1 after the public documentation for Jobs migrates to the 2.2 API version.

## Tests
Unit tests verify that only specific Jobs API paths are updated.

- [ ] `make test` passing
- [ ] `make fmt` applied


renaudhartert-db (approved):
Code and tests look good to me. I haven't verified the claim about the API interoperability though.

PR #994, opened 2024-07-22
[Internal] Log a warning when declaring inline entities 

renaudhartert-db:
## Changes

Inlined anonymous entities (object and enum) is discouraged. Rather, the proper way to define these entities is to declare them explicitly in the component part of the OpenAPI specification. 

This PR adds a warning to inform users that their schema contains the discouraged pattern.

This PR also includes a couple of noop refactoring to improve readability:
- Move the part responsible for handling references in its own function.
- Use a switch to emphasize that an entity can only be one of a fixed set of types. 

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/openapi/code/package.go
+++ b/openapi/code/package.go
@@ -5,6 +5,7 @@ package code
 import (
 	"context"
 	"fmt"
+	"log"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


openapi/code/package.go
```diff
@@ -162,40 +134,80 @@ func (pkg *Package) schemaToEntity(s *openapi.Schema, path []string, hasName boo
 	if s.JsonPath != "" {
 		processedEntities[s.JsonPath] = e
 	}
-	// pull embedded types up, if they can be defined at package level
-	if s.IsDefinable() && !hasName {
```
hectorcast-db, line 166: Why not using `IsDefinable`? I also don't see other usages if that function. We may want to remove it.

hectorcast-db (approved):

openapi/code/package.go
renaudhartert-db: I think keeping the function makes the code less readable. By having the code like below, you know in one go that (i) it only applies to `object` and `enum`, and more importantly (ii) that the inner if statement is handling all the cases and is correct. 

```go
if (s.IsObject() || s.IsEnum()) && !hasName {
		fieldType := "enum"
		if s.IsObject() {
			fieldType = "object"
		}
```

I didn't realize that the function was only called there so yes, I think it's better to remove it.

PR #997, opened 2024-07-24
[Internal]  Generate SDK from Open API 

Divyansh-db:
## Changes
Generate SDK from Open API


apierr/error_overrides.go
```diff
@@ -33,6 +33,4 @@ var allTransientErrors = []*regexp.Regexp{
 	regexp.MustCompile(`Unknown worker environment`),
 	regexp.MustCompile(`ClusterNotReadyException`),
 	regexp.MustCompile(`worker env .* not found`),
-	regexp.MustCompile(`Timed out after `),
```
hectorcast-db, line 36: This should not be changed. If they do, it means that you are not using the latest deco tool. Update it and rerun the generation.

hectorcast-db (approved):

PR #1000, opened 2024-07-29
[Internal] Enable mixins via struct embedding

renaudhartert-db:
## Changes

This PR changes the template to auto-generate Go code to enable mixins by leveraging Go embedding. 

One positive side-effect of this PR is that it significantly reduces the amount of auto-generated code.

After this change, methods can be overridden by adding the overriding method to the `ServiceNameAPI` struct. For example:

```go
// -------------------------------------------------------
// Generated code.
// -------------------------------------------------------

type FooService interface {
	Foo() string
	Bar() string
}

type FooAPI struct {
	FooService
}

type fooImpl struct{}

func (impl *fooImpl) Foo() string {
	return "foo"
}

func (impl *fooImpl) Bar() string {
	return "bar"
}

// -------------------------------------------------------
// Method that shadows the original Foo method inherited
// from fooImpl.
// -------------------------------------------------------

func (api *FooAPI) Foo() string {
	return fmt.Sprintf("extension: %s", api.fooImpl.Foo())
}
```

## Tests

Run unit tests and integrations tests. 

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/.codegen/api.go.tmpl
+++ b/.codegen/api.go.tmpl
@@ -111,7 +111,7 @@ controlPlane *{{.ControlPlaneService.PascalName}}API,
 {{end -}}
 ) *{{.PascalName}}API {
 	return &{{.PascalName}}API{
-		impl: &{{.CamelName}}Impl{
+		{{.PascalName}}Service: &{{.CamelName}}Impl{
 			client: client,
 			{{- if .IsDataPlane}}
 			dataPlaneService: oauth2.NewDataPlaneService(),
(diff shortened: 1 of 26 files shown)


mgyucht (commented):
What is the implication of this change for documentation? Are methods from the embedded structure visible on the API-level structs? Or would users be linked to the embedded struct's documentation?

mgyucht (approved):
LGTM for this change. In a follow-up, we should remove the deprecated "Service" types  & WithImpl()/Impl() methods and embed the underlying struct to improve doc experience.

PR #1004, opened 2024-08-01
[Fix] Allowed filtering for graviton DBRs in `ClusterAPI.SelectSparkVersion`

nkvuong:
## Changes
- We need to filter out Graviton DBRs, otherwise `SelectSparkVersion` will fail with `spark versions query returned multiple results`

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


renaudhartert-db:
Hi @nkvuong, could you resolve the conflict in your change?

renaudhartert-db (commented):
I don't have major concerns about the change itself. That being said, it would be great if we could update the documentation of this function to match its current name and maybe clarify a little what the function is doing for users with limited context.

renaudhartert-db (approved):
Thanks!

PR #1009, opened 2024-08-07
[Internal] Update Jobs GetRun API to support paginated responses for jobs and ForEach tasks with >100 runs

gkiko10:
## Changes
Jobs GetRun 2.2 behind the scenes pagination

## Tests

- [x] `make test` passing
- [x] `make fmt` applied


ricardoc-db (approved):
Not a golang expert, but logic looks OK.

service/jobs/ext_api.go
```diff
@@ -0,0 +1,30 @@
+package jobs
+
+import "context"
+
+func (a *JobsAPI) GetRun(ctx context.Context, request GetRunRequest) (*Run, error) {
```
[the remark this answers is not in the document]
renaudhartert-db: It uses Go embedding: https://go.dev/doc/effective_go#embedding. The public methods from `jobsImpl` are automatically promoted to the interface of the wrapper except if the wrapper already has a method with the same name.

service/jobs/ext_api.go
```diff
@@ -0,0 +1,30 @@
...
+	run, err := a.jobsImpl.GetRun(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	isPaginatingIterations := run.Iterations != nil && len(run.Iterations) > 0
+
+	for len(run.NextPageToken) != 0 {
```
dvkuznietsov, line 13: curious - NextPageToken is non-nullable in Go SDK?

dvkuznietsov (approved):

service/jobs/ext_api.go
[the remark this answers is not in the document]
ricardoc-db: Ah, cool. 👍

service/jobs/ext_api_test.go
```diff
@@ -0,0 +1,299 @@
...
+		mockJobsImpl := &jobsImpl{
+			client: client,
+		}
+		api := &JobsAPI{jobsImpl: *mockJobsImpl}
+
+		request := GetRunRequest{RunId: 514594995218126}
+		run, err := api.GetRun(ctx, request)
+
+		assert.NoError(t, err)
+		assert.EqualValues(t, 123, run.Tasks[0].RunId)
```
mgyucht, line 225: For this test, I would also add assertions on the length of the `Tasks` and `Iterations` fields for clarity.

service/jobs/ext_api_test.go
```diff
@@ -0,0 +1,299 @@
...
+		api := &JobsAPI{jobsImpl: *mockJobsImpl}
+
+		request := GetRunRequest{RunId: 111222333}
+		run, err := api.GetRun(ctx, request)
+
+		assert.NoError(t, err)
+		assert.Equal(t, 4, len(run.Tasks))
+		assert.Empty(t, run.Iterations)
+		assert.Empty(t, run.NextPageToken)
+		assert.Empty(t, run.PrevPageToken)
```
mgyucht, line 245: Not now, but longer-term, we may want to hide these fields from the SDK in general. cc @hectorcast-db @renaudhartert-db

service/jobs/ext_api_test.go
```diff
@@ -0,0 +1,299 @@
...
+					Tasks: []RunTask{
+						{
+							RunId: 999,
+						},
+					},
+					PrevPageToken: "token1-reverse",
+				},
+			},
+		},
+		//API 2.1 stubs
```
mgyucht, line 128: Are these used in any test?

service/jobs/ext_api_test.go
```diff
@@ -0,0 +1,299 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/databricks/databricks-sdk-go/qa"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func commonFixtureWithStatusResponse() qa.HTTPFixtures {
```
mgyucht, line 12: IMO the common fixture like this is less readable. Each test depends only on one or two request mocks. In that case, I think they should be next to the test itself. Otherwise, the reader needs to bounce back and forth between the test and this definition. For example:
```go
func Test...() {
    t.Run("my test", func() {
        mock := []qa.HTTPFixture{...}
        client, server := mock.Client(t)
        ...
    })
}
```

service/jobs/ext_api_test.go
```diff
@@ -0,0 +1,299 @@
...
+	return append(
+		// API 2.2 stubs
+		[]qa.HTTPFixture{
+			{
+				Method:       "GET",
+				ReuseRequest: true,
```
mgyucht, line 18: You don't need to set `ReuseRequest` on any of these mocks, since each request is called only once per client instance (i.e. per test).

mgyucht (commented):
First round of comments. Mostly looks good, my main feedback is about the tests/comments but the behavior seems OK. Were you able to manually test this? 

@dvkuznietsov @gkiko10: I think we should think carefully about whether we want to include the non-paginated repeated fields that have a fixed value in every response (i.e. that the same JobClusters are returned in each GetRun request for different pages). This means that if you want to add pagination for any other fields in the response (like JobClusters), that would be a behavior change of your API, and we wouldn't be able to release the client code in the SDKs until that behavior change is released everywhere. Otherwise, those fields would contain a copy of the nested resources for each requested page (e.g. if there was 1 job cluster but five pages of tasks, the SDK would return a run with 5 job clusters). If we get this behavior right from the start, we can avoid this.

mgyucht (approved):

vsamoilov (approved):

renaudhartert-db (approved):

PR #1016, opened 2024-08-14
[Internal] Added Service.NamedIdMap

andrewnester:
## Changes
This new function on the service allows to get access to NameIdMap from service instead of list method as it was done previously. 

Previously in code generation relying on `.List.NameIdMap` could lead to undetermenistic behaviour if there were more than 1 List operation on service. Why this is allowed, service can have only 1 name to id method anyway.

Thus, new `Service.NamedIdMap` will find `NamedIdMap` from all methods it has and if there are more than 1 - will panic. 

## Tests
Used here: 

- [x] `make test` passing
- [x] `make fmt` applied

### DIFF
--- a/openapi/code/service.go
+++ b/openapi/code/service.go
@@ -99,9 +99,38 @@ func (svc *Service) Methods() (methods []*Method) {
 	return methods
 }
 
+// NamedIdMap allows to retrieve a special NamedIdMap object from all methods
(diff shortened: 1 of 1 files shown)


renaudhartert-db (approved):
LGTM. It's a shame that the code of the current library doesn't allow us to handle this properly with an error but that is out of the scope of this PR.

https://go.dev/wiki/CodeReviewComments#dont-panic

PR #1021, opened 2024-08-20
[Fix] Do not specify `--tenant` flag when fetching managed identity access token from the CLI

mgyucht:
## Changes
Addresses https://github.com/databricks/terraform-provider-databricks/issues/3918.

The Azure CLI's `az account get-access-token` command does not allow specifying `--tenant` flag if it is authenticated via the CLI.

See https://github.com/hashicorp/go-azure-sdk/pull/910 for context; this implementation mimics the work done there.

## Tests
Unit tests ensure that all expected cases are treated as managed identities.

Added an integration test to verify that this works on machines using MSI with the Azure CLI installed.

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/config/auth_azure_cli_test.go
+++ b/config/auth_azure_cli_test.go
@@ -255,10 +255,10 @@ func TestAzureCliCredentials_DoNotSpecifyTenantIdWithMSI(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.userName, func(t *testing.T) {
 			env.CleanupEnvironment(t)
-			os.Setenv("PATH", testdataPath())
-			os.Setenv("FAIL_IF_TENANT_ID_SET", "true")
-			os.Setenv("AZ_USER_NAME", c.userName)
-			os.Setenv("AZ_USER_TYPE", "servicePrincipal")
+			t.Setenv("PATH", testdataPath())
+			t.Setenv("FAIL_IF_TENANT_ID_SET", "true")
+			t.Setenv("AZ_USER_NAME", c.userName)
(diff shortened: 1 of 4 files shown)


config/auth_azure_cli_test.go
```diff
@@ -245,6 +245,30 @@ func TestAzureCliCredentials_DoNotFetchIfTenantIdAlreadySet(t *testing.T) {
...
+		{"systemAssignedIdentity"},
+		{"userAssignedIdentity"},
+	}
+	for _, c := range cases {
+		t.Run(c.userName, func(t *testing.T) {
+			env.CleanupEnvironment(t)
+			os.Setenv("PATH", testdataPath())
+			os.Setenv("FAIL_IF_TENANT_ID_SET", "true")
+			os.Setenv("AZ_USER_NAME", c.userName)
+			os.Setenv("AZ_USER_TYPE", "servicePrincipal")
```
renaudhartert-db, line 261: Prefer using `t.Setenv` ([doc](https://pkg.go.dev/testing#T.Setenv))

```suggestion
			t.Setenv("PATH", testdataPath())
			t.Setenv("FAIL_IF_TENANT_ID_SET", "true")
			t.Setenv("AZ_USER_NAME", c.userName)
			t.Setenv("AZ_USER_TYPE", "servicePrincipal")
```

mgyucht: Awesome, thanks for that tip!

hectorcast-db (approved):

PR #1033, opened 2024-08-30
[Fix] Fail fast when authenticating if host is not configured

mgyucht:
## Changes
The SDK currently hangs when making API requests if the Host is set to a non-empty but invalid endpoint, such as `https://:443`. This PR adds a check during authentication to ensure that the hostname is defined, failing early if not. The underlying error is exported so other clients, such as the CLI or Terraform Provider, can provide specific advice when this happens.

## Tests
Added a unit test for this case.

- [ ] `make test` passing
- [ ] `make fmt` applied

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -373,6 +373,9 @@ func (c *Config) authenticateIfNeeded() error {
 	c.credentialsProvider = credentialsProvider
 	c.AuthType = c.Credentials.Name()
 	c.fixHostIfNeeded()
+	if err := c.assertHostIsSet(); err != nil {
```
tanmay-db, line 376: is this something that we should do inside `fixHostIfNeeded()`?

tanmay-db (approved):

config/config.go
mgyucht: Currently, fixHostIfNeeded is called before Configure() and from other places, and the Configure() credential providers are responsible for populating the host if the user specified an azure workspace resource ID. If we could separate those elements from one another, then we could have the following logic:

1. populate host if needed
2. fixHostIfNeeded: at this point, host must be set, so it can fail.
3. Configure()

However, because fixHostIfNeeded is called from different codepaths, it may not be safe to have this method error if host is not set.

mgyucht: Actually, fixHostIfNeeded is only called from CanonicalHostname() and this pathway, so I think it is better to do it as you describe. However, we can ignore ErrNoHostConfigured in the first call.

PR #1081, opened 2024-11-04
[Fix] Support custom retry logic per method

mgyucht:
## Changes
Some requests can be safely retried when the API returns a 504 Gateway Timeout, but not all. In this PR, we refactor the retry logic slightly to allow discriminating based on the request and response as well as any potential `error` returned from the Go HTTP client or parsed from the API response. The logic for this is consolidated into a single place, the `ErrorRetriable` field of the HTTP Client's config.

In addition, we add retries for the Get Permissions API, which can fail in this way, to provide better stability for callers of this endpoint. Today, we don't have a clear indication of which API methods are idempotent, so the approach is very conservative, only retrying requests whose method and path match a given pattern.

This PR incidentally rewrites the retry logic for the DefaultErrorRetriable, reading the status code directly from the `*http.Response` instead of from the parsed error. This is less brittle in case a custom error parser is provided that doesn't deserialize 4xx/5xx responses using the `HttpError` type.

## Tests
The first commit is a refactor of the existing code. All existing tests pass.

A new test is added that verifies that the Get Permissions API retries when the API responds with a 504. I verified that this fails on the initial commit.

- [ ] `make test` passing
- [ ] `make fmt` applied


config/api_client.go
```diff
@@ -73,17 +75,22 @@ func (c *Config) NewApiClient() (*httpclient.ApiClient, error) {
...
-		},
+		ErrorRetriable: httpclient.CombineRetriers(
+			func(ctx context.Context, _ *http.Request, _ *common.ResponseWrapper, err error) bool {
+				var apiErr *apierr.APIError
+				if errors.As(err, &apiErr) {
+					return apiErr.IsRetriable(ctx)
+				}
+				return false
+			},
+			httpclient.RetryUrlErrors,
```
mgyucht, line 87: I moved this out of the ApiClient to have a single codesite where the retry logic is defined for the client. The downside is that you need to add this explicitly in your ErrorRetriable if you don't specify DefaultErrorRetriable. Happy to make this a default behavior, let me know what you think.

config/api_client_test.go
```diff
@@ -0,0 +1,54 @@
...
+		}),
+	}
+	client, err := c.NewApiClient()
+	require.NoError(t, err)
+	ctx := context.Background()
+	var res map[string][]string
+	err = client.Do(ctx, "GET", "/api/2.0/permissions/object/id", httpclient.WithResponseUnmarshal(&res))
+	assert.NoError(t, err)
+	assert.Equal(t, map[string][]string{"permissions": {"can_run_queries"}}, res)
+}
```
renaudhartert-db, line 54: I'd recommend structuring the unit tests differently by having this test focused on how `ApiClient` manages mocked `ErrorRetrier` (or the absence of an `ErrorRetrier`). The tests of the `ErrorRetrier` themselves (e.g. verify that the path pattern match properly) should happen  in `httpclient/errors_test.go`. 

This test could look like something such as the following (I did not verify that the code works):

```go
type mock struct {
	MaxFails      int            // number of times the failed Response is returned
	FailResponse  *http.Response // response to return in case of fail
	FailError     error          // error to return in case of fail
	NumCalls      int            // total number of calls
}

func (m *mock) RoundTrip(r *http.Request) (*http.Response, error) {
	m.NumCalls++
	if m.NumCalls <= m.MaxFails {
		return m.FailResponse, n.FailError
	}
	return &http.Response{
		Request:    r,
		StatusCode: http.StatusOK,
		Body:       io.NopCloser(strings.NewReader(`{}`)),
	}, nil
}

func (m *mock) SkipRetryOnIO() bool {
	return true
}

func TestApiClient_Do_retries(t *testing.T) {
	testCases := []struct{
		desc         string
		config       *Config
		errorRetrier ErrorRetrier
		wantNumCalls int
	} {
		{
			desc: "nil retrier",
			mock: &mock{
				MaxFails: 1,
				FailResponse: &http.Response{StatusCode: http.StatusGatewayTimeout}
			}
			wantNumCalls: 1,
		},
		{
			desc: "no retry",
			mock: &mock{
				MaxFails: 1,
				FailResponse: &http.Response{StatusCode: http.StatusGatewayTimeout}
			}
			errorRetrier: func(context.Context, *http.Request, *common.ResponseWrapper, error) bool {
				return false
			},
			wantNumCalls: 1,
		},
		{
			desc: "retry 1 time",
			mock: &mock{
				MaxFails: 1,
				FailResponse: &http.Response{StatusCode: http.StatusGatewayTimeout}
			}
			errorRetrier: func(context.Context, *http.Request, *common.ResponseWrapper, error) bool {
				return true
			},
			wantNumCalls: 2,
		},
		{
			desc: "retry 2 times",
			mock: &mock{
				MaxFails: 2,
				FailResponse: &http.Response{StatusCode: http.StatusGatewayTimeout}
			}
			errorRetrier: func(_ context.Context, _ *http.Request, _ *common.ResponseWrapper, _ error) bool {
				return true
			},
			wantNumCalls: 3,
		},
		{
			desc: "retry 3 times",
			mock: &mock{
				MaxFails: 3,
				FailResponse: &http.Response{StatusCode: http.StatusGatewayTimeout}
			}
			errorRetrier: func(_ context.Context, _ *http.Request, _ *common.ResponseWrapper, _ error) bool {
				return true
			},
			wantNumCalls: 4,
		},
	} 


	func _, tc := range testCases {
		t.Run(tc.desc, func(t *testing.T) {
			cfg := &Config{HTTPTransport: tc.mock} 
			client, err := cfg.NewApiClient()
			client.ErrorRetrier = tc.errorRetrier

			err = client.Do(context.Background(), "GET", "test-path") 
			gotNumCalls = tc.mock.NumCalls

			if gotNumCalls != tc.wantNumCalls {
				t.Errorf("got %d calls, want %d", gotNumCalls, tc.wantNumCalls)
			}
		})
	}
}
```

Please feel free to ignore this comment if this is too much work or if the ApiClient cannot be instrumented that easily.

mgyucht: It definitely can be instrumented this way, and this is a nice test case to use (I'll adapt it and include it in this PR). However, I did want to specifically test the get permissions pathway. Essentially, this tests that "the client returned by Config.GetApiClient() correctly implements retry on 504." I will add more test cases here though.

client/client.go
```diff
@@ -63,8 +63,7 @@ func (c *DatabricksClient) GetOAuthToken(ctx context.Context, authDetails string
 
 // Do sends an HTTP request against path.
 func (c *DatabricksClient) Do(ctx context.Context, method, path string,
-	headers map[string]string, request, response any,
-	visitors ...func(*http.Request) error) error {
+	headers map[string]string, request, response any, visitors ...func(*http.Request) error) error {
```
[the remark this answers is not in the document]
mgyucht: Yeah. This `Do()` method should be formulated either as a struct argument or with functional options, since everything after `path` is optional. I'd prefer not to change this signature unless necessary, as it is used in multiple places in the TF provider.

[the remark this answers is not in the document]
renaudhartert-db: Agreed with not using a struct as this is out of scope for this PR. Let's at least format idiomatically then. Our Go style does not mandate a 80 char line length which makes the current formatting quite arbitrary. I'm fine with either having everything on a single line or one argument per line.

config/api_client_test.go
renaudhartert-db: > I did want to specifically test the get permissions pathway

Sounds good to me as long as this complements the overall testing of the retry logic.

httpclient/api_client.go
```diff
@@ -28,16 +27,25 @@ type ClientConfig struct {
 	AuthVisitor RequestVisitor
 	Visitors    []RequestVisitor
 
-	RetryTimeout       time.Duration
+	// The maximum amount of time to retry requests that return retriable errors.
+	// If unset, the default is 5 minutes.
+	RetryTimeout time.Duration
+
+	// Returns the amount of time to wait after the given attempt.
+	RetryBackoff retries.BackoffFunc
```
mgyucht, line 35: This is added here and exposed initially for testing, though it should be safe for users to configure this if they need to modify the backoff behavior.

mgyucht:
We will not go forward with this approach, rather will incorporate retriable errors into the REST API definition itself.

Gijsreyn:
Because of issue #1102, I was wondering how this incorperates into DABs @mgyucht ?

PR #1108, opened 2025-01-03
[Feature] Implement U2M Authentication in the Go SDK

mgyucht:

This PR moves logic about U2M OAuth login from the CLI to the Go SDK. This eliminates a cyclic dependency between the SDK and CLI for interacting with the OAuth token cache and enables U2M support directly in the Go SDK without need for the CLI to be installed.

Most of this code is carried over from the CLI, but I have made specific refactors to generalize it where needed.

Currently, the token cache key follows a specific structure:
- `https://<accounts-host>/oidc/accounts/<account-id>` for account-based sessions
- `https://<workspace host>` for workspace-based sessions
This can be generalized to allow callers to cache tokens in other manners. For example, users may want to cache tokens per principal or per set of OAuth scopes for their own OAuth applications.

Additionally, products should be able to use an isolated token cache or a token cache backed by a different storage medium, like a database. This is simple to do by introducing a token cache interface. Implementers can define their own credential strategy and reuse the PersistentAuth type to handle the negotiation.

## How is this tested?

Carried over tests from the CLI.

More tests are forthcoming, once a decision is made on this approach.


credentials/oauth/persistent_auth.go
```diff
@@ -0,0 +1,237 @@
...
+
+// PersistentAuth is an OAuth manager that handles the U2M OAuth flow. Tokens
+// are stored in and looked up from the provided cache. Tokens include the
+// refresh token. On load, if the access token is expired, it is refreshed
+// using the refresh token.
+type PersistentAuth struct {
+	// Cache is the token cache to store and lookup tokens.
+	cache cache.TokenCache
+	// Locker is the lock to synchronize token cache access.
+	locker sync.Locker
```
renaudhartert-db, line 45: It looks like `locker` is never locked. Am I missing something?

renaudhartert-db (changes requested):
Thanks for the PR — always good to get the SDK more self-contained.

credentials/cache/file.go
```diff
@@ -0,0 +1,108 @@
...
+	ownerReadWrite = 0o600
+
+	// format versioning leaves some room for format improvement
+	tokenCacheVersion = 1
+)
+
+var ErrNotConfigured = errors.New("databricks OAuth is not configured for this host")
+
+// this implementation requires the calling code to do a machine-wide lock,
+// otherwise the file might get corrupt.
```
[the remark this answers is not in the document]
mgyucht: Good question. It needs to be at least locked in PersistentAuth: the Load() method will read from the cache and then, if the access token is expired, refresh it and update the cache, which replaces it entirely. That requires some form of user-wide mutual exclusion that lasts longer than an individual method call, thus the lock file. I don't see any issues with adding an in-memory mutex within FileTokenCache as well in case an application is directly managing and updating tokens, considering what that would be protecting against.

credentials/cache/file.go
```diff
@@ -0,0 +1,108 @@
...
+type FileTokenCache struct {
```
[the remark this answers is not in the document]
mgyucht: Great, will use this description. Sorry, most of this is copied directly from the CLI without carefully reviewing for style or comments, but thanks for pointing these out, and I'm happy to correct them here.

credentials/oauth/persistent_auth.go
mgyucht: You're not missing anything, I forgot to add it before, but I have it locally and will include it in my next update.

httpclient/oauth_token.go
```diff
@@ -22,6 +21,19 @@ type GetOAuthTokenRequest struct {
 	Assertion string `url:"assertion"`
 }
 
+// OAuthToken represents an OAuth token as defined by the OAuth 2.0 Authorization Framework.
+// https://datatracker.ietf.org/doc/html/rfc6749
+type OAuthToken struct {
```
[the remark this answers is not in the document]
mgyucht: oauth2.Token has an "Expiry" field, which is computed based on the "expires_in" field. This mimics the unexported `tokenJSON` type from that library.

httpclient/oauth_token.go
```diff
@@ -22,6 +21,19 @@ type GetOAuthTokenRequest struct {
...
+	// The access token issued by the authorization server. This is the token that will be used to authenticate requests.
+	AccessToken string `json:"access_token"  auth:",sensitive"`
```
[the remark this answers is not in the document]
mgyucht: In some cases (like failed authentication), debug logs will include configuration parameters. Config attributes marked as sensitive are redacted in these logs. I don't think that also applies to this struct though. I'll take a closer look here.

httpclient/oidc.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+)
+
+var ErrOAuthNotSupported = errors.New("databricks OAuth is not supported for this host")
+
+func (c *ApiClient) GetOidcEndpoints(ctx context.Context, host, accountId string) (*OAuthAuthorizationServer, error) {
```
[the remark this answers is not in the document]
mgyucht: This method returns the OAuth endpoints for the given host/account ID pair. To authenticate, we need to know the token endpoint and (for U2M authentication) the authorization endpoint. It is called OIDC Endpoints here because we are using the endpoint discovery mechanism which is part of OpenID Connect (OIDC) and technically not in the OAuth specification.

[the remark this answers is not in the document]
mgyucht: This is a good point. We should probably define the `Do()` method as an interface, since this only depends on that method.

httpclient/oidc.go
```diff
@@ -0,0 +1,35 @@
...
+	prefix := host
+	if /* cfg.IsAccountClient() && */ accountId != "" {
```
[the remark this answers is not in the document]
mgyucht: This is commented because where it was moved from was in the `config` package, where it had access to this Config variable.

To be honest, this should probably be separated into two functions: one to get account-level OAuth endpoints and one to get workspace-level OAuth endpoints.

httpclient/oidc.go
```diff
@@ -0,0 +1,35 @@
...
+		// TODO: technically, we could use the same config profile for both workspace
+		// and account, but we have to add logic for determining accounts host from
+		// workspace host.
+		prefix := fmt.Sprintf("%s/oidc/accounts/%s", host, accountId)
+		return &OAuthAuthorizationServer{
+			AuthorizationEndpoint: fmt.Sprintf("%s/v1/authorize", prefix),
+			TokenEndpoint:         fmt.Sprintf("%s/v1/token", prefix),
+		}, nil
```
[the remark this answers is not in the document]
mgyucht: Good suggestion, will incorporate this!

config/auth_u2m.go
```diff
@@ -0,0 +1,198 @@
...
+	// token fails. If not specified, any error will cause Configure() to return
+	// said error.
+	ErrorHandler func(context.Context, *Config, oauth.OAuthArgument, error) error
+
+	name string
+}
+
+// Name implements CredentialsStrategy.
+func (u U2MCredentials) Name() string {
+	if u.name != "" {
```
mgyucht, line 49: This allows the unexported databricks-cli implementation to continue to use its existing name "databricks-cli", but if other users extend this strategy, the name will become `oauth-u2m`. We don't want users to choose a new name for this for consistency in our telemetry.

config/auth_u2m.go
```diff
@@ -0,0 +1,198 @@
...
+		return oauth.NewBasicAccountOAuthArgument(cfg.Host, cfg.AccountID)
+	}
+	return oauth.NewBasicWorkspaceOAuthArgument(cfg.Host)
+}
+
+var _ CredentialsStrategy = U2MCredentials{}
+
+// CliInvalidRefreshTokenError is a special error type that is returned when a
+// new access token could not be retrieved because the refresh token is invalid.
+// It is returned only by the `databricks-cli` credentials strategy.
```
mgyucht, line 117: Because SDK users may reuse a session where they authenticated with the CLI, there is a slight cyclic dependency here. It isn't a code dependency, rather a process one: users may need to use the CLI to reauthenticate if a refresh token is invalid. Thus, the SDK will now include a small bit of logic to generate the command to reauthenticate, which can be logged to instruct the user to reauthenticate.

config/auth_u2m.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+			// If there is an existing token but the refresh token is invalid,
+			// return a special error message for invalid refresh tokens. If the
+			// `databricks` CLI is on the PATH, include a command that the user can
+			// run to reauthenticate.
+			if _, ok := err.(*oauth.InvalidRefreshTokenError); ok {
+				var loginCommand string
+				if _, execErr := pathLooker.LookPath("databricks"); execErr == nil {
+					loginCommand = buildLoginCommand(ctx, cfg.Profile, arg)
+				}
```
mgyucht, line 185: Move this to the error message.

credentials/cache/file.go
```diff
@@ -0,0 +1,204 @@
...
+	locker sync.Locker
+}
+
+// NewFileTokenCache creates a new FileTokenCache. By default, the cache is
+// stored in "~/.databricks/token-cache.json". The cache file is created if it
+// does not already exist. The cache file is created with owner permissions
+// 0600 and the directory is created with owner permissions 0700. If the cache
+// file is corrupt or if its version does not match tokenCacheVersion, an error
+// is returned.
+func NewFileTokenCache(opts ...FileTokenCacheOpt) (*FileTokenCache, error) {
```
renaudhartert-db, line 81: I think we should return a `TokenCache` here rather than a `FileTokenCache`. The reasons are:

1. we do not want users to create a `FileTokenCache` without using this constructor,
2. the content of the FileTokenCache is private anyway, and
3. `FileTokenCache` does not expose new methods compared to the interface it implements.  

```suggestion
func NewFileTokenCache(opts ...FileTokenCacheOpt) (TokenCache, error) {
```

credentials/cache/file.go
```diff
@@ -0,0 +1,204 @@
...
+	c := &FileTokenCache{}
+	for _, opt := range opts {
+		opt(c)
+	}
+	if err := c.init(); err != nil {
+		return nil, err
+	}
+	// verify the cache is working
```
renaudhartert-db, line 89: [nit] Let's try to follow the same rules for all comments: https://go.dev/wiki/CodeReviewComments#comment-sentences

credentials/cache/file.go
```diff
@@ -0,0 +1,204 @@
...
+	if f.Version != tokenCacheVersion {
+		// in the later iterations we could do state upgraders,
+		// so that we transform token cache from v1 to v2 without
+		// losing the tokens and asking the user to re-authenticate.
+		return nil, fmt.Errorf("needs version %d, got version %d", tokenCacheVersion, f.Version)
+	}
+	return f, nil
+}
+
+var _ TokenCache = (*FileTokenCache)(nil)
```
renaudhartert-db, line 204: Not necessary if the constructor returns a `TokenCache`.

credentials/cache/file.go
```diff
@@ -0,0 +1,204 @@
...
+			Tokens:  map[string]*oauth2.Token{},
+		}
+		raw, err := json.MarshalIndent(f, "", "  ")
+		if err != nil {
+			return fmt.Errorf("marshal: %w", err)
+		}
+		if err := os.WriteFile(c.fileLocation, raw, ownerReadWrite); err != nil {
+			return fmt.Errorf("write: %w", err)
+		}
+	}
```
renaudhartert-db, line 167: I think we can simplify this slightly. 

```suggestion
	// create the file if it doesn't already exist
	if _, err := os.Stat(c.fileLocation); err != nil {
		if !os.IsNotExist(err) {
			return fmt.Errorf("stat file: %w", err)
		}
		
		// Create the parent directories if needed. 
		if err := os.MkdirAll(filepath.Dir(c.fileLocation), ownerExecReadWrite); err != nil {
			return fmt.Errorf("mkdir: %w", err)
		}
		
		// Create an empty file cache. 
		f := &tokenCacheFile{
			Version: tokenCacheVersion,
			Tokens:  map[string]*oauth2.Token{},
		}
		raw, err := json.MarshalIndent(f, "", "  ")
		if err != nil {
			return fmt.Errorf("marshal: %w", err)
		}
		if err := os.WriteFile(c.fileLocation, raw, ownerReadWrite); err != nil {
			return fmt.Errorf("write: %w", err)
		}
	}
```

credentials/cache/file.go
```diff
@@ -0,0 +1,204 @@
...
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return fmt.Errorf("home: %w", err)
+		}
+
+		c.locker, err = newLocker(filepath.Join(home, lockFilePath))
+		if err != nil {
+			return fmt.Errorf("locker: %w", err)
+		}
+	}
```
renaudhartert-db, line 179: See comment above, I would prefer to keep the locker internal.

credentials/cache/file.go
```diff
@@ -0,0 +1,204 @@
...
+// load loads the token cache file from disk. If the file is corrupt or if its
+// version does not match tokenCacheVersion, it returns an error.
+func (c *FileTokenCache) load() (*tokenCacheFile, error) {
+	raw, err := os.ReadFile(c.fileLocation)
+	if err != nil {
+		return nil, fmt.Errorf("read: %w", err)
+	}
+	f := &tokenCacheFile{}
+	err = json.Unmarshal(raw, &f)
+	if err != nil {
```
renaudhartert-db, line 192: [nit] This is more idiomatic and slightly more robust as we do not expect this error to "exist" outside the if statement. 

```suggestion
	if err := json.Unmarshal(raw, &f); err != nil {
```

config/auth_u2m.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Configure implements CredentialsStrategy.
+func (u u2mCredentials) Configure(ctx context.Context, cfg *Config) (credentials.CredentialsProvider, error) {
+	if cfg.Host == "" {
+		return nil, nil
+	}
+	a := u.auth
+	if a == nil {
+		var err error
+		a, err = oauth.NewPersistentAuth(ctx)
```
renaudhartert-db, line 62: Should we update `u.auth`?

renaudhartert-db (changes requested):
Added a few comments, specifically around the `cache` package. 

My remaining concerns are about the new `oauth` package and I think it would take time to review it properly, taking the whole auth library in consideration. 

If there's an easy way to provide a first implementation of `u2m` that does not rely on the `oauth` package, then I would recommend doing this first, and consider introducing the `oauth` package as a follow-up.

renaudhartert-db:
> Added a few comments, specifically around the `cache` package.
> 
> My remaining concerns are about the new `oauth` package and I think it would take time to review it properly, taking the whole "auth" library in consideration.
> 
> If there's an easy way to provide a first implementation of `u2m` that does not rely on the `oauth` package, then I would recommend doing this first, and consider introducing the `oauth` package as a follow-up.

Discussed offline:

- I misunderstood the scope of this package as being more general than `u2m`. 
- We agreed to make it clear that the package is specifically designed for `u2m`.

credentials/u2m/error.go
```diff
@@ -0,0 +1,18 @@
...
+
+func (e *InvalidRefreshTokenError) Error() string {
+	return e.err.Error()
+}
+
+func (e *InvalidRefreshTokenError) Unwrap() error {
+	return e.err
+}
+
+var _ error = &InvalidRefreshTokenError{}
```
renaudhartert-db, line 18: I think this could just be this:
```suggestion
// InvalidRefreshTokenError is returned from PersistentAuth's Load() method
// if the access token has expired and the refresh token in the token cache
// is invalid.
type InvalidRefreshTokenError struct {
	error
}
```

config/auth_u2m.go
```diff
@@ -0,0 +1,146 @@
...
+
+// u2mCredentials is a credentials strategy that uses the U2M OAuth flow to
+// authenticate with Databricks. It loads a token from the token cache for the
+// given workspace or account, refreshing it if needed. If the user has not
+// authenticated with OAuth U2M, it falls back to the next credentials strategy.
+// If they have but their access and refresh tokens are both invalid, it returns
+// a special error message that instructs the user how to reauthenticate.
+type u2mCredentials struct {
+	// auth is the persistent auth object that manages the token cache.
+	auth *u2m.PersistentAuth
```
renaudhartert-db, line 25: Should this just be a token source?

```suggestion
	ts oauth2.TokenSource
```

credentials/u2m/persistent_auth.go
```diff
@@ -0,0 +1,347 @@
...
+	return t, nil
+}
+
+// Challenge initiates the OAuth2 login flow for the given OAuthArgument. The
+// OAuth2 flow is started by opening the browser to the OAuth2 authorization
+// URL. The user is redirected to the callback server on appRedirectAddr. The
+// callback server listens for the redirect from the identity provider and
+// exchanges the authorization code for an access token. It returns the OAuth2
+// token on success.
+func (a *PersistentAuth) Challenge() (*oauth2.Token, error) {
```
renaudhartert-db, line 207: This method does not seem to be used. Also, its documentation does not look up to date. Is it necessary to implement some interface?

credentials/u2m/persistent_auth.go
```diff
@@ -0,0 +1,347 @@
...
+			return &l, nil
+		})
+	if err != nil {
+		return fmt.Errorf("listener: %w", err)
+	}
+	a.ln = *listener
+	return nil
+}
+
+func (a *PersistentAuth) Close() error {
```
renaudhartert-db, line 265: This method does not seem to be used publicly, can we make it private?

credentials/u2m/persistent_auth.go
```diff
@@ -0,0 +1,347 @@
...
+
+// WithBrowser sets the browser function for the PersistentAuth.
+func WithBrowser(b func(url string) error) PersistentAuthOption {
+	return func(a *PersistentAuth) {
+		a.browser = b
+	}
+}
+
+// NewPersistentAuth creates a new PersistentAuth with the provided options.
+func NewPersistentAuth(ctx context.Context, opts ...PersistentAuthOption) (*PersistentAuth, error) {
```
renaudhartert-db, line 89: Should this return a TokenSource so that the implementation is hidden? Ultimately, it does not look like the Challenge and Close methods are used outside of this package.

config/auth_u2m_test.go
```diff
@@ -0,0 +1,196 @@
...
+		auth       *u2m.PersistentAuth
+		expectErr  string
+		expectAuth string
+	}{
+		{
+			name: "happy path",
+			cfg: &Config{
+				Host: "https://myworkspace.cloud.databricks.com",
+			},
+			auth: must(
```
renaudhartert-db, line 58: I would have naturally leaned toward mocking `PersistentAuth` rather than the `OAuthClient` — especially if `PersitentAuth` is treated as a `TokenSource` (see other comments in `persistent_auth.go`).

credentials/u2m/client.go
```diff
@@ -0,0 +1,68 @@
...
+	"net/http"
+
+	"github.com/databricks/databricks-sdk-go/httpclient"
+)
+
+// OAuthClient provides the http functionality needed for interacting with the
+// Databricks OAuth APIs.
+type OAuthClient interface {
+	// GetHttpClient returns an HTTP client for OAuth2 requests.
+	GetHttpClient(context.Context) *http.Client
```
renaudhartert-db, line 16: This method bothers me. I'm wondering if we would not be better off passing the HTTPClient directly to `PersistentAuth`. I'm also wondering if this interface needs to be public (see comments about mocking `PersistentAuth` in `auth_u2m_test.go`).

credentials/u2m/persistent_auth.go
mgyucht: It will be used by the CLI and by any clients who want to use U2M but want to manage the token cache independently from the CLI.

credentials/u2m/client.go
mgyucht: We can split the OAuth endpoints and the HTTP client into separate parameters. However, because oauth2 depends on an http.Client, and to preserve the existing behavior, we do need to pass in the existing ApiClient somehow.

In an ideal world, we would have a pipelined API client specifically for OAuth that would be independent of the ApiClient package. We could actually just duplicate the code in order to break that dependency.

credentials/u2m/cache/file.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+// fileTokenCache caches tokens in "~/.databricks/token-cache.json". fileTokenCache
+// implements the TokenCache interface.
+type fileTokenCache struct {
+	fileLocation string
+
+	// locker protects the token cache file from concurrent reads and writes.
+	locker sync.Locker
```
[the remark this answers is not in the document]
mgyucht: Accepted!

credentials/u2m/error.go
mgyucht: Accepted!

credentials/u2m/cache/file.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	// Fail fast if the cache is not working.
+	if _, err := c.load(); err != nil {
+		return nil, fmt.Errorf("load: %w", err)
+	}
+	return c, nil
+}
+
+// Store implements the TokenCache interface.
+func (c *fileTokenCache) Store(key string, t *oauth2.Token) error {
```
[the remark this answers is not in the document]
mgyucht: Accepted.

config/auth_u2m.go
mgyucht: Absolutely right, in the context of the SDK, we never use the challenge component of PersistentAuth, so we only depend on the TokenSource interface.

renaudhartert-db (approved):
LGTM, thanks Miles!

PR #1110, opened 2025-01-03
[Internal] Decouple serving and oauth2 package

parthban-db:

Package `oauth2` contains a `DataPlaneService` that is not intended for users to use directly. This service is not auto-generated and is only used in the `serving` package. To remove the dependency between the `serving` package and the `oauth2` package, this PR moves the `DataPlaneService` from the `oauth2` to the `serving` package.

## How is this tested?
Existing unit tests


service/serving/ext_data_plane.go
```diff
@@ -1,35 +1,36 @@
...
 
-	"golang.org/x/oauth2"
+	"github.com/databricks/databricks-sdk-go/service/oauth2"
+	goauth "golang.org/x/oauth2"
 )
 
 // DataPlaneService is an interface for services that access DataPlane.
 type DataPlaneService interface {
-	GetDataPlaneDetails(method string, params []string, refresh func(*DataPlaneInfo) (*oauth2.Token, error), infoGetter func() (*DataPlaneInfo, error)) (string, *oauth2.Token, error)
+	GetDataPlaneDetails(method string, params []string, refresh func(*oauth2.DataPlaneInfo) (*goauth.Token, error), infoGetter func() (*oauth2.DataPlaneInfo, error)) (string, *goauth.Token, error)
```
renaudhartert-db, line 13: Message `DataPlaneInfo` is actually not at all in `oauth2`. I would simply make a copy of that message here.

renaudhartert-db: Discussed offline, this is OK.

renaudhartert-db (approved):

PR #1119, opened 2025-01-09
Renaud.hartert/sdkmod logger

renaudhartert-db:


databricks/log/log.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// InfoContext logs a message at the info level with context.
+func InfoContext(ctx context.Context, format string, v ...any) {
+	defaultLogger.Log(ctx, LevelInfo, format, v...)
+}
+
+// Warning logs a message at the warn level.
+func Warning(format string, v ...any) {
+	WarningContext(context.Background(), format, v...)
```
hectorcast-db, line 129: Isn't it common to mark methods which use formatting with an "f" suffix? (e.g., "Errorf")

databricks/log/log_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+// loggedParams returns the last set of parameters that were logged (if any)
+// by executing logFunc. It returns a zero params struct if logFunc did not
+// log anything.
+func loggedParams(logFunc func()) params {
+	old := defaultLogger
+	defer func() { defaultLogger = old }()
+	m := &mockLogger{}
+	defaultLogger = m
+	logFunc()
```
hectorcast-db, line 169: Nit: This also makes that the logged parameters are never logged by the real logger.

examples/http-proxy/client/main.go
```diff
@@ -3,18 +3,18 @@ package main
...
 	"github.com/databricks/databricks-sdk-go"
-	"github.com/databricks/databricks-sdk-go/logger"
+	"github.com/databricks/databricks-sdk-go/databricks/log"
 	"github.com/databricks/databricks-sdk-go/service/compute"
 )
 
 func main() {
 	if os.Getenv("HTTPS_PROXY") != "https://localhost:8443" {
-		log.Printf(`To run this example, first start the proxy server in the examples/http-proxy/proxy directory:
+		golog.Printf(`To run this example, first start the proxy server in the examples/http-proxy/proxy directory:
```
hectorcast-db, line 17: Why do we keep using the old logger?

httpclient/fixtures/named.go
```diff
@@ -0,0 +1,116 @@
+package fixtures
```
hectorcast-db, line 1: What is this new file?

openapi/code/load.go
```diff
@@ -1,129 +0,0 @@
-package code
```
hectorcast-db, line 1: Was the logger the only think keeping is from deleting those? Is that the reason we are deleting them now?

httpclient/fixtures/stub.go
```diff
@@ -57,7 +55,7 @@ func bodyStub(req *http.Request) (string, error) {
 	// which is not something i'm willing to write on my weekend
 	expectedRequest += "ExpectedRequest: XXX {\n"
 	for key, value := range receivedRequest {
-		camel := (&code.Named{Name: key}).PascalName()
```
hectorcast-db, line 60: Is this the only usage of Named? Should be reduce it's content to only what we need?

examples/http-proxy/client/main.go
renaudhartert-db: Good question. This example was already using the Go logger instead of the Databricks one before this change. I'd prefer to keep reconsidering the logic of the example outside of this PR's scope.

httpclient/fixtures/named.go
renaudhartert-db: It's a copy of the functionality that was used in the `openapi` library. The reason I'm removing the `openapi` library in this PR (rather than in a follow-up PR) is to reduce the replacement boilerplate as it uses the logger too.

httpclient/fixtures/stub.go
renaudhartert-db: TBH, I'm questioning whether the `fixtures` package itself should exists. I'd prefer to not spend too much effort cleaning it up since it might just disappear. Note that I've already trimmed the `named` file to what it needs to implement `PascalName`.

openapi/code/load.go
renaudhartert-db: It's kept in the original version for backward compatibility reasons. We do not need to honor these with SDK-Mod.

databricks/log/log.go
renaudhartert-db: You cannot imagine how long I spent hesitating about that... The current interface of the Logger is essentially a clone of the `slog` interface which does not use the `f` pattern.

hectorcast-db (approved):

PR #1120, opened 2025-01-13
[Internal] Scope the traversing directory in the Recursive list workspace test

parthban-db:
What - The PR modifies the scope of directory traversal in the recursive list test. It limits the search to the ".sdk" directory, where the relevant notebook is created.

Why - The test was previously hitting API rate limits because it was scanning too many elements in the current traversing directory. By narrowing the scope to the ".sdk" directory, the test can now run more efficiently without encountering these limits.

## How is this tested?
This is itself an integration test.


renaudhartert-db (approved):
Nice!

PR #1121, opened 2025-01-13
[Internal] Move all "core" packages in the "databricks" package. 

renaudhartert-db:

This PR moves all the "core" packages in the "databricks" directory which is meant to become the core library. 

This PR is also removing the mocked files while syncing with the latest version of main.

Note that this PR also changes generated code. This is intended as the code generator has to be updated to account for these changes.

## How is this tested?

Unit tests and integration tests are still passing.


CHANGELOG.md
```diff
@@ -1,5 +1,37 @@
 # Version changelog
 
+## [Release] Release v0.55.0
+
+### Internal Changes
```
hectorcast-db, line 5: [nit] I assume that this is from merging main. Can you add a note in the PR description?

hectorcast-db (approved):
[nit] Let's add in the PR description that we are also removing the mock files.

CHANGELOG.md
renaudhartert-db: Done

PR #1123, opened 2025-01-14
[Internal] Arrange service in a per package sdk approach

parthban-db:

Provide the readers and reviewers with the information they need to understand
this PR in a comprehensive manner. 

Specifically, try to answer the two following questions:

- **WHAT** changes are being made in the PR? This should be a summary of the 
  major changes to allow the reader to quickly understand the PR without having
  to look at the code. 
- **WHY** are these changes needed? This should provide the context that the 
  reader might be missing. For example, were there any decisions behind the 
  change that are not reflected in the code itself? 

The “why part” is the most important of the two as it usually cannot be 
inferred from the code itself. A well-written PR description will help future
developers (including your future self) to know how to interact and update your
code.

## How is this tested?

Describe any tests you have done; especially if test tests are not part of
the unit tests (e.g. local tests).

**ALWAYS ANSWER THIS QUESTION:** Answer with "N/A" if tests are not applicable
to your PR (e.g. if the PR only modifies comments). Do not be afraid of 
answering "Not tested" if the PR has not been tested. Being clear about what 
has been done and not done provides important context to the reviewers.


workspace/v2/ext_utilities.go
```diff
@@ -60,12 +59,44 @@ type workspaceAPIUtilities interface {
...
+	content = TrimLeadingWhitespace(content)
 	request, _ := PythonNotebookOverwriteReader(path,
 		strings.NewReader(content))
 	return request
 }
 
+// TrimLeadingWhitespace removes leading whitespace, so that Python code blocks
+// that are embedded into Go code still could be interpreted properly.
+// TODO: for note this is from the compute package
+func TrimLeadingWhitespace(commandStr string) (newCommand string) {
```
parthban-db, line 71: Copied from the compute package to break dependency

serving/v2/impl.go
```diff
@@ -181,48 +188,15 @@ func (a *servingEndpointsImpl) UpdatePermissions(ctx context.Context, request Se
 
 // unexported type that holds implementations of just ServingEndpointsDataPlane API methods
 type servingEndpointsDataPlaneImpl struct {
```
parthban-db, line 190: Need to fix this.

service/vectorsearch/interface.go
```diff
@@ -1,94 +0,0 @@
-// Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
```
parthban-db, line 1: We don't need an interface, right?

service/files/doc.go
```diff
@@ -1,49 +0,0 @@
-// Databricks File System (DBFS) API
```
parthban-db, line 1: Do we need this?

service/compute/ext_utilities.go
```diff
@@ -1,164 +0,0 @@
-package compute
```
parthban-db, line 1: Need to implement wait for state first.

iam/v2/client.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type PermissionsClient struct {
+	cfg       *config.Config
+	apiClient *httpclient.ApiClient
+
+	permissions PermissionsInterface
+}
+
+func NewPermissionsClientFromConfig(c ...*config.Config) (*PermissionsClient, error) {
```
parthban-db, line 18: Just a copy of how we make workspace client.

serving/v2/impl.go
parthban-db: This will come in future PRs.

renaudhartert-db (approved):

PR #1124, opened 2025-01-17
[Fix] Support Query parameters for all HTTP operations

hectorcast-db:
Support Query parameters for all HTTP operations. Previously, only GET/HEAD/DELETE were supported.

## How is this tested?
* Added Unit Tests
* Run Integration Tests


client/client_test.go
```diff
@@ -68,9 +70,11 @@ func TestSimpleRequestFailsAPIError(t *testing.T) {
 	require.NoError(t, err)
 	err = c.Do(context.Background(), "GET", "/a/b", map[string]string{
 		"e": "f",
-	}, map[string]string{
-		"c": "d",
-	}, nil)
+	},
+		nil,
+		map[string]string{
+			"c": "d",
```
renaudhartert-db, line 76: Can we also cover cases where the value is not a string?

httpclient/request_test.go
```diff
@@ -141,3 +141,26 @@ func TestEncodeMultiSegmentPathParameter(t *testing.T) {
 	// # and ? should be encoded.
 	assert.Equal(t, "a%23b%3Fc", EncodeMultiSegmentPathParameter("a#b?c"))
 }
+
+func TestMakeRequestBodyExplicitQueryParams(t *testing.T) {
+	type x struct {
```
mgyucht, line 146: To be clear, the first scenario should not happen, IIUC. For GET/HEAD/DELETE requests, extra query parameters are never passed in.

Could the second scenario happen, where the request contains `url` tags that are overridden by the explicit query parameters?

client/client.go
```diff
@@ -63,12 +63,13 @@ func (c *DatabricksClient) GetOAuthToken(ctx context.Context, authDetails string
 
 // Do sends an HTTP request against path.
 func (c *DatabricksClient) Do(ctx context.Context, method, path string,
-	headers map[string]string, request, response any,
+	headers map[string]string, queryParams map[string]any, request, response any,
```
mgyucht, line 66: For external clients (like Terraform), the migration path is to pass nil for GET/HEAD/DELETE requests and to explicitly pass the query parameters here?

mgyucht (approved):

client/client.go
hectorcast-db: That's the normal path. 
However, since you have integration tests for each resource in terraform, it would also be possible to pass the query parameters always and validate that the resource has not changed.

httpclient/request_test.go
hectorcast-db: > To be clear, the first scenario should not happen, IIUC. For GET/HEAD/DELETE requests, extra query parameters are never passed in.
Correct. 

> Could the second scenario happen, where the request contains url tags that are overridden by the explicit query parameters?
The logic to add the params is the same as before (just the source has changed). I am not aware of having a path with url tags.

mgyucht: As I think about it, I think this can only happen for APIs using `google.api.http.body`. Protobuf-based APIs that don't use this send all parameters in either the path or body. We could see if there are any OpenAPIs that set `x-databricks-body-field` and have parameters with `"in": "query"`, since they might have the same behavior.

PR #1126, opened 2025-01-17
[Internal] Delete examples/mocking module

parthban-db:
This PR deletes the `examples/mocking` directory as it is no longer maintained. This mocker example is built on the assumption that every `service` has `WithImpl()` that is used to set the stubs for every `service`. But this is no longer the case. This mocker is not even able to compile.

## How is this tested?
N/A


examples/mocking/dbfs_test.go
```diff
@@ -1,48 +0,0 @@
...
-		Data:   "YWJj",
-	}))
-	mockDbfs.EXPECT().Close(gomock.Any(), gomock.Eq(dbfs.Close{
-		Handle: 123,
-	}))
-
-	w, err := databricks.NewWorkspaceClient()
-	require.NoError(t, err)
-
-	w.Dbfs.WithImpl(mockDbfs)
```
parthban-db, line 44: @renaudhartert-db What do you say? Should we delete this module or find a way to support it? Random fact: this mocker is too old that it even uses the `Dbfs` package, which does not even exist anymore :)

renaudhartert-db: Let's remove it. Not only it is not relevant, it also promotes a practice that we do not recommend: client mocking should be done on the using interface defined by the SDK users.

renaudhartert-db (approved):

PR #1132, opened 2025-01-21
[Internal] GetRun logic paginates more arrays

gkiko10:

The existing code only paginates tasks and iterations. With this PR we update the logic to also paginate job_clusters, job_parameters and repair_history. This changes are needed for [Jobs API 2.2](https://docs.databricks.com/api/workspace/jobs/getrun) compatibility.

## How is this tested?

I enabled API 2.2 calls by modifying URL string `/api/2.2/jobs/runs/get` in service/jobs/impl.go. Then I ran unit test from service/jobs/ext_api_test.go

### DIFF
--- a/service/jobs/ext_api.go
+++ b/service/jobs/ext_api.go
@@ -14,6 +14,7 @@ func (a *JobsAPI) GetRun(ctx context.Context, request GetRunRequest) (*Run, erro
 	// When querying a ForEach task run, a page token is returned when there are more than 100 iterations. Only a single task is returned, corresponding to the ForEach task itself. Therefore, the client only reads the iterations from the next page and not the tasks.
 	isPaginatingIterations := len(run.Iterations) > 0
 
+	// runs/get response includes next_page_token as long as there are more pages to fetch.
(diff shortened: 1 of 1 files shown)


service/jobs/ext_api.go
```diff
@@ -27,7 +26,10 @@ func (a *JobsAPI) GetRun(ctx context.Context, request GetRunRequest) (*Run, erro
 		} else {
 			run.Tasks = append(run.Tasks, nextRun.Tasks...)
 		}
-		pageToken = nextRun.NextPageToken
+		run.JobClusters = append(run.JobClusters, nextRun.JobClusters...)
+		run.JobParameters = append(run.JobParameters, nextRun.JobParameters...)
+		run.RepairHistory = append(run.RepairHistory, nextRun.RepairHistory...)
+		run.NextPageToken = nextRun.NextPageToken
```
renaudhartert-db, line 32: Could you add a comment to explain why we are doing this?

hectorcast-db (approved):

PR #1133, opened 2025-01-21
[Internal] Update Jobs GetJob API to support paginated responses

gkiko10:

Introduces logic in extension for jobs GetJob call that paginates tasks and other arrays in the response. This change is necessary for SDK and API 2.2 compatibility. API 2.2 serves paginated responses as long as next_page_token field is present in the response. The pagination logic is not exposed to the customer.

## How is this tested?

I enabled API 2.2 calls by modifying URL string /api/2.2/jobs/get in service/jobs/impl.go. Then I ran unit test from service/jobs/ext_api_test.go

### DIFF
--- a/service/jobs/ext_api.go
+++ b/service/jobs/ext_api.go
@@ -32,3 +32,27 @@ func (a *JobsAPI) GetRun(ctx context.Context, request GetRunRequest) (*Run, erro
 
 	return run, nil
 }
+
+func (a *JobsAPI) Get(ctx context.Context, request GetJobRequest) (*Job, error) {
+	job, err := a.jobsImpl.Get(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	pageToken := job.NextPageToken
+	for pageToken != "" {
+		request.PageToken = pageToken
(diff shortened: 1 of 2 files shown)


service/jobs/ext_api.go
```diff
@@ -32,3 +32,28 @@ func (a *JobsAPI) GetRun(ctx context.Context, request GetRunRequest) (*Run, erro
...
+		if err != nil {
+			return nil, err
+		}
+
+		job.Settings.Tasks = append(job.Settings.Tasks, nextJob.Settings.Tasks...)
+		job.Settings.JobClusters = append(job.Settings.JobClusters, nextJob.Settings.JobClusters...)
+		job.Settings.Parameters = append(job.Settings.Parameters, nextJob.Settings.Parameters...)
+		job.Settings.Environments = append(job.Settings.Environments, nextJob.Settings.Environments...)
+		job.NextPageToken = nextJob.NextPageToken
+	}
```
renaudhartert-db, line 56: Could you add a comment to explain to a reader with limited context why we are doing this?

hectorcast-db (approved):

PR #1135, opened 2025-01-22
[Feature] Add support for async OAuth token refreshes

renaudhartert-db:

This PR aims at eliminating long-tail latency due to OAuth token refreshes in scenarios where a single client is responsible for a relatively high (e.g. > 1 QPS) continuous outbound traffic. The feature is disabled by default — which arguably makes this PR a functional no-op.

Precisely, the PR introduces a new token cache which attempts to always keep its token fresh by asynchronously refreshing the token before it expires. We differentiate three token states: 

- `fresh`: The token is valid and is not close to its expiration.
- `stale`: The token is valid but will expire soon.
- `expired`: The token has expired and cannot be used.

Each time a request tries to access the token, we do the following:

- If the token is `fresh`, return the current token;
- If the token is `stale`, trigger an asynchronous refresh and return the current token;
- If the token is `expired`, make a blocking refresh call to update the token and return it. 

In particular, asynchronous refreshes use a lock to guarantee that there can only be one pending refresh at a given time. 

The performance of the algorithm depends on the length of the `stale` and `fresh` periods. On the first hand, the `stale` period must be long enough to prevent tokens from entering the expired state. On the other hand, a long `stale` period reduces the length of the `fresh` period, thus increasing the refresh frequency. 

Right now, the `stale` period is configured to 3 minutes by default (i.e. 5% of the expected token lifespan of 1 hour). This value might be changed in the future to guarantee that the default behavior achieves the best performance for the majority of users. 

**For reviewers:** 

- This PR only uses the new cache in control-plane auth flows; I plan to send a follow-up PR to enable asynchronous refresh in data-plane flows once this one has been merged.
- Interface `oauth2.TokenSource` is likely not sufficient for us and we would need one with a `Token` method that takes a `context.Context` as parameter: `Token(context.Context) (Token, error)`.

## How is this tested?

Complete test coverage with a focus on various concurrency scenarios.


config/experimental/auth/auth.go
```diff
@@ -0,0 +1,202 @@
...
+// function), it is returned as is.
+func NewCachedTokenSource(ts oauth2.TokenSource, opts ...Option) oauth2.TokenSource {
+	if cts, ok := ts.(*cachedTokenSource); ok {
+		return cts
+	}
+
+	cts := &cachedTokenSource{
+		tokenSource:   ts,
+		staleDuration: defaultStaleDuration,
+		disableAsync:  defaultDisableAsyncRefresh,
```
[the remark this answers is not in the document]
renaudhartert-db: Agreed in general but I think this should be a little more nuanced. 

Ultimately, the intended default behavior is to enable async. Using `disable` as the field guarantees that the zero value is the default, and that any modification of the field is part of the "non-default" flow — which would typically happen at higher level of indentation. 

Note that the public option is not a negation: `WithAsyncRefresh`.

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,202 @@
...
+func (cts *cachedTokenSource) asyncToken() (*oauth2.Token, error) {
+	cts.mu.Lock()
+	ts := cts.tokenState()
+	cts.mu.Unlock()
+
+	switch ts {
+	case fresh:
+		cts.mu.Lock()
+		defer cts.mu.Unlock()
+		return cts.cachedToken, nil
```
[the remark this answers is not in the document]
renaudhartert-db: I agree that making the operation atomic would speed up the code. Though, I'm not sure I see what the race condition would be. It's true that we might return a token that is older than what the state suggest but that can happen even if the operation is atomic, it is just a little more likely right now.

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,202 @@
...
+	for _, opt := range opts {
+		opt(cts)
+	}
+
+	return cts
+}
+
+type cachedTokenSource struct {
+	tokenSource   oauth2.TokenSource
+	staleDuration time.Duration
```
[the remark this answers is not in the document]
renaudhartert-db: Done 👍

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,202 @@
...
+		cts.mu.Lock()
+		defer cts.mu.Unlock()
+		return cts.cachedToken, nil
+	case stale:
+		cts.triggerAsyncRefresh()
+		cts.mu.Lock()
+		defer cts.mu.Unlock()
+		return cts.cachedToken, nil
+	default: // expired
+		return cts.blockingToken()
```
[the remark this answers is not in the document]
renaudhartert-db: > By locking inside this function and not deduping the calls to Token() become sequential if there is a failure.

I might be misunderstanding your comment but shouldn't the calls be sequential even if we were to cache the error? The `blockingToken` function already contains a path (lines 186-188) to avoid calling the TokenSource if a call (either blocking or async) successfully refreshed the token.

> I think it is fair to try and acquire the token once and also cache the failure.

Considering blockingCall errors as retriable is intended. Though, I'm happy to reconsider this design decision if we believe an error should "close" the TokenSource forever.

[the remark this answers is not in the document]
pietern: That makes sense. Could you clarify this on L186.

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+	cts.mu.Lock()
+
+	// The lock is kept for the entire operation to ensure that only one
+	// blockingToken operation is running at a time.
+	defer cts.mu.Unlock()
+
+	// This is important to recover from potential previous failed attempts
+	// to refresh the token asynchronously, see declaration of refreshErr for
+	// more information.
+	cts.isRefreshing = false
```
pietern, line 184: This must also reset `cts.refreshErr`.

pietern: The reset can happen after updating the new token, to indicate we're ready to do async refreshes again.

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+func (cts *cachedTokenSource) triggerAsyncRefresh() {
+	cts.mu.Lock()
+	defer cts.mu.Unlock()
+	if !cts.isRefreshing && cts.refreshErr == nil {
+		cts.isRefreshing = true
+		go cts.asyncRefresh()
+	}
+}
+
+func (cts *cachedTokenSource) asyncRefresh() {
```
pietern, line 207: This function can only be called from the goroutine spawned in the function above.

Could you add a comment to capture this requirement?

config/experimental/auth/auth.go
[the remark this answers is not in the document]
renaudhartert-db: Yes that makes total sense, I was worried that I missed something more problematic

config/experimental/auth/auth.go
renaudhartert-db: I've inlined it so that there's absolutely no room for ambiguity.

config/experimental/auth/auth.go
renaudhartert-db: Good catch, fixed.

pietern (approved):
The logic is clear now (and simpler with one less function and less lock acquisition). @denik Could you take an async (LOL) look at the PR as well?

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+func WithCachedToken(t *oauth2.Token) Option {
+	return func(cts *cachedTokenSource) {
+		cts.cachedToken = t
+	}
+}
+
+// WithStaleDuration sets the duration for which a token is considered stale.
+// Stale tokens are still valid but will trigger an asynchronous refresh if
+// async refresh is enabled. The default value is 3 minutes.
+func WithStaleDuration(d time.Duration) Option {
```
denik, line 36: I don't see this option being used outside of tests. Is there a follow up PR where it is in use?

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+	return func(cts *cachedTokenSource) {
+		cts.staleDuration = d
+	}
+}
+
+// WithAsyncRefresh enables or disables the asynchronous token refresh.
+func WithAsyncRefresh(b bool) Option {
+	return func(cts *cachedTokenSource) {
+		cts.disableAsync = !b
```
denik, line 45: It would be great if option name matched the internal field exactly:

- staleDuration -> WithStaleDuration
- disableAsync -> WithDisableAsync (I also agree that enableAsync would be more natural).

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+// The token cache is safe for concurrent use by multiple goroutines and will
+// guarantee that only one token refresh is triggered at a time.
+//
+// The token cache does not take care of retries in case the token source
+// returns and error; it is the responsibility of the provided token source to
+// handle retries appropriately.
+//
+// If the TokenSource is already a cached token source (obtained by calling this
+// function), it is returned as is.
+func NewCachedTokenSource(ts oauth2.TokenSource, opts ...Option) oauth2.TokenSource {
```
denik, line 62: Why return an interface rather than explicit class? This is all experimental so we don't need to worry about details changing. [Preemptive interfacing](https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b) is often considered anti-pattern. Maybe you have a good reason - but it's no clear.

If you agree to this, then follow up question is why not export the class cachedTokenSource -> CachedTokenSource?

Once you exported the class, you can get rid of options and let users set StaleDuration and DisableAsync directly on the class. 

In other words, it would be simpler if this followed Cmd/Command pattern:

https://pkg.go.dev/os/exec#Cmd
https://pkg.go.dev/os/exec#Command

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+	if cts, ok := ts.(*cachedTokenSource); ok {
```
denik, line 63: It's an interesting pattern, can you add some comments why is it needed? Is it programming error where this function is called twice unexpectedly?

config/experimental/auth/auth.go
```diff
@@ -0,0 +1,218 @@
...
+	default:
+		return fresh
+	}
+}
+
+func (cts *cachedTokenSource) asyncToken() (*oauth2.Token, error) {
+	cts.mu.Lock()
+	ts := cts.tokenState()
+	t := cts.cachedToken
+	cts.mu.Unlock()
```
denik, line 158: nit: maybe a helper function that returns both tokenState() and cachedToken and takes care of locking internally would be useful there, because it would allow you to switch to 'defer cts.mu.Unlock()' so that you know that mutex is unlocked on panics as well.

config/experimental/auth/auth.go
renaudhartert-db: I'm perfectly fine removing this until the need is actually introduced.

config/experimental/auth/auth.go
denik: If that's the pattern you're using everywhere in SDK then it makes sense to follow it here as well.

denik (approved):

config/experimental/auth/auth.go
renaudhartert-db: I agree this would make sense if we were in a "all other things being equal" situation; though, I don't think that is not the case here. As mentioned in another comment, I believe there is value in keeping the internal field to the negative form so that the zero and (intended) default values are the same. Though, I don't believe that the negative form is the right user interface — hence the negation. 

I'm open to alternative names for the option but I'm not sure `EnableAsync` is a better name as `WithX` is the dominant pattern we've been following. All that being said, I'd be happy to consider dropping the `WithX` pattern in SDK-Mod.

PR #1142, opened 2025-02-06
[Internal] Add support for asynchronous data plane token refreshes

renaudhartert-db:
DO NOT MERGE: `api.go` and `impl.go` are here to help with review and will be removed before merge. These are supposed to be added later on via code generation.

This PR is a step towards enabling asynchronous refreshes of data plane tokens.

The PR introduces a new interface `dataplane.EndpointTokenSource` which returns data plane tokens for a given endpoint and authorization detail. The default implementation relies on `auth.NewCachedTokenSource` to wrap the underlying TokenSources. 

Note: async refreshes are disabled at the moment and will be enabled in a follow-up PR. 

**About Config.GetTokenSource():**

My first attempt to implement `GetTokenSource()` returned an error if the `TokenSource` could not be built (e.g. because of config initialization errors). This didn't work because:

1. service constructors do not return errors and thus had to panic;
2. service constructors are called systematically as soon as the `WorkspaceClient` is created — sometimes with a config that is not compatible with `GetTokenSource()`.

Returning a `TokenSource` provider instead of a `TokenSource` would have solved the problem but did not feel right. The current solution (i.e. return an ever-failing token source) is the cleanest solution I could think off.

## How is this tested?

Complete unit test coverage of the new package.

### DIFF
--- a/service/serving/impl.go
+++ b/service/serving/impl.go
@@ -4,13 +4,13 @@ package serving
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/databricks/databricks-sdk-go/client"
+	"github.com/databricks/databricks-sdk-go/config/experimental/auth/dataplane"
 	"github.com/databricks/databricks-sdk-go/httpclient"
-	goauth "golang.org/x/oauth2"
 )
 
 // unexported type that holds implementations of just ServingEndpoints API methods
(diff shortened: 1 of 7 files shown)


service/serving/impl.go
```diff
@@ -209,39 +209,40 @@ func (a *servingEndpointsImpl) UpdatePermissions(ctx context.Context, request Se
...
+		info = response.DataPlaneInfo.QueryInfo
+		a.infos.Store(key, info)
 	}
-	getParams := []string{
-		request.Name,
+	return info.(*DataPlaneInfo), nil
+}
+
+func (a *servingEndpointsDataPlaneImpl) Query(ctx context.Context, request QueryEndpointInput) (*QueryEndpointResponse, error) {
+	key := "Query" + request.Name
```
hectorcast-db, line 235: More of a note for the template:
Should we add a separator between the operation and the parameter, to avoid "clashes" in case two endpoints have the same prefix? (e.g., "Query" + "PreviewModel" vs "QueryPreview" + "Model")

renaudhartert-db: Agreed 👍

PR #1143, opened 2025-02-06
[Feature] Enable async refreshes for OAuth tokens

renaudhartert-db:

This PR enables async refreshes for cached token sources. 

This PR also: 

- Fixes a race condition bug in the unit tests by adding a small delay to the test refresh.
- Remove one test in AzureCLI that is not correct anymore due to how the code is structured. Refactoring this to guarantee that the test pass is not trivial and ultimately out of the scope of the new system — as the cached token source is the one providing the guarantee. 

## How is this tested?

Fix unit tests.

### DIFF
--- a/config/experimental/auth/auth_test.go
+++ b/config/experimental/auth/auth_test.go
@@ -39,8 +39,8 @@ func TestNewCachedTokenSource_default(t *testing.T) {
 	if got.staleDuration != defaultStaleDuration {
 		t.Errorf("NewCachedTokenSource() staleDuration = %v, want %v", got.staleDuration, defaultStaleDuration)
 	}
-	if got.disableAsync != defaultDisableAsyncRefresh {
-		t.Errorf("NewCachedTokenSource() disableAsync = %v, want %v", got.disableAsync, defaultDisableAsyncRefresh)
+	if got.disableAsync != false {
+		t.Errorf("NewCachedTokenSource() disableAsync = %v, want %v", got.disableAsync, false)
 	}
 	if got.cachedToken != nil {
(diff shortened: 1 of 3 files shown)


config/experimental/auth/auth_test.go
```diff
@@ -227,7 +227,7 @@ func TestCachedTokenSource_Token(t *testing.T) {
 			desc:          "[Async] stale cached token, expired token returned",
 			cachedToken:   &oauth2.Token{Expiry: now.Add(1 * time.Minute)},
 			returnedToken: &oauth2.Token{Expiry: now.Add(-1 * time.Second)},
-			wantCalls:     10,
+			wantCalls:     1,
```
parthban-db, line 230: This is a little weird as if the token is not valid, then technically, all threads should try to get a new token.

renaudhartert-db: > all threads should try to get a new token.

Avoiding this is one of the main goals of the new caching strategy: there should always be at most one refresh at any time, no matter the number of thread.

renaudhartert-db: Reference: https://github.com/databricks/databricks-sdk-go/pull/1135

parthban-db: > Avoiding this is one of the main goals of the new caching strategy: there should always be at most one refresh at any time, no matter the number of thread.

Agreed, but shouldn't all threads test the validity of the token, and if it is expired or stale, they should try to get a new one? In this case, the token is not valid anytime during the run, so no one gets a valid token during the run.

parthban-db (approved):

PR #1151, opened 2025-02-17
[Internal] Update Jobs `list_runs` function to support paginated responses

gkiko10:

Introduces logic in the extension for jobs ListRuns call. The extended logic accounts for the new response format of API 2.2. API 2.1 format returns all tasks and job_cluster for each run in the runs list. API 2.2 format truncates tasks and job_cluster to 100 elements. The extended ListRuns logic calls GetRun for each run in the list to populate the full list of tasks and job_clusters.

I added logic that reads runs from the list and produces custom iterator struct `expandedRunsIterator` that is supposed to mimic python generators. The goal is to only read necessary elements from the API endpoint and not more.

## How is this tested?

Unit tests and manual tests. Manual tests were done in two modes: using API 2.2 and using API 2.1. So this code is compatible with both API versions.

### DIFF
--- a/service/jobs/ext_api.go
+++ b/service/jobs/ext_api.go
@@ -6,6 +6,9 @@ import (
 	"github.com/databricks/databricks-sdk-go/listing"
 )
 
+// ListRuns fetches a list of job runs.
+// If expand_tasks is true, the response will include the full list of tasks and job_clusters for each run.
+// This function handles pagination two ways: paginates all the runs in the list and paginates all the tasks and job_clusters for each run.
 func (a *JobsAPI) ListRuns(ctx context.Context, request ListRunsRequest) listing.Iterator[BaseRun] {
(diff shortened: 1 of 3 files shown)


service/jobs/ext_api.go
```diff
@@ -1,6 +1,58 @@
 package jobs
 
-import "context"
+import (
+	"context"
+
+	"github.com/databricks/databricks-sdk-go/listing"
+)
+
+func (a *JobsAPI) ListRuns(ctx context.Context, request ListRunsRequest) listing.Iterator[BaseRun] {
```
renaudhartert-db, line 9: Could we document this function? I don't think the documentation from the underlying function will be ported.

renaudhartert-db (approved):
LGTM modulo open comment.

renaudhartert-db (approved):

PR #1161, opened 2025-03-03
[Internal] Moved heading descriptions in comments for PR Template

tanmay-db:

The description for headings is meant to be read by the PR author to add proper information and hence should be moved under comments. This saves time by not having to manually remove these heading descriptions as they add noise to the actual PR changes if left as is. 

If this is approved, then will do this for other SDKs as well. 

## How is this tested?

N/A

NO_CHANGELOG=true

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,5 @@
 ## What changes are proposed in this pull request?
-<!--
-Provide the readers and reviewers with the information they need to understand
+<!-- Provide the readers and reviewers with the information they need to understand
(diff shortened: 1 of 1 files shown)


renaudhartert-db:
> The description for headings is meant to be read by the PR author to add proper information and hence should be moved under comments. This saves time by not having to manually remove these heading descriptions as they add noise to the actual PR changes if left as is.

This is working as intended:

- The removal friction serves as a "check-box" to confirm that the PR author knows that the instructions exist.
- The whole PR description is part of the Git history and we want to avoid adding noise there. Having the instructions as comments increases that risk.

I shared the original document with you internally.

PR #1171, opened 2025-03-14
[Internal] Remove Config from the API Clients

renaudhartert-db:

This PR helps disentangling the SDK by removing the `config.Config` object from the API clients. 

## How is this tested?

Normal CI. 

NO_CHANGELOG=true


databricks/client/client.go
```diff
@@ -31,36 +31,27 @@ func NewWithClient(cfg *config.Config, client *httpclient.ApiClient) (*Databrick
...
 	return &DatabricksClient{
-		Config: cfg,
-		client: client,
+		accountID: cfg.AccountID,
+		client:    client,
 	}, nil
 }
 
 type DatabricksClient struct {
-	Config *config.Config
```
parthban-db, line 40: This is used in two places in SDK Beta apart from the accountID usage. [this](https://github.com/databricks/databricks-sdk-go/blob/main/service/compute/ext_utilities.go#L155) and [this](https://github.com/databricks/databricks-sdk-go/blob/main/service/serving/api.go#L447).

parthban-db: The DataPlane one seems to be legit.

renaudhartert-db: This is SDK-Mod, it is not used.

parthban-db (approved):

PR #1187, opened 2025-03-20
[Internal] Remove service client interfaces

renaudhartert-db:

This PR heavily refactors the service clients.

The new structure for each service `Foo` is a stack of three clients:

- `fooImpl` is the bottom, fully generated internal client. It serves as direct thin layer to interact with the REST API with no additional logic. 
- `fooBaseImpl` extends `fooImpl` by providing higher-level functionalities such as waiters. It is internal and fully generated.
- `FooClient` is the actual user facing client which embeds `fooBaseImpl`. By default, it does not add any additional function to what is already provided by `fooBaseImpl`. However, it can easily be extended with handcrated mixins.

All mixins have been updated to use `FooClient` instead of `FooAPI` (which has been renamed to `fooBaseImpl`). 

## How is this tested?

Normal CI.

NO_CHANGELOG=true


compute/v2/ext_commands.go
```diff
@@ -8,15 +8,11 @@ import (
 )
 
 type CommandExecutorV2 struct {
-	executionAPI *CommandExecutionAPI
+	executionAPI *CommandExecutionClient
```
parthban-db, line 11: [nit]:
```suggestion
	executionClient *CommandExecutionClient
```

files/v2/ext_utilities.go
```diff
@@ -60,7 +60,7 @@ func (f *fileHandleWriter) error(err error) error {
 // Internal only state for a DBFS file handle.
 type fileHandle struct {
 	ctx  context.Context
-	api  *DbfsAPI
+	api  *DbfsClient
```
parthban-db, line 63: [nit]:
```suggestion
	client  *DbfsClient
```

parthban-db (approved):

PR #1197, opened 2025-03-31
API Client timeouts should be int64 (breaks linux_arm target)

rayalex:

Implementation uses int(...) for default handling which overflows on linux_arm architecture:

```
../../../../go/pkg/mod/github.com/databricks/databricks-sdk-go@v0.60.0/httpclient/api_client.go:68:70: constant 30000000000 overflows int
```

I've changed it to int64 which also matches `time.Duration` size.

## How is this tested?

PR test suite.

### DIFF
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -73,10 +73,10 @@ func (cfg ClientConfig) httpTransport() http.RoundTripper {
 }
 
 func NewApiClient(cfg ClientConfig) *ApiClient {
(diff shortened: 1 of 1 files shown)


httpclient/api_client.go
```diff
@@ -392,7 +392,11 @@ func (c *ApiClient) perform(
 	return resp, nil
 }
 
-func orDefault(configured, _default int) int {
+type number interface {
+	~int | ~int64 | ~float64 | ~float32
+}
+
+func orDefault[T number](configured, _default T) T {
```
rayalex, line 399: There are some other users of this function in the package. I opted for generic version so not to change other consumers.

PR #1203, opened 2025-04-07
Fix example test names to comply with Go 1.24 vet checks

rauchy:

This PR updates the names of many `Example...` functions in `*_usage_test.go` files across multiple service packages to comply with stricter vet rules introduced in Go 1.24.

- **WHAT**: Renamed example functions so that their names match real exported identifiers.
- **WHY**: In Go 1.24, the `tests` analyzer in `go vet` enforces that example function names must map to valid exported identifiers. These stricter checks caused previously valid tests to fail during `go test`, even though the actual test logic remained unchanged.

## How is this tested?

Validated by running `make test` locally with Go 1.24. All example tests now pass without build or vet errors.

NO_CHANGELOG=true


parthban-db:
Can we start running our unit tests in go1.24?

tanmay-db (approved):

PR #1205, opened 2025-04-08
[Internal] Remove `TestAccDashboards` as legacy dashboard creation on has been discontinued

tanmay-db:
As of April 7, 2025: Official support for the legacy version of dashboards has ended.
https://docs.databricks.com/gcp/en/sql/user/dashboards. Removing the test and keeping it commented out for context as discussed. 

## How is this tested?
N/A

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -9,5 +9,6 @@
 ### Documentation
 
 ### Internal Changes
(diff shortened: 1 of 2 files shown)


mgyucht (commented):
Can we just disable the test entirely? The feature is being wound down, and this will probably start failing in other clouds as well.

mgyucht (approved):

PR #1215, opened 2025-05-05
Add support for OIDC ID token authentication using an environment variable.

renaudhartert-db:

This PR adds a new way to authenticate by reading OIDC ID tokens from environment variables. By default, the new credential strategy attempts to read the token from `DATABRICKS_OIDC_TOKEN`. This default value can be overwritten by setting `DATABRICKS_OIDC_TOKEN_ENV`. 

The new authentication mode will be tested before Github OIDC. The rationale is that we consider setting up `DATABRICKS_OIDC_TOKEN` as a stronger signal of intent than enabling OIDC for the whole Github Action.

This PR also moves `IDTokenSource` in its own `oidc` package within the `experimental/auth` package to clarify that these interfaces are still being validated.

## How is this tested?

Complete test coverage of the new source.

### DIFF
--- a/config/auth_default.go
+++ b/config/auth_default.go
@@ -18,14 +18,6 @@ func buildOidcTokenCredentialStrategies(cfg *Config) []CredentialsStrategy {
 	}
 	idTokenSources := []namedIdTokenSource{
 		{
-			name: "github-oidc",
-			tokenSource: &githubIDTokenSource{
-				actionsIDTokenRequestURL:   cfg.ActionsIDTokenRequestURL,
-				actionsIDTokenRequestToken: cfg.ActionsIDTokenRequestToken,
-				refreshClient:              cfg.refreshClient,
-			},
-		},
-		{
 			name: "env-oidc",
 			// If the OIDCTokenEnv is not set, use DATABRICKS_OIDC_TOKEN as
 			// default value.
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -108,6 +108,9 @@ type Config struct {
 	// specified by this argument. This argument also holds currently selected auth.
 	AuthType string `name:"auth_type" env:"DATABRICKS_AUTH_TYPE" auth:"-"`
 
+	// Environment variable name that contains an OIDC ID token.
+	OIDCTokenEnv string `name:"oidc_token_env" env:"DATABRICKS_OIDC_TOKEN_ENV" auth:"-"`
```
hectorcast-db, line 112: Does it make sense to read this dynamically from the environment to support "refreshes"?
Note that neither AWS, GCP or Azure seem to support the refresh.

config/auth_default.go
```diff
@@ -24,8 +25,21 @@ func buildOidcTokenCredentialStrategies(cfg *Config) []CredentialsStrategy {
 				refreshClient:              cfg.refreshClient,
 			},
 		},
+		{
+			name: "env-oidc",
```
hectorcast-db, line 29: For this and file-oidc, it may make sense to add it to the top of the list, since a user adding `DATABRICKS_OIDC_TOKEN` is clearly not choosing the github-oidc one.

renaudhartert-db: Agreed 👍

config/config.go
renaudhartert-db: The actual environment variable referred to by this environment variable is read each time the IDTokenSource is called. I'm not sure it's worth also reading this one dynamically.

hectorcast-db: True. I missread this as if `DATABRICKS_OIDC_TOKEN_ENV` itself was the token.

hectorcast-db (approved):

PR #1216, opened 2025-05-07
Add external_access_enabled parameter to Metastore API (Fixes #1200)

### LINKED ISSUE #1200: Enable external data access on the metastore by terraform
**Problem Statement**
The current implementation of the databricks metastore Terraform resources does not support the option of "external_access_enabled" check API link https://docs.databricks.com/api/workspace/metastores/update#external_access_enabled
**Proposed Solution**
Modify or create a new terraform resource in order to be able to enable or disable this option of "external_access_enabled"

Krishnan9074:

This PR adds support for the `external_access_enabled` parameter in the Databricks Metastore API to resolve issue #1200.

**WHAT:**
- Added the `ExternalAccessEnabled` field to the `UpdateMetastore` struct in `service/catalog/model.go`
- Updated the example test in `service/catalog/metastores_usage_test.go` to demonstrate usage of the new field
- The field is properly tagged with `json:"external_access_enabled,omitempty"` for API compatibility

**WHY:**
This change is needed because the current implementation of the Databricks metastore Terraform resources does not support the option of `external_access_enabled`. This parameter is documented in the [Databricks API reference](https://docs.databricks.com/api/workspace/metastores/update#external_access_enabled) and allows or prevents non-Databricks Runtime clients from directly accessing entities under the metastore.

Adding this field will enable Terraform users to control this important security setting through the Databricks Terraform provider, as requested in issue #1200.

## How is this tested?

The change has been tested by:
1. Building the code with `go build -mod vendor ./...` to verify compilation
2. Updating the example test to include the new field in the `UpdateMetastore` call

The full end-to-end functionality would be tested when integrated with the Databricks Terraform Provider, but this PR only implements the necessary SDK changes to enable that integration.

Fixes #1200


UriSaShavit:
Hi there,
Apologies for dropping in uninvited — just wanted to kindly ask if there’s any update on this PR?
Thanks in advance!

PR #1223, opened 2025-05-20
Return auth error when auth type is set

renaudhartert-db:

This PR refactors the default auth flow to surface authentication errors when an auth type is specified. 

One side effect of this change is that errors that were previously surfaced during the default auth flow are now swallowed and only accessible via trace log.  

The PR also slightly simplify the token source strategy.

## How is this tested?

Unit and integration tests.

### DIFF
--- a/config/auth_default.go
+++ b/config/auth_default.go
@@ -76,7 +76,7 @@ func (c *DefaultCredentials) Configure(ctx context.Context, cfg *Config) (creden
 		logger.Tracef(ctx, "Attempting to configure auth: %q", s.Name())
 		cp, err := s.Configure(ctx, cfg)
 		if err != nil || cp == nil {
-			logger.Tracef(ctx, "Failed to configure auth: %q", s.Name())
+			logger.Debugf(ctx, "Failed to configure auth: %q", s.Name())
 			continue
 		}
 		c.name = s.Name()
--- a/config/auth_default.go
+++ b/config/auth_default.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/databricks/databricks-sdk-go/config/credentials"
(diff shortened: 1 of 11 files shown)


config/auth_default.go
```diff
@@ -2,125 +2,123 @@ package config
...
+	strategies := []CredentialsStrategy{
 		PatCredentials{},
 		BasicCredentials{},
 		M2mCredentials{},
-		DatabricksCliCredentials,
-		MetadataServiceCredentials{})
-	strategies = append(strategies, buildOidcTokenCredentialStrategies(cfg)...)
-	strategies = append(strategies,
-		// Attempt to configure auth from most specific to most generic (the Azure CLI).
+		u2mCredentials{},
```
parthban-db, line 43: I don't know if this is expected or not. But databricksCli is replaced with u2m.

config/auth_default.go
```diff
@@ -2,125 +2,123 @@ package config
...
+			}
+		}
+		return nil, fmt.Errorf("auth type %q not found, please check %s for a list of supported auth types", cfg.AuthType, authDocURL)
+	}
+
+	// If no auth type is specified, try the strategies in order. If a strategy
+	// succeeds, returns the credentials provider. If a strategy fails, swallow
+	// the error and try the next strategy.
+	for _, s := range strategies {
+		logger.Tracef(ctx, "Attempting to configure auth: %q", s.Name())
```
parthban-db, line 76: Why at the trace level and not debug?

config/auth_default.go
```diff
@@ -2,125 +2,123 @@ package config
...
+		cp, err := s.Configure(ctx, cfg)
+		if err != nil || cp == nil {
+			logger.Tracef(ctx, "Failed to configure auth: %q", s.Name())
```
parthban-db, line 79: ditto

config/auth_default_test.go
```diff
@@ -1,19 +1,48 @@
...
+	}
+	if gotErr == nil {
+		t.Errorf("DefaultCredentials.Configure: got error %v, want non-nil", gotErr)
+	}
+	if !strings.Contains(gotErr.Error(), "auth type \"unknown-mode-for-test\" not found") {
+		t.Errorf("DefaultCredentials.Configure: got error %v, want error containing \"auth type \"unknown-mode-for-test\" not found\"", gotErr)
+	}
+}
+
+func TestDefaultCredentials_Configure_noValidAuth(t *testing.T) {
```
parthban-db, line 30: Why not table tests?

renaudhartert-db: No strong reason, the tests started different but ultimately converged to the same pattern. Happy to change that.

config/auth_default.go
renaudhartert-db: Done.

parthban-db (approved):

PR #1233, opened 2025-06-18
Allow localhost on u2m auth

shreyas-goenka:

This PR relaxes the validation performed on the host URL to allow `http` URLs if the URL points to localhost.

This is needed to add acceptance tests in the Databricks CLI for the `databricks auth login` command. For context: The acceptance test framework in the CLI has the ability to spawn a testserver on localhost and record API requests sent to it. We can use the testserver to validate that the correct API requests are being sent during `auth login`, and that the command does not fail.

Prompted by https://github.com/databricks/cli/pull/2988#discussion_r2154081183 where I almost let a big regression pass.

NO_CHANGELOG=true

## How is this tested?

Unit test


credentials/u2m/workspace_oauth_argument.go
```diff
@@ -22,13 +22,24 @@ type BasicWorkspaceOAuthArgument struct {
 	host string
 }
 
-// NewBasicWorkspaceOAuthArgument creates a new BasicWorkspaceOAuthArgument.
-func NewBasicWorkspaceOAuthArgument(host string) (BasicWorkspaceOAuthArgument, error) {
+func validateHost(host string) error {
+	if strings.HasPrefix(host, "http://127.0.0.1") {
+		// Allow http for localhost
```
renaudhartert-db, line 27: [nit] The code is self-explanatory, either remove the comment or — better — explain _why_ we are doing this (e.g. to enable local testing).

renaudhartert-db (approved):

PR #1238, opened 2025-06-26
Remove usage of testify in `persistent_auth_test.go`.

renaudhartert-db:

This PR removes usage of `testify` in `persistent_auth_test`. The change is not focused on improving the test ergonomy but rather in cutting the dependency on `testify` in preparation to another PR.

## How is this tested?

N/A

NO_CHANGELOG=true

### DIFF
--- a/credentials/u2m/persistent_auth_test.go
+++ b/credentials/u2m/persistent_auth_test.go
@@ -1,4 +1,4 @@
-package u2m_test
+package u2m
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


credentials/u2m/persistent_auth_test.go
```diff
@@ -22,51 +20,59 @@ type tokenCacheMock struct {
 }
 
 func (m *tokenCacheMock) Store(key string, t *oauth2.Token) error {
-	if m.store == nil {
-		panic("no store mock")
-	}
```
parthban-db, line 27: Just for my understanding, why are we removing this?

parthban-db (approved):

credentials/u2m/persistent_auth_test.go
renaudhartert-db: The extra code has very little value. If the test was broken, we would still receive a panic due to the nil pointer exception which will have a very clear root cause.

PR #1245, opened 2025-07-07
Start decoupling `ApiClient` and `DatabricksClient`.

renaudhartert-db:

The SDK has two clients which serve very similar purposes: `ApiClient` and `DatabricksClient`. This PR is a first step to pay to merge these two client into a single consistent client. Specifically, it brings some of the functionality from `DatabricksClient` in `ApiClient` in preparation for a change of code generation logic.

This PR also slightly changes the way the `ApiClient` default parameter values are set by moving the logic in its constructor. 

## How is this tested?

Unit and integration tests.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -15,13 +15,12 @@ func New(cfg *config.Config) (*DatabricksClient, error) {
 	if err != nil {
 		return nil, err
 	}
-	client, err := cfg.NewApiClient()
+	clientCfg, err := config.HTTPClientConfigFromConfig(cfg)
 	if err != nil {
 		return nil, err
 	}
 	return &DatabricksClient{
-		Config: cfg,
-		client: client,
+		client: httpclient.NewApiClient(clientCfg),
 	}, nil
 }
 
--- a/client/client.go
+++ b/client/client.go
@@ -20,6 +20,7 @@ func New(cfg *config.Config) (*DatabricksClient, error) {
 		return nil, err
 	}
 	return &DatabricksClient{
+		Config: cfg,
 		client: httpclient.NewApiClient(clientCfg),
 	}, nil
 }
--- a/common/environment/environments.go
+++ b/common/environment/environments.go
@@ -8,10 +8,9 @@ import (
 type Cloud string
 
 const (
-	CloudUnknown Cloud = "Unknown"
-	CloudAWS     Cloud = "AWS"
-	CloudAzure   Cloud = "Azure"
-	CloudGCP     Cloud = "GCP"
+	CloudAWS   Cloud = "AWS"
+	CloudAzure Cloud = "Azure"
+	CloudGCP   Cloud = "GCP"
 )
 
 type DatabricksEnvironment struct {
--- a/common/environment/environments.go
+++ b/common/environment/environments.go
@@ -8,9 +8,10 @@ import (
 type Cloud string
 
 const (
-	CloudAWS   Cloud = "AWS"
-	CloudAzure Cloud = "Azure"
-	CloudGCP   Cloud = "GCP"
+	CloudUnknown Cloud = "Unknown"
+	CloudAWS     Cloud = "AWS"
+	CloudAzure   Cloud = "Azure"
+	CloudGCP     Cloud = "GCP"
 )
 
 type DatabricksEnvironment struct {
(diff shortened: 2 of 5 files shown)


config/api_client.go
```diff
@@ -9,47 +9,56 @@ import (
...
-		DebugHeaders:       c.DebugHeaders,
-		DebugTruncateBytes: c.DebugTruncateBytes,
-		InsecureSkipVerify: c.InsecureSkipVerify,
-		Transport:          c.HTTPTransport,
-		AuthVisitor:        c.Authenticate,
+
+	cloud := cloudFromConfig(cfg)
+	if cloud == environment.CloudUnknown {
+		return httpclient.ClientConfig{}, fmt.Errorf("unable to determine cloud from config")
+	}
```
parthban-db, line 34: Don't we have truly serverless environments like the community edition? I am not sure if this safe change.

common/environment/environments.go
```diff
@@ -8,9 +8,10 @@ import (
 type Cloud string
 
 const (
-	CloudAWS   Cloud = "AWS"
-	CloudAzure Cloud = "Azure"
-	CloudGCP   Cloud = "GCP"
+	CloudUnknown Cloud = "Unknown"
+	CloudAWS     Cloud = "AWS"
+	CloudAzure   Cloud = "Azure"
+	CloudGCP     Cloud = "GCP"
```
parthban-db, line 14: What about truly serverless environments?

client/client.go
```diff
@@ -15,13 +15,12 @@ func New(cfg *config.Config) (*DatabricksClient, error) {
 	if err != nil {
 		return nil, err
 	}
-	client, err := cfg.NewApiClient()
+	clientCfg, err := config.HTTPClientConfigFromConfig(cfg)
 	if err != nil {
 		return nil, err
 	}
 	return &DatabricksClient{
-		Config: cfg,
```
parthban-db, line 23: This config is public. Can we simply stop setting this? Is this safe? It will be a behavioral change. If you really don't want to set this, let's make it private so the user doesn't get fake expectations from it.

common/environment/environments.go
renaudhartert-db: Fair point, I think I'm going to remove the cloud change and rather try to better understand whether `compute` actually truly need to have the cloud info. It seems it is only use to set an AWS property in a request and I feel like it is not the responsibility of the function to that. In any case, this is out of scope for this PR.

config/api_client.go
```diff
@@ -14,42 +14,44 @@ import (
...
-		InsecureSkipVerify: c.InsecureSkipVerify,
-		Transport:          c.HTTPTransport,
-		AuthVisitor:        c.Authenticate,
+
+	return httpclient.ClientConfig{
+		AccountID:          cfg.AccountID,
+		Host:               cfg.Host,
+		RetryTimeout:       time.Duration(cfg.RetryTimeoutSeconds) * time.Second,
+		HTTPTimeout:        time.Duration(cfg.HTTPTimeoutSeconds) * time.Second,
+		RateLimitPerSecond: cfg.RateLimitPerSecond,
```
parthban-db, line 35: Doesn't having no defaults change the behavior?

httpclient/api_client.go
```diff
@@ -73,10 +105,13 @@ func (cfg ClientConfig) httpTransport() http.RoundTripper {
 }
 
 func NewApiClient(cfg ClientConfig) *ApiClient {
+	// Set defaults for config values that are not set.
 	cfg.HTTPTimeout = time.Duration(orDefault(int(cfg.HTTPTimeout), int(30*time.Second)))
 	cfg.DebugTruncateBytes = orDefault(cfg.DebugTruncateBytes, 96)
 	cfg.RetryTimeout = time.Duration(orDefault(int(cfg.RetryTimeout), int(5*time.Minute)))
-	cfg.HTTPTimeout = time.Duration(orDefault(int(cfg.HTTPTimeout), int(30*time.Second)))
+	cfg.HTTPTimeout = time.Duration(orDefault(int(cfg.HTTPTimeout), int(60*time.Second)))
```
denik, line 112: Was this change intentional? It is not mentioned in the description. Also there is another line above setting cfg.HTTPTimeout to 30s.

PR #1261, opened 2025-08-07
Include ResponseWrapper in APIError

denik:

APIError includes ResponseWrapper field.

This allows for clients to generate their own error message that includes this critical bits:
- method + full URL
- HTTP status code

For more verbose modes:
- HTTP status message
- request/response headers 
- request/response bodies

I plan to use it in CLI/DABs, currently I'm annotating errors with endpoint name, I'd be able to get rid of this boilerplate: https://github.com/databricks/cli/blob/0a7bd1a/bundle/terranova/tnresources/job.go#L34-L36

## How is this tested?

Updated existing unit tests to (partially) compare this new field.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Release v0.80.0
 
 ### New Features and Improvements
+* Added `ResponseWrapper` field to `APIError` struct ([1261](https://github.com/databricks/databricks-sdk-go/pull/1261)).
 
 ### Bug Fixes
 


denik:
Thanks! This is CLI integration: https://github.com/databricks/cli/pull/3379 that shows how we going to use that:

```
Error: deploying jobs.foo: creating: Shared job cluster feature is only supported in multi-task jobs. (400 INVALID_PARAMETER_VALUE)

Endpoint: POST [DATABRICKS_URL]/api/2.2/jobs/create
HTTP Status: 400 Bad Request
API error_code: INVALID_PARAMETER_VALUE
API message: Shared job cluster feature is only supported in multi-task jobs.
```

PR #1273, opened 2025-08-21
Test Well Known Types

hectorcast-db:
DO NOT MERGE


common/types/duration.go
```diff
@@ -0,0 +1,88 @@
...
+// It embeds time.Duration, so all standard methods (Seconds, String, etc.)
+// are directly accessible. The underlying time.Duration value can be
+// accessed via the .Duration field.
+//
+// Example:
+//
+//	customDur := types.NewDuration(30 * time.Second)
+//	goDur := customDur.Duration // Access the underlying time.Duration
+type Duration struct {
+	time.Duration
```
parthban-db, line 22: I thought this would be a wrapper over the Google types, as the marshalling/unmarshalling function will be defined by the Google library. It's just that it is not a JSON marshal.

parthban-db (commented):
I think we should use the google type instead of the standard types. As that's what we will offer in the future.

common/types/duration.go
hectorcast-db: We were split between native types (better CUX) and well known types (easier code) and decided for the second. Since well known types do support native json marshall, we do need to wrap and convert anyway. So the reason to use well known does not exist anymore.

PR #1302, opened 2025-10-06
Minor refactor of `api_client.go` to better highlight the client's default value.

renaudhartert-db:

This PR is a no-op that isolates the logic to set default values from the rest of the constructor. The goal is to make the default value more apparent. 

## How is this tested?

Unit + Integration tests.


Divyansh-db (commented):
The orDefault function needs to be removed as it is no longer used

PR #1331, opened 2025-11-14
[Fix] Support for Spark versions with multiple Scala versions

alexott:

This is a fix for https://github.com/databricks/terraform-provider-databricks/issues/5218. The issue is cased by the fact that `Scala` selector had default value of `2.12` (for Terraform only), and as result was filtering out 17.x versions that are released only with Scala 2.13.

Fix is to relax the default value, but handle the case when there could be 2.12 and 2.13 versions for the same DBR version (16.x only).  

Removed all Terraform customizations from the Go SDK struct, will be handled separately in Terraform itself

Also bumped staticcheck to 0.6.1 to support compilation with Go 1.25

## How is this tested?

Added unit test, tested manually


service/compute/ext_spark_version.go
```diff
@@ -76,7 +76,28 @@ func (sv GetSparkVersionsResponse) Select(req SparkVersionRequest) (string, erro
...
+						break
+					}
+				}
+				if !found {
+					uniqueVersions = append(uniqueVersions, v)
+				}
+			}
+			if len(uniqueVersions) > 1 {
+				return "", fmt.Errorf("spark versions query returned multiple results %#v. Please change your search criteria and try again", uniqueVersions)
+			}
```
tanmay-db: Does this mean the list size should be 2? i.e. same version with 2.12 and 2.13? If yes, we can just check these 2 values rather than sort and loop.

service/compute/ext_spark_version.go
```diff
@@ -13,16 +13,16 @@ import (
...
+	LongTermSupport bool   `json:"long_term_support,omitempty"`
+	Beta            bool   `json:"beta,omitempty"`
+	Latest          bool   `json:"latest,omitempty"`
+	ML              bool   `json:"ml,omitempty"`
+	Genomics        bool   `json:"genomics,omitempty"`
+	GPU             bool   `json:"gpu,omitempty"`
+	Scala           string `json:"scala,omitempty"`
+	SparkVersion    string `json:"spark_version,omitempty"`
+	Photon          bool   `json:"photon,omitempty"`
+	Graviton        bool   `json:"graviton,omitempty"`
```
tanmay-db: Why are tf tags not needed now?

service/compute/ext_spark_version.go
tanmay-db: @alexott, I meant that if the list size >=3, then we already error because the duplicate will only be possible for 2 different scala versions right? And if list size == 2, then we check those 2 elements

alexott: Yes, you're right... Let me reimplement it a bit differently

tanmay-db (approved):

PR #1374, opened 2025-12-15
Add custom scopes support in OAuth authentication methods

tejaskochar-db:
## Summary

 + Adds new `Scopes` and `DisableOAuthRefreshToken` fields.
 + Adds support for parsing lists in config files.

## Notes

- Environment variable support is **not** provided for scopes because we do not think users would actually want this. It can be easily added as a follow up if there are requests for it.

## Testing
 - Parsing logic tested by loading profiles from a test `.databrickscfg` file.

Subsequent PRs add support for custom scopes in OAuth authentication methods:
 - M2M: https://github.com/databricks/databricks-sdk-go/pull/1388
 - WIF: https://github.com/databricks/databricks-sdk-go/pull/1389
 - U2M: https://github.com/databricks/databricks-sdk-go/pull/1390

---
NO_CHANGELOG=true


config/config_attribute.go
```diff
@@ -69,6 +70,11 @@ func (a *ConfigAttribute) SetS(cfg *Config, v string) error {
 			return err
 		}
 		return a.Set(cfg, vv)
+	case reflect.Slice:
+		parts := strings.FieldsFunc(v, func(r rune) bool {
+			return r == ',' || r == ' ' || r == '\t' || r == '\n' || r == '\r'
+		})
```
parthban-db: similar to my java comment: why we need this many delimiter? Also, I think we don't have a coverage of weird cases which will arise from this many delimiter.

Tejas-Kochar: Fixed this to only use comma as delimiter and trim spaces afterwards.

config/config_file_test.go
```diff
@@ -22,3 +23,39 @@ func TestConfigFileLoad(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			env.CleanupEnvironment(t)
+			t.Setenv("HOME", "testdata")
+
+			cfg := &Config{Profile: tt.profile}
+			err := cfg.EnsureResolved()
+			require.NoError(t, err)
+			assert.Equal(t, tt.expected, cfg.GetScopes())
```
parthban-db: We are proactively removing the testify library from our codebase. So, we shouldn't add new code with testify. Can you use the cmp library instead for comparison?

config/config.go
```diff
@@ -135,6 +138,25 @@ type Config struct {
...
+	// compatibility for this feature.
+	//
+	// Scopes is a list of OAuth scopes to request when authenticating.
+	// If not specified, defaults to ["all-apis"] for backwards compatibility.
+	// For U2M authentication, "offline_access" is automatically appended to
+	// include a refresh token (disable via DisableOAuthRefreshToken).
+	//
+	// Note: Setting scopes via environment variables is not supported.
+	// Note: The slice is sorted in-place during config resolution.
+	// Note: U2M flow's token cache does not support scopes yet.
```
parthban-db: IIUC, with the splitting in the PR. This PR doesn't add support of scopes even in M2M authentication. So, I think we should update the description to say this feature is still in development and may not work as expected.

config/config.go
```diff
@@ -445,6 +469,11 @@ func (c *Config) EnsureResolved() error {
 			},
 		}
 	}
+	// Sort scopes in-place for better de-duplication in the refresh token cache,
+	// once scopes are supported in its cache key.
```
renaudhartert-db: Should we also dedup here?

config/config_file_test.go
```diff
@@ -22,3 +24,43 @@ func TestConfigFileLoad(t *testing.T) {
...
+		{
+			name:     "multiple scopes sorted",
+			profile:  "scope-multiple",
+			expected: []string{"clusters", "files:read", "iam:read", "jobs", "mlflow", "model-serving", "pipelines"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			env.CleanupEnvironment(t)
```
renaudhartert-db: Is this needed? I believe `t.Setenv` automatically undoes the environment setting when the test completes.

config/config.go
Tejas-Kochar: I couldn't think of a good reason not to, so I added it.

config/config_file_test.go
Tejas-Kochar: It does indeed undo environment settings but its seems there are other environment variables set in the test environment that interfere with the test. I tried removing the line and the test started failing because it could not find the config file anymore.

config/config_test.go
```diff
@@ -299,3 +300,61 @@ func TestConfig_getOAuthArgument_Unified(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := sortAndDedupeSlice(tt.input)
+			if diff := cmp.Diff(tt.expected, result); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
```
renaudhartert-db: nit: we generally try to stick to "want" and "got". 

```suggestion
		want []string
	}{
		{
			name:     "nil slice",
			input:    nil,
			want: nil,
		},
		{
			name:     "empty slice",
			input:    []string{},
			want: []string{},
		},
		{
			name:     "single element",
			input:    []string{"a"},
			want: []string{"a"},
		},
		{
			name:     "already sorted no duplicates",
			input:    []string{"a", "b", "c"},
			want: []string{"a", "b", "c"},
		},
		{
			name:     "unsorted no duplicates",
			input:    []string{"c", "a", "b"},
			want: []string{"a", "b", "c"},
		},
		{
			name:     "with duplicates",
			input:    []string{"b", "a", "b", "c", "a"},
			want: []string{"a", "b", "c"},
		},
		{
			name:     "all duplicates",
			input:    []string{"a", "a", "a"},
			want: []string{"a"},
		},
		{
			name:     "adjacent duplicates",
			input:    []string{"a", "a", "b", "b", "c"},
			want: []string{"a", "b", "c"},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got := sortAndDedupeSlice(tt.input)
			if diff := cmp.Diff(tt.want, got); diff != "" {
				t.Errorf("mismatch (-want +got):\n%s", diff)
			}
		})
	}
```

config/config.go
```diff
@@ -445,10 +468,29 @@ func (c *Config) EnsureResolved() error {
 			},
 		}
 	}
+	// Sort and deduplicate scopes in-place for better cache key matching
+	// in the refresh token cache.
+	c.Scopes = sortAndDedupeSlice(c.Scopes)
 	c.resolved = true
```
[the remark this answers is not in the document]
Tejas-Kochar: `slices` was introduced in go 1.21, while we support go 1.18+. I have done what I found to be the standard practice from when Compact wasn't introduced.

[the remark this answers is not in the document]
Tejas-Kochar: Nice. I've switched from the helper to using `slices`.

config/config_file_test.go
```diff
@@ -3,22 +3,79 @@ package config
...
+		if section == nil {
+			t.Fatalf("expected section %q to be non-nil", name)
+		}
+		if got, want := section.Key("password").String(), "%Y#X$Z"; got != want {
+			t.Errorf("password mismatch for %q: got %q, want %q", name, got, want)
+		}
+	}
+}
+
+func TestConfigFile_Scopes(t *testing.T) {
```
renaudhartert-db: Should we remove this test? It looks redundant with the two tests we have added. If you decide to keep it, please update:

```go
			withMockEnv(t, map[string]string{})
			t.Setenv("HOME", "testdata/scopes")
```

to

```go
			withMockEnv(t, map[string]string{"HOME": "testdata/scopes"})
```

PR #1388, opened 2025-12-22
Custom scopes support in M2M authentication

tejaskochar-db:
## Summary
One-line change passing configured scopes instead of hard-coded value "all-apis".

You might also see changes from parent branch https://github.com/databricks/databricks-sdk-go/pull/1374

## Testing
Test the (mocked) authentication of requests configured for M2M with various different scopes (nil, empty, single, multiple unsorted).
Custom scopes tested for both Account and Workspace level requests.

---
NO_CHANGELOG=true

### DIFF
--- a/config/auth_m2m_test.go
+++ b/config/auth_m2m_test.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
-	"strings"
 	"testing"
 
 	"github.com/databricks/databricks-sdk-go/credentials/u2m"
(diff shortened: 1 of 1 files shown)


config/auth_m2m_test.go
```diff
@@ -90,5 +92,79 @@ func TestM2mNotSupported(t *testing.T) {
...
+				Host:          "a",
+				ClientID:      "b",
+				ClientSecret:  "c",
+				AuthType:      "oauth-m2m",
+				Scopes:        tt.scopes,
+				HTTPTransport: transport,
+				ConfigFile:    "/dev/null",
+			}
+
+			req, _ := http.NewRequest("GET", "http://localhost", nil)
```
parthban-db: Please catch the error here. We shouldn't suppress error. This will make the test pass even if something bad happens.

config/auth_m2m_test.go
```diff
@@ -90,5 +92,79 @@ func TestM2mNotSupported(t *testing.T) {
 			},
 		},
 	})
-	require.ErrorIs(t, err, u2m.ErrOAuthNotSupported)
+	if !errors.Is(err, u2m.ErrOAuthNotSupported) {
```
parthban-db: 💯

PR #1389, opened 2025-12-22
Custom scopes support in WIF authentication

tejaskochar-db:
## Summary

 - 5-line change (in `auth_default.go` and `tokensource.go`) passing scopes through in WIF authentication instead of using "all-apis"

## Testing

 - Integration tests with mocking ensuring scopes are propogated and used for token exchange (`auth_default_test.go`).
 - Unit tests to ensure custom scopes propogated in different account and workspace configurations (`tokensource_test.go`).
---
NO_CHANGELOG=true


config/experimental/auth/oidc/tokensource_test.go
```diff
@@ -275,6 +275,7 @@ func TestDatabricksOidcTokenSource(t *testing.T) {
 				ClientID:              tc.clientID,
 				AccountID:             tc.accountID,
 				Host:                  tc.host,
+				Scopes:                []string{"all-apis"},
```
parthban-db: Who is the user of DatabricksOIDCTokenSourceConfig? Is it internal only or can it be external users? It seems we have added a backward incompatible change in which we need to set the scopes value.

config/experimental/auth/oidc/tokensource_test.go
```diff
@@ -319,3 +320,76 @@ func TestDatabricksOidcTokenSource(t *testing.T) {
...
+			}
+
+			ts := NewDatabricksOIDCTokenSource(cfg)
+
+			// The scope assertion: verifies the token source sends the correct scope parameter.
+			expectedRequest := url.Values{
+				"client_id":          {"client-id"},
+				"scope":              {tt.want},
+				"subject_token_type": {"urn:ietf:params:oauth:token-type:jwt"},
+				"subject_token":      {"id-token"},
```
parthban-db: I believe the value here and in line 354 should be same? Can you make this as a shared variable to convey the intend to the reader?

config/auth_default_test.go
```diff
@@ -47,3 +52,95 @@ func TestDefaultCredentials_Configure(t *testing.T) {
...
+				Host:                       databricksServer.URL,
+				ClientID:                   "test-client-id",
+				ActionsIDTokenRequestURL:   githubServer.URL + "/github-token?version=1",
+				ActionsIDTokenRequestToken: "github-request-token",
+				TokenAudience:              "databricks-test-audience",
+				AuthType:                   "github-oidc",
+				Scopes:                     tt.scopes,
+			}
+
+			req, _ := http.NewRequest("GET", databricksServer.URL+"/api/test", nil)
```
parthban-db: Ditto we shouldn't suppress the error.

config/experimental/auth/oidc/tokensource.go
```diff
@@ -39,6 +39,20 @@ type DatabricksOIDCTokenSourceConfig struct {
 
 	// IDTokenSource returns the IDToken to be used for the token exchange.
 	IDTokenSource IDTokenSource
+
+	// scopes is the list of OAuth scopes to request.
+	scopes []string
+}
+
+func (c *DatabricksOIDCTokenSourceConfig) GetScopes() []string {
```
parthban-db: Can you add a doccomment, as this is a public method?

PR #1406, opened 2026-01-08
Make token store writes atomic

tejaskochar-db:

This makes writes to the token store to avoid cache corruption from concurrent writes. It does this using write-then-rename.

## How is this tested?

Existing token store tests (`credentials/u2m/cache/file_test.go`).

Manually tested databricks-cli auth type with existing profile, it continued to work.

---
NO_CHANGELOG=true

### DIFF
--- a/credentials/u2m/cache/file.go
+++ b/credentials/u2m/cache/file.go
@@ -103,7 +103,10 @@ func (c *fileTokenCache) Store(key string, t *oauth2.Token) error {
 	if err != nil {
 		return fmt.Errorf("marshal: %w", err)
 	}
-	return c.atomicWriteFile(raw)
+	if err := c.atomicWriteFile(raw); err != nil {
+		return fmt.Errorf("error storing token in local cache: %w", err)
+	}
+	return nil
 }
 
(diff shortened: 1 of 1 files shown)


credentials/u2m/cache/file.go
```diff
@@ -177,3 +177,43 @@ func (c *fileTokenCache) load() (*tokenCacheFile, error) {
...
+		return fmt.Errorf("create temp file: %w", err)
+	}
+	tmpName := tmp.Name()
+
+	success := false
+	defer func() {
+		if !success {
+			os.Remove(tmpName)
+		}
+	}()
```
renaudhartert-db: You can simply do this I believe and remove subsequent calls to Close.

```suggestion
	tmp, err := os.CreateTemp(dir, ".token-cache-*.tmp")
	if err != nil {
		return fmt.Errorf("create temp file: %w", err)
	}
	defer func() {
		tmp.Close()
		os.Remove(tmp.Name()) // clean-up
	}()
```

renaudhartert-db (approved):
👍

PR #1415, opened 2026-01-14
Custom formatting for slice type config attributes

tejaskochar-db:
## What changes

Format what GetString returns for slice type config attributes as a,b,c instead of the default [a b c].

## Why
The CLI uses this function for saving profiles to the .databrickscfg file. The convention for lists in INI files is comma delimited. 

## How is this tested?
Added unit tests

---
NO_CHANGELOG=true

### DIFF
--- a/config/config_attributes_test.go
+++ b/config/config_attributes_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


config/config_attribute.go
```diff
@@ -122,6 +122,10 @@ func (a *ConfigAttribute) IsZero(cfg *Config) bool {
 func (a *ConfigAttribute) GetString(cfg *Config) string {
 	rv := reflect.ValueOf(cfg)
 	field := rv.Elem().Field(a.num)
+	if a.Kind == reflect.Slice {
```
hectorcast-db: Let's use a switch/case for future proofing. And let's add some tests.

hectorcast-db (approved):

PR #1484, opened 2026-02-18
Add DefaultCredentialStrategy package-level variable.

renaudhartert-db:
This PR provides a way for users to change the default credentials strategy used in config resolution. The pattern is similar to the one used in the standard `http` and `log` packages.

**Why a producer?** The current default credentials strategy has state and needs to be generated on the fly.

## Usage

```go
  config.DefaultCredentialStrategyProvider = func() config.CredentialsStrategy {
      return &myCustomStrategy{}
  }
```

## Test plan

- New TestDefaultCredentialStrategy verifies the factory is used when Config.Credentials is nil

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -14,4 +14,6 @@
 
 ### Internal Changes
 
+* Make clusters creation in AWS not depend on cloud parsed from host. Changed default AWS availability for auto-created utility clusters from SPOT to SPOT_WITH_FALLBACK (API default). 
+
 ### API Changes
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -66,6 +66,17 @@ const (
 
 var defaultScopes = []string{"all-apis"}
 
+// DefaultCredentialStrategy is the default factory for creating a
+// [CredentialsStrategy] when [Config.Credentials] is not set. It can be
+// overridden to change the default authentication behavior for all Config
+// instances.
```
hectorcast-db: Shall we document here whether this is thread safe?

config/config.go
```diff
@@ -66,6 +66,17 @@ const (
...
+//
+// Each invocation must return a new instance to avoid shared mutable state
+// between Config instances.
+var DefaultCredentialStrategy func() CredentialsStrategy = func() CredentialsStrategy {
```
hectorcast-db: Nit. Not fully convinced about the name. Being a function I would expect either `NewDefaultCredentialStrategy` or `DefaultCredentialStrategyFactory`.

PR #1501, opened 2026-02-25
Add DiscoveryURL field and ResolveHostMetadata to Config

hectorcast-db:
## 🥞 Stacked PR
Use this [link](https://github.com/databricks/databricks-sdk-go/pull/1501/files) to review incremental changes.
- [**stack/config-auto-complete-3**](https://github.com/databricks/databricks-sdk-go/pull/1501) [[Files changed](https://github.com/databricks/databricks-sdk-go/pull/1501/files)]

---------
## Changes

Ported from [databricks-sdk-py#1290](https://github.com/databricks/databricks-sdk-py/pull/1290).

Wire `DiscoveryURL` and `ResolveHostMetadata` into `Config`:

- Add `DiscoveryURL` field to `Config` (`discovery_url` / `DATABRICKS_DISCOVERY_URL`).
  When set, `getOidcEndpoints` fetches OIDC endpoints directly from that URL,
  bypassing host-type detection.
- Add `ResolveHostMetadata()` method to `Config`. Fetches
  `/.well-known/databricks-config` and back-fills `AccountID`, `WorkspaceID`,
  and `DiscoveryURL` (substituting any `{account_id}` placeholder). Returns an
  error if `account_id` cannot be resolved or no `oidc_endpoint` is present.

## Tests

New tests in `config/config_test.go`:
- `TestGetOidcEndpoints_DiscoveryURL`: verifies endpoints are fetched from `DiscoveryURL` when set
- `TestResolveHostMetadata_*`: verifies `ResolveHostMetadata` correctly fills in missing fields,
  handles templates, returns errors when required fields are absent, and is a no-op when `Host` is empty

NO_CHANGELOG=true

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -252,6 +252,13 @@ type Config struct {
 
 	// Marker for unified hosts. Will be redundant once we can recognize unified hosts by their hostname.
 	Experimental_IsUnifiedHost bool `name:"experimental_is_unified_host" env:"DATABRICKS_EXPERIMENTAL_IS_UNIFIED_HOST" auth:"-"`
+
+	// OpenID Connect discovery URL. When set, OIDC endpoints are fetched directly
+	// from this URL instead of the default host-type-based well-known endpoint logic.
+	// Mirrors discoveryUrl in the Java SDK.
+	//
+	// Experimental: subject to change.
+	DiscoveryURL string `name:"discovery_url" env:"DATABRICKS_DISCOVERY_URL" auth:"-"`
 }
 
 // NewWithWorkspaceHost returns a new instance of the Config with the host set to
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -620,6 +632,46 @@ func (c *Config) getOidcEndpoints(ctx context.Context) (*u2m.OAuthAuthorizationS
...
+	if err != nil {
+		return err
+	}
+	if c.AccountID == "" && meta.AccountID != "" {
+		logger.Debugf(ctx, "Resolved account_id from host metadata: %q", meta.AccountID)
+		c.AccountID = meta.AccountID
+	}
+	if c.AccountID == "" {
+		return fmt.Errorf("account_id is not configured and could not be resolved from host metadata")
+	}
```
tanmay-db: Should we check for config account ID and meta account ID? i.e. error if they differ (which shouldn't happen right?)

What will happen for case where one spog host can have multiple account IDs in future?

hectorcast-db: > Should we check for config account ID and meta account ID? i.e. error if they differ (which shouldn't happen right?)

It should not unless the user made a mistake.

> What will happen for case where one spog host can have multiple account IDs in future?

Same as for account.databricks.com: We will not get an account ID, so the user must specify it.

PR #1589, opened 2026-03-24
Use pure-Go DNS resolver to fix Private Link/PSC connectivity on macOS

DrFaust92:
## Summary

- Set `PreferGo: true` on the `net.Resolver` used by the SDK's default HTTP transport dialer
- Fixes DNS resolution issues on macOS when using Private Link / Private Service Connect

## Problem

On macOS, Go's default cgo DNS resolver can bypass split-horizon DNS configurations. When a Databricks workspace is configured with Private Link or Private Service Connect, the workspace URL should resolve to a private IP via a private DNS zone. However, the cgo resolver may resolve to the public IP instead, causing requests to be blocked by IP ACLs with a confusing error:

```
Error: cannot update permissions: Source IP address: x.x.x.x is blocked by 
Databricks IP ACL for workspace: xxxxx
```

The workaround today is setting `GODEBUG=netdns=go` globally, but users have no way to know this from the error message alone.

## Solution

Set `PreferGo: true` on the `net.Resolver` in the SDK's default HTTP transport dialer. This:

- Ensures the pure-Go DNS resolver is used, which correctly respects system DNS settings including private DNS zones
- Is scoped to the SDK's HTTP client only — does not affect other libraries or providers in the same process
- Falls back to the cgo resolver automatically if the Go resolver fails (it's a *preference*, not a hard requirement)
- Matches what `GODEBUG=netdns=go` does, but without requiring a global environment variable

## Test plan

- [x] Verified the change compiles
- [ ] Test on macOS with a Private Service Connect workspace — workspace URL should resolve to private IP
- [ ] Test on Linux — no behavioral change expected (Go resolver is already the default on Linux)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* Use pure-Go DNS resolver to fix Private Link/PSC connectivity on macOS ([#1589](https://github.com/databricks/databricks-sdk-go/pull/1589)).
+
 ### Documentation
 
 ### Internal Changes
--- a/httpclient/api_client.go
+++ b/httpclient/api_client.go
@@ -82,6 +82,7 @@ func makeDefaultTransport() *http.Transport {
 		DialContext: (&net.Dialer{
 			Timeout:   30 * time.Second,
 			KeepAlive: 30 * time.Second,
+			Resolver:  &net.Resolver{PreferGo: true},
 		}).DialContext,
 		ForceAttemptHTTP2:     true,
 		MaxIdleConns:          100,


DrFaust92:
To answer your questions briefly:

**(i)** macOS DNS resolution in Go has a well-documented history of issues with VPNs and split-horizon DNS ([golang/go#12524](https://github.com/golang/go/issues/12524), [hashicorp/terraform#3536](https://github.com/hashicorp/terraform/issues/3536), [helm/helm#11807](https://github.com/helm/helm/issues/11807)). While Go 1.20+ improved things significantly by adding direct libSystem bindings, there are still behavioral differences between the cgo path (`getaddrinfo` → mDNSResponder) and the Go-managed path (`res_search` via libSystem) — particularly around how VPN DNS zone changes are picked up. In our case, `GODEBUG=netdns=go` resolves the Private Link resolution failure. With `PreferGo: true` + cgo fallback, there's no downside — if the Go resolver fails, it falls back to the system resolver automatically.

**(ii)** This affects any macOS user with Private Link/PSC behind a corporate VPN — a very common enterprise setup. The error (`Source IP address is blocked by IP ACL`) gives no hint that DNS is the root cause, so users spend hours debugging network/firewall issues. `GODEBUG=netdns=go` isn't a realistic workaround when the error doesn't point to DNS at all.

**(iii)** Typical setup: workspace with IP ACL + Private Link/PSC, macOS + corporate VPN with split-tunnel DNS routing private zones to an internal nameserver. The SDK transport is vanilla `net.Dialer` — nothing unusual — but since it's the HTTP client for all Go-based Databricks tooling (including the Terraform provider, where most reports come from), fixing it here has the broadest impact.

This is actively disruptive to local development on macOS without the fix or the `GODEBUG` workaround.

DrFaust92:
@renaudhartert-db I've rebased this onto the latest `main` (resolved the `NEXT_CHANGELOG.md` conflicts) so it's mergeable again, and `httpclient` tests pass locally.

On your questions about layering — fair points. A couple of things that I think keep this low-risk:

- The change is scoped to the SDK's *own* `makeDefaultTransport`, so it only affects clients the SDK constructs, not `http.DefaultTransport` or other tools in the process. Callers who pass their own transport are unaffected, and the fallback behavior is unchanged.
- I'm not attached to changing the default if you'd rather keep the surface minimal — documenting `GODEBUG=netdns=go` as the supported workaround is a reasonable alternative and I'm happy to switch to that instead.

If it'd help, I can write up the exact split-horizon DNS setup that triggers it (the `/etc/resolver/*` scoped-resolver case) so we're aligned on whether this is a narrow macOS-private-DNS issue or something broader. Let me know which direction you prefer and I'll adjust.

PR #1605, opened 2026-03-30
Fix CLI token source --profile fallback with version detection

mihaimitrea-db:
## Summary

Fix the broken `--profile` fallback in `CliTokenSource` by replacing error-based detection with version-based CLI detection at init time.

## Why

The `--profile` flag on `databricks auth token` is a **global Cobra flag** ([defined as a persistent flag](https://github.com/databricks/cli/blob/77101c9b8567/cmd/root/root.go) on the root command). Old CLIs (< v0.207.1) silently accept it — they don't report `"unknown flag: --profile"` but instead fail later with `"cannot fetch credentials"`. This means the existing `isUnknownFlagError` check ([config/cli_token_source.go:120](https://github.com/databricks/databricks-sdk-go/blob/main/config/cli_token_source.go#L120)) never matches, and the `--host` fallback is **dead code**.

This was verified by testing against CLI v0.207.0 vs v0.207.1:
- v0.207.0: `databricks auth token --profile workspace` → `Error: init: cannot fetch credentials` (not "unknown flag")
- v0.207.1: `databricks auth token --profile workspace` → returns a valid token

### Approaches considered

Three approaches were evaluated for detecting whether the installed CLI supports `--profile`:

- **Error-based detection (try-and-retry)** — the current approach on `main`. Run `databricks auth token --profile <name>` and check whether the error contains `"unknown flag: --profile"`. **This is broken**: because `--profile` is a global Cobra flag, old CLIs accept it silently and fail with a different error (`"cannot fetch credentials"`), so the fallback to `--host` never triggers.

- **`--help` flag parsing** (`databricks auth token --help` + substring matching) was rejected because the `--help` output format is not a stable API. More importantly, `--profile` would appear in `--help` output even on old CLIs that don't actually implement profile-based token lookup — it shows up because it's a global persistent flag, not because the `auth token` subcommand uses it. This approach has the same fundamental flaw as error-based detection.

- **Version detection** (`databricks version` + semver comparison) — **the approach taken here**. Run `databricks version` at init time, parse the semver (e.g., `"Databricks CLI v0.207.1"`), and compare against known minimum versions for each flag. This is reliable because the version string is a stable output format, and the mapping between flags and CLI versions is well-defined ([databricks/cli#855](https://github.com/databricks/cli/pull/855) for `--profile` in v0.207.1). If version detection fails, the SDK falls back to the most conservative command (`--host` only).

### References

- `--profile` support added in CLI v0.207.1: [databricks/cli#855](https://github.com/databricks/cli/pull/855) (Oct 2023)

### Interface changes

None. `CliTokenSource` is not part of the public API surface.

`NewCliTokenSource` now takes `context.Context` as its first parameter, needed for `exec.CommandContext` when running `databricks version`. This is consistent with every `CredentialsStrategy.Configure` method in the codebase, and the single caller (`auth_u2m.go`) already has `ctx` in scope.

### Behavioral changes

- When `cfg.Profile` is set but the CLI is too old (< v0.207.1), the SDK now correctly falls back to `--host`. Previously this fallback was dead code.
- A warning is logged when the `--profile` flag is not supported and the SDK falls back to `--host`.

### Internal changes

- **`cliVersion` type**: semver parsing with `AtLeast()` comparison and `String()` formatting.
- **`getCliVersion(ctx, cliPath)`**: runs `databricks version` and parses the output.
- **`parseCliVersion`**: parses `"Databricks CLI v0.207.1"` → `cliVersion{0, 207, 1}`.
- **`resolveCliCommand`**: bridges version detection and command building. Falls back to zero version on detection failure.
- **`buildCliCommand`**: pure function — takes a version, returns a single resolved command. No exec calls, easy to test.
- **`CliTokenSource`** simplified to a single `cmd []string` field. No `hostCmd`, no runtime fallback.
- **`Token()`** is now one line: `return c.execCliCommand(ctx, c.cmd)`.
- **Removed**: `isUnknownFlagError`, `buildCliCommands` (plural), `buildHostCommand`, `hostCmd` field.

## How is this tested?

Manual tests on versions 0.207.0 and 0.207.1

Unit tests in `config/cli_token_source_test.go`:
- `TestParseCliVersion` — standard versions, patch versions, malformed output, empty string, missing prefix.
- `TestCliVersion_AtLeast` — equal, higher/lower patch/minor/major, zero vs zero, zero vs nonzero.
- `TestBuildCliCommand` — table-driven: version x config → expected command. Covers: host-only, account host, profile+new CLI (uses --profile), profile+old CLI (falls back to --host), profile-only+old CLI (nil), zero version (detection failed, falls back to --host), neither profile nor host (nil).
- `TestNewCliTokenSource` — success with host, success with profile, CLI not found, neither profile nor host.
- `TestCliTokenSource_Token` — success, CLI error, invalid JSON.


config/cli_token_source.go
```diff
@@ -31,62 +32,110 @@ type cliTokenResponse struct {
...
-// buildHostCommand constructs the legacy --host based CLI command.
-func buildHostCommand(cliPath string, cfg *Config) []string {
-	cmd := []string{cliPath, "auth", "token", "--host", cfg.Host}
-	switch cfg.HostType() {
-	case AccountHost:
-		cmd = append(cmd, "--account-id", cfg.AccountID)
+// Token fetches an OAuth token by shelling out to the Databricks CLI.
+// If the working command has already been identified, it is called directly.
+// Otherwise, [probeAndExec] tries each command in order to find one that works.
+func (c *CliTokenSource) Token(ctx context.Context) (*oauth2.Token, error) {
```
renaudhartert-db, line 141: Separate the part that is about the public API (what users will see) and the part that is about the implementation details (what the developer will see). The latter is not useful to the users, they cannot see the methods that we are referencing. 

```suggestion
// Token fetches an OAuth token by shelling out to the Databricks CLI.
func (c *CliTokenSource) Token(ctx context.Context) (*oauth2.Token, error) {
  // If the working command has already been identified, it is called directly.
  // Otherwise, [probeAndExec] tries each command in order to find one that works.
```

mihaimitrea-db:
**Update**: Reopening this PR with a new approach.

The try-and-retry fallback mechanism from the previous iteration does not work for \`--profile\` because it is a [global Cobra flag](https://github.com/databricks/cli/blob/77101c9b8567/cmd/root/root.go) — old CLIs (< v0.207.1) silently accept it instead of reporting \`"unknown flag: --profile"\`, failing later with \`"cannot fetch credentials"\`. This made the \`--host\` fallback dead code.

The new approach uses **version detection**: run \`databricks version\` at init time, parse the semver, and use it to decide which flags the CLI supports. This also enables passing \`--force-refresh\` ([databricks/cli#4767](https://github.com/databricks/cli/pull/4767)) when the CLI version supports it.

See the updated PR description for details.

config/cli_token_source.go
```diff
@@ -31,62 +31,115 @@ type cliTokenResponse struct {
 	Expiry      string `json:"expiry"`
 }
 
+// cliVersion represents a parsed Databricks CLI semver version.
+type cliVersion struct {
```
renaudhartert-db: Could we use `golang.org/x/mod/semver` instead? It's already part of the `go.mod`.

config/cli_token_source.go
```diff
@@ -31,62 +31,115 @@ type cliTokenResponse struct {
...
+}
+
+// buildCliCommand constructs the CLI command for fetching an auth token.
+// The CLI version determines which flags are used.
+func buildCliCommand(ctx context.Context, cliPath string, cfg *Config, ver cliVersion) []string {
+	var cmd []string
+
+	// --profile is a global Cobra flag — old CLIs accept it silently but
+	// fail with "cannot fetch credentials" instead of "unknown flag".
+	// We use version detection to decide --profile vs --host.
```
renaudhartert-db:
```suggestion
	// Flag --profile is a global CLI flag and is recognized for all commands even
	// the ones that do not support it. Only use flag --profile in CLI versions that 
	// are known to support it in command `auth token`.
```

config/cli_token_source.go
```diff
@@ -31,62 +31,115 @@ type cliTokenResponse struct {
...
 	if cfg.Profile != "" {
-		primary := []string{cliPath, "auth", "token", "--profile", cfg.Profile}
-		if cfg.Host != "" {
-			// Build a --host fallback for old CLIs that don't support --profile.
-			return primary, buildHostCommand(cliPath, cfg)
+		if ver.AtLeast(cliVersionForProfile) {
+			cmd = []string{cliPath, "auth", "token", "--profile", cfg.Profile}
+		} else {
+			logger.Warnf(ctx, "Profile %q was specified but Databricks CLI %s does not support --profile (requires >= %s). Falling back to --host.", cfg.Profile, ver, cliVersionForProfile)
```
renaudhartert-db:
```suggestion
			logger.Warnf(ctx, "Databricks CLI %s does not support --profile (requires >= %s). Falling back to --host.", ver, cliVersionForProfile)
```

config/cli_token_source.go
```diff
@@ -31,62 +31,115 @@ type cliTokenResponse struct {
...
-	return buildHostCommand(cliPath, cfg), nil
-}
 
-// buildHostCommand constructs the legacy --host based CLI command.
-func buildHostCommand(cliPath string, cfg *Config) []string {
-	cmd := []string{cliPath, "auth", "token", "--host", cfg.Host}
-	switch cfg.HostType() {
-	case AccountHost:
-		cmd = append(cmd, "--account-id", cfg.AccountID)
+	if cmd == nil && cfg.Host != "" {
```
renaudhartert-db: This is equivalent but less bugprone in case `cmd` is initialized as an empty slice.

```suggestion
	if len(cmd) == 0 && cfg.Host != "" {
```

config/cli_token_source.go
```diff
@@ -31,62 +31,115 @@ type cliTokenResponse struct {
...
+	// --profile is a global Cobra flag — old CLIs accept it silently but
+	// fail with "cannot fetch credentials" instead of "unknown flag".
+	// We use version detection to decide --profile vs --host.
 	if cfg.Profile != "" {
-		primary := []string{cliPath, "auth", "token", "--profile", cfg.Profile}
-		if cfg.Host != "" {
-			// Build a --host fallback for old CLIs that don't support --profile.
-			return primary, buildHostCommand(cliPath, cfg)
+		if ver.AtLeast(cliVersionForProfile) {
+			cmd = []string{cliPath, "auth", "token", "--profile", cfg.Profile}
```
renaudhartert-db: Return directly here, the rest of the function is irrelevant if we reach this point.

config/cli_token_source.go
```diff
@@ -31,62 +31,115 @@ type cliTokenResponse struct {
...
+	cmd := buildCliCommand(ctx, cliPath, cfg, ver)
+	if cmd == nil {
+		return nil, fmt.Errorf("cannot configure CLI token source: neither profile nor host is set")
+	}
+	return cmd, nil
+}
+
+// buildCliCommand constructs the CLI command for fetching an auth token.
+// The CLI version determines which flags are used.
+func buildCliCommand(ctx context.Context, cliPath string, cfg *Config, ver cliVersion) []string {
```
renaudhartert-db: [optional] I believe a more readable and idiomatic way to write this code would be to organize it so that (i) the "normal branch" (i.e. use `--profile`) is on the lowest level of indentation, and (ii) each branch is terminal. Something like this:

```go
func buildCliCommand(ctx context.Context, cliPath string, cfg *Config, ver cliVersion) []string {
	if cfg.Profile == "" {
		return buildCliHostCommand(ctx context.Context, cliPath, cfg)
	}

	if !ver.AtLeast(cliVersionForProfile) {
		logger.Warnf(ctx, "CLI version XXX does not support --profile")
		return buildCliHostCommand(ctx context.Context, cliPath, cfg)
	}

	cmd := []string{cliPath, "auth", "token", "--profile", cfg.Profile}
	if cfg.HostType() == AccountHost {
		cmd = append(cmd, "--account-id", cfg.AccountID)
	}
	return cmd
}

func buildCliHostCommand(ctx context.Context, cliPath string, cfg *Config, ver cliVersion) []string {
	cmd = []string{cliPath, "auth", "token", "--host", cfg.Host}
	if cfg.HostType() == AccountHost {
		cmd = append(cmd, "--account-id", cfg.AccountID)
	}
	return cmd
}
```

It is a little longer but much easier to follow. Fallin back to using `--host` is now clearly the exception path and one does not need to understand how the host command is built to understand that.

config/cli_token_source.go
```diff
@@ -95,6 +148,8 @@ func (c *CliTokenSource) execCliCommand(ctx context.Context, args []string) (*oa
 	if err != nil {
 		var exitErr *exec.ExitError
 		if errors.As(err, &exitErr) {
+			// We intentionally discard exec.ExitError — the stderr text is the
+			// CLI's error contract; exit codes and process state are not useful.
```
renaudhartert-db: Why aren't they useful? Naively, I'd imagine that one might want to access this information for debugging what is happening with the CLI.

I could understand that we discard it because we do not want the error to be part of the SDK API contract but then I wonder why we are not discarding it at line 155 too.

config/cli_token_source_test.go
```diff
@@ -158,65 +253,117 @@ func TestBuildCliCommands(t *testing.T) {
...
+	path := filepath.Join(dir, "databricks")
+	content := append([]byte(body), make([]byte, databricksCliMinSize)...)
+	if err := os.WriteFile(path, content, 0755); err != nil {
+		t.Fatalf("failed to create mock CLI: %v", err)
+	}
+	return path
+}
 
-	cliName := "databricks"
+func TestNewCliTokenSource(t *testing.T) {
```
renaudhartert-db, line 333: Could we have this as table tests and merge them with the one above so that we have one comprehensive test suite for `NewCliTokenSource`?

config/cli_token_source.go
```diff
@@ -31,62 +32,123 @@ type cliTokenResponse struct {
...
 
+// Minimum CLI versions for flag support. Versions are in the format accepted
+// by [golang.org/x/mod/semver] (with leading "v").
+const cliVersionForProfile = "v0.207.1" // https://github.com/databricks/cli/pull/855
+
+// devBuildVersionPrefix is the sentinel the Databricks CLI uses when
+// buildVersion wasn't injected by goreleaser (e.g. a plain `go build`).
+// Such builds always compare less than any real release, which would disable
+// every feature gate; we surface an informational hint instead.
+// See https://github.com/databricks/cli/blob/main/internal/build/info.go.
```
renaudhartert-db: Let's move this part of the comment next to the logic that processes the value.

config/cli_token_source.go
```diff
@@ -31,62 +32,123 @@ type cliTokenResponse struct {
...
+	if cmd == nil {
+		return nil, fmt.Errorf("cannot configure CLI token source: neither profile nor host is set")
+	}
+	return cmd, nil
+}
+
+// buildCliCommand constructs the CLI command for fetching an auth token.
+// The CLI version determines which flags are used. Returns nil when neither
+// profile nor host is usable.
+func buildCliCommand(ctx context.Context, cliPath string, cfg *Config, ver string) []string {
```
renaudhartert-db: Let's pipe the error so that we can show a precise message (is it the host? is it the profile?). This is also more extendable. 

```suggestion
func buildCliCommand(ctx context.Context, cliPath string, cfg *Config, ver string) ([]string, error) {
```

config/cli_token_source.go
```diff
@@ -95,7 +157,10 @@ func (c *CliTokenSource) execCliCommand(ctx context.Context, args []string) (*oa
 	if err != nil {
 		var exitErr *exec.ExitError
 		if errors.As(err, &exitErr) {
-			return nil, fmt.Errorf("cannot get access token: %s", strings.TrimSpace(string(exitErr.Stderr)))
+			// Surface the CLI's stderr explicitly — exec.ExitError's own
+			// message is just "exit status N". Wrap err with %w so callers
+			// can still recover the *exec.ExitError via errors.As.
+			return nil, fmt.Errorf("cannot get access token: %s: %w", strings.TrimSpace(string(exitErr.Stderr)), err)
```
renaudhartert-db, line 174: Are we going to print two time the same thing? `Stderr` and `err`?

renaudhartert-db (approved):
LGTM modulo resolution of open comments.

PR #1621, opened 2026-04-13
Upload of big files (> 5Gb) to UC Volumes using multipart chunking

alexott:
## Summary

Implement upload of big files (> 5Gb) to UC Volumes using multipart chunking. That functionality already exists in Python SDK, but is missing in the Go SDK.

It will require some modifications in the codegen to support new functionality

## Why

https://github.com/databricks/terraform-provider-databricks/issues/5521 is asking for support of big files, but Go SDK doesn't have this functionality

### Interface changes

New functions are added to Files interface (requires codegen changes)

## How is this tested?

Unit tests were added


alexott:
@chrisst - that would be useful for TF integration

PR #1628, opened 2026-04-14
Pass --force-refresh to CLI auth token command

mihaimitrea-db:
## Summary

Pass `--force-refresh` to the Databricks CLI `auth token` command when the CLI supports it (>= v0.296.0), bypassing the CLI's internal token cache.

## Why

The SDK manages its own token caching via `CachedTokenSource`. When the SDK decides it needs a new token and shells out to `databricks auth token`, the CLI may return a cached token that is about to expire (or has already expired from the SDK's perspective). This creates unnecessary refresh failures and retry loops.

The CLI added a `--force-refresh` flag ([databricks/cli#4767](https://github.com/databricks/cli/pull/4767), motivated by [databricks/cli#4564](https://github.com/databricks/cli/issues/4564)) that bypasses its internal cache. By using this flag, the SDK is guaranteed a freshly minted token every time it asks for one.

With the version detection infrastructure from the parent PR (#1605), adding `--force-refresh` is a one-constant, one-`if` change.

### Interface changes

None. `CliTokenSource` is not part of the public API surface.

### Behavioral changes

- The SDK now appends `--force-refresh` when invoking `databricks auth token` if the CLI version is >= v0.296.0.
- A warning is logged when the CLI is too old for `--force-refresh`: `"Databricks CLI v%s does not support --force-refresh (requires >= v%s). The CLI's token cache may provide stale tokens."`

### Internal changes

- **`cliVersionForForceRefresh`**: new version constant `{0, 296, 0}`.
- **`buildCliCommand`**: adds a nil guard and appends `--force-refresh` after the base command is resolved.

## How is this tested?

Additional test cases in `TestBuildCliCommand`:
- host + v0.296.0 → appends `--force-refresh`
- host + v0.200.0 → no `--force-refresh`
- account host + v0.296.0 → appends `--force-refresh`
- profile + v0.296.0 → `--profile` + `--force-refresh`
- profile + v0.207.1 → `--profile` only (no `--force-refresh`)
- profile-only + v0.296.0 → `--profile` + `--force-refresh`
- zero version → `--host` only, no `--force-refresh`


renaudhartert-db (approved):
LGTM assuming resolution of the comments on the underlying PR.

PR #1636, opened 2026-04-17
Add SetDefaultHostMetadataResolverFactory

simonfaltum:
## Changes

[PR #1572](https://github.com/databricks/databricks-sdk-go/pull/1572) added `Config.HostMetadataResolver` so callers could override the SDK's `/.well-known/databricks-config` fetch on a per-Config basis. That covers "I have one Config and I want to wrap it."

The gap: programs that construct many Configs across their command surface (e.g. the Databricks CLI) end up copying the same `cfg.HostMetadataResolver = ...` assignment at every construction site, in the CLI roughly 10 sites across 7 files plus a guardrail test to catch drift.

This PR adds a package-level default consulted when a Config has no explicit resolver set. Callers set a factory once during startup; every subsequent Config gets the same resolver without per-site wiring. The Config-level field still takes precedence, so PR #1572's contract is unchanged.

### API

```go
// config/host_metadata.go
var DefaultHostMetadataResolverFactory func(*Config) HostMetadataResolver
```

Plain public variable, set once at init. Matches the stdlib pattern for single-default hooks: `http.DefaultClient`, `http.DefaultTransport`, `log.Default`. Callers needing per-Config or dynamic behaviour should use `Config.HostMetadataResolver` instead.

### Resolution order inside `Config.EnsureResolved`

1. If `Config.HostMetadataResolver` is set, use it.
2. Else, if `DefaultHostMetadataResolverFactory` is non-nil, invoke it with the resolving Config and use its return value. If it returns nil, fall through.
3. Else, SDK's default HTTP fetch (unchanged behavior for all existing callers).

## How the Databricks CLI will use this

The canonical Go idiom for "library A registers itself with library B" is a blank import that triggers an `init()` in A. This is how `database/sql` drivers (`_ "github.com/lib/pq"`), image codecs (`_ "image/png"`), and encoding formats register themselves.

After this PR lands and is bumped into the CLI, [CLI PR #5011](https://github.com/databricks/cli/pull/5011) will collapse from ~10 wired-in `hostmetadata.Attach(cfg)` calls + a guardrail test down to two small pieces:

**`repos/cli/libs/hostmetadata/resolver.go`** — set the caching factory at package init:

```go
func init() {
    config.DefaultHostMetadataResolverFactory = func(cfg *config.Config) config.HostMetadataResolver {
        return NewResolver(cfg.DefaultHostMetadataResolver())
    }
}
```

**`repos/cli/cmd/databricks/main.go`** — one blank import to pull the package in at startup:

```go
import (
    // Registers a disk-cached HostMetadataResolver with the SDK so every
    // Config the CLI constructs reuses the cached /.well-known lookup.
    _ "github.com/databricks/cli/libs/hostmetadata"
)
```

That's the full integration. Every Config the CLI creates, now and in the future from any new command a developer adds, automatically gets caching. No per-site `Attach` call to remember, no guardrail test to maintain, no new developer ever has to learn this mechanism exists to benefit from it.

### Experimental

Marked experimental to match the existing `HostMetadataResolver` field. No default behavior change for callers that never set `DefaultHostMetadataResolverFactory`.

## Tests

Three new tests in `config/config_test.go`, each using a small `withDefaultHostMetadataResolverFactory(t, factory)` helper that captures and restores the prior value, so tests never clobber each other via the package-level default:

- Factory is invoked when Config has no resolver; back-fill works end-to-end.
- Config-level resolver takes precedence (factory not consulted).
- Factory returning nil falls through to the SDK's HTTP fetch.

- `make fmt test lint` clean
- `go test ./config/... -count=1 -race` clean

Signed-off-by: simon <simon.faltum@databricks.com>

### DIFF
--- a/config/host_metadata.go
+++ b/config/host_metadata.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/databricks/databricks-sdk-go/common/environment"
 	"github.com/databricks/databricks-sdk-go/httpclient"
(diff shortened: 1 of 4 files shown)


config/host_metadata.go
```diff
@@ -38,6 +39,39 @@ type HostMetadata struct {
...
+// factory is invoked with the resolving Config and must return the resolver
+// to use for that Config (or nil to fall through to the SDK's default HTTP
+// fetch).
+//
+// Intended for programs that want a single hook to install a caching or
+// otherwise-customised resolver across every Config they construct, without
+// per-site wiring. Typically called once from an init() block in a package
+// that is blank-imported by the main binary — the canonical Go idiom for
+// library-registered defaults, as used by [database/sql] drivers and the
+// [image] package codecs.
```
renaudhartert-db: AI slope?

```suggestion
// per-site wiring. Typically called once from an init() block in a package
// that is blank-imported by the main binary.
```

config/host_metadata.go
```diff
@@ -38,6 +39,39 @@ type HostMetadata struct {
 // This allows callers to provide cached metadata without the SDK making an HTTP call.
 type HostMetadataResolver func(ctx context.Context, host string) (*HostMetadata, error)
 
+var (
+	defaultHostMetadataResolverFactoryMu sync.RWMutex
+	defaultHostMetadataResolverFactory   func(*Config) HostMetadataResolver
+)
```
renaudhartert-db: [optional] If users are expected to set this at `init` time, then you don't need the mutex. See `http` and `log` as references from the standard library.  

```
var DefaultHostMetadataResolverFactory func(*Config) HostMetadataResolver = nil
```

Users who need to use different resolvers within their code should use `Config.HostMetadataResolver` instead.

renaudhartert-db (approved):
LGTM, modulo comments.

config/host_metadata.go
simonfaltum: Switched to a plain `var DefaultHostMetadataResolverFactory`, dropped the mutex and the sync import. Matches the stdlib precedent you called out (http.DefaultClient, http.DefaultTransport, log.Default).

config/host_metadata.go
simonfaltum: Accepted the suggestion in 628abd1f.

PR #1639, opened 2026-04-21
apierr: Alias ErrResourceAlreadyExists to ErrAlreadyExists

simonfaltum:
## Changes

`apierr` defined two 409 sentinels with identical messages but distinct pointers:

- `ErrAlreadyExists` &mdash; mapped from `error_code: ALREADY_EXISTS` (gRPC canonical)
- `ErrResourceAlreadyExists` &mdash; mapped from `error_code: RESOURCE_ALREADY_EXISTS` (Databricks-specific)

Consumers using `errors.Is` with one sentinel did not match errors wrapped with the other, even though both represent the same class of 409 conflict. Which of the two `error_code` strings the backend returns is not guaranteed, so consumers had to either assert the generic `ErrResourceConflict` or duplicate both sentinels in their checks.

This came up downstream in https://github.com/databricks/cli/pull/5043 where a `TestDashboardAssumptions_WorkspaceImport` assertion on `ErrResourceAlreadyExists` started failing because the Lakeview API returns `ALREADY_EXISTS` rather than `RESOURCE_ALREADY_EXISTS`.

Alias `ErrResourceAlreadyExists` to `ErrAlreadyExists` so `errors.Is` matches either `error_code` regardless of which the backend emits. The exported symbol is preserved, so this is source- and binary-compatible for all existing callers.

Also extended `TestErrorMapping` to cover both error_codes against both sentinels and against the shared `ErrResourceConflict` parent.

## Tests

- [x] `go test ./apierr/...` passes, including new `RESOURCE_ALREADY_EXISTS` ↔ `ErrAlreadyExists` and `ALREADY_EXISTS` ↔ `ErrResourceAlreadyExists` cases.
- [x] `make fmt` clean.
- [x] `make lint` clean.


mihaimitrea-db (approved):
LGTM

PR #1640, opened 2026-04-21
Add WithDiscoveryHost option for discovery login flow

simonfaltum:
## Why

The discovery login flow (`WithDiscoveryLogin`) opens a browser to `https://login.databricks.com` to let the user authenticate and select a workspace. That host was hardcoded, which makes it impossible to exercise the flow against non-production instances of the login service during development or testing.

## Changes

Before: `discoveryTokenSource` used the `loginDatabricksHost` constant directly, with no way for callers to override.

Now: a new `WithDiscoveryHost(host)` option on `PersistentAuth` threads a host override through to `discoveryTokenSource`. When empty (the default), the production host is used, so existing callers behave identically.

Implementation notes:
- `BuildDiscoveryAuthorizeURL` keeps its public signature for backwards compatibility. It now delegates to an internal `buildDiscoveryAuthorizeURL(host, ...)` helper with the production default.
- Trailing slashes on the override host are trimmed to keep the resulting URL well-formed.
- The constant was renamed `defaultLoginDatabricksHost` to make its role as a fallback explicit.

## Test plan

- [x] New table-driven test covering override host and override host with trailing slash (default host is already covered by the existing `TestBuildDiscoveryAuthorizeURL`)
- [x] Existing `TestBuildDiscoveryAuthorizeURL` and `TestDiscoveryTokenSource_Challenge` still pass unchanged
- [x] `make fmt test lint` passes

### DIFF
--- a/credentials/u2m/persistent_auth.go
+++ b/credentials/u2m/persistent_auth.go
@@ -106,6 +106,10 @@ type PersistentAuth struct {
 	// When true, Challenge() uses the discovery token source instead of
 	// the standard authhandler flow.
 	discoveryMode bool
+
+	// discoveryHost overrides the default login.databricks.com host used by
+	// the discovery flow. Empty means the production host.
+	discoveryHost string
 }
 
 type PersistentAuthOption func(*PersistentAuth)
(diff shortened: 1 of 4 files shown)


credentials/u2m/persistent_auth.go
```diff
@@ -181,6 +185,16 @@ func WithDiscoveryLogin() PersistentAuthOption {
 	}
 }
 
+// WithDiscoveryHost overrides the default https://login.databricks.com host
+// used by the discovery login flow. Intended for testing and development
+// against non-production environments; has no effect unless WithDiscoveryLogin
+// is also set. Trailing slashes on host are trimmed.
+func WithDiscoveryHost(host string) PersistentAuthOption {
```
hectorcast-db, line 194: Do we want to validate or add the protocol it not set?
Feels like this would be a common mistake:
```
WithDiscoveryHost("myhost.com") 
```

simonfaltum: Good catch, fixed in dfa3a3b. If the host has no scheme we now prepend https:// automatically (and preserve http:// for localhost testing), matching the trailing-slash trimming we already do. Added a table-driven test for the normalization cases.

hectorcast-db (approved):

PR #1646, opened 2026-04-21
credentials/u2m: remove file token cache and host-key dualWrite

simonfaltum:
> **Do not merge until databricks/cli#5056 ships in a released CLI.** That PR gives the CLI its own `FileTokenCache` and `DualWrite` helper. Until it lands in a release, this SDK change would silently downgrade the CLI's OAuth persistence to in-memory on every non-auth command.

## Changes

Second of a three-PR sequence that moves file-based OAuth token cache management out of the Go SDK and into the Databricks CLI. This PR is the SDK half: delete what the CLI now owns, keep everything the CLI still depends on.

**Deleted:**
- `credentials/u2m/cache/file.go` and its tests — removes `FileTokenCache`, `NewFileTokenCache`, `FileTokenCacheOption`, `WithFileLocation`, and the private `tokenCacheFile` type.
- `PersistentAuth.dualWrite` and `PersistentAuth.hostCacheKey` methods in `credentials/u2m/persistent_auth.go`.
- The host-key fallback branch inside `PersistentAuth.loadToken`. Lookup is now a single `GetCacheKey()` call.
- dualWrite- and fallback-specific tests in `persistent_auth_test.go` and `discovery_token_source_test.go`.

**Added:**
- `credentials/u2m/cache/memory.go` with `NewInMemoryTokenCache()` returning a mutex-protected `TokenCache`. Semantics match the old `FileTokenCache` (`Store(key, nil)` deletes, `Lookup` miss returns `ErrNotFound`). Tests run cleanly under `-race`.

**Changed:**
- `NewPersistentAuth` now defaults to `cache.NewInMemoryTokenCache()` when no `WithTokenCache` option is passed. This is a **breaking change** for any direct consumer that relied on the previous implicit file-backed default. The SDK workdir audit found no consumers outside the CLI, which is addressed by #5056 in its companion.
- Call sites in `refresh()`, `Challenge()`, and `discovery_token_source.go` now `Store` once under `GetCacheKey()` instead of going through `dualWrite`.
- `loadToken` and `ForceRefreshToken` doc comments rewritten to describe the single-key world.
- `HostCacheKeyProvider` doc comment updated: `PersistentAuth` no longer uses the interface internally, but it remains exported so that external token cache implementations (like the CLI's `FileTokenCache`) can type-assert on it to mirror tokens under the legacy host key for cross-SDK compatibility with older Python/Java SDKs.

**Kept exported (no changes):**
- `TokenCache` interface and `ErrNotFound` sentinel in `credentials/u2m/cache/cache.go`.
- `OAuthArgument`, `HostCacheKeyProvider`, `DiscoveryOAuthArgument` interfaces in `credentials/u2m/`.
- `BasicWorkspaceOAuthArgument`, `BasicAccountOAuthArgument`, `BasicUnifiedOAuthArgument` and their `GetHostCacheKey()` methods plus compile-time interface assertions.
- Their `GetHostCacheKey` tests.

## Tests

- `make fmt` clean, `make lint` clean, `make test` passes (1375 tests).
- `go test -race ./credentials/u2m/... ./credentials/u2m/cache/... ./config/...` clean.
- New `memory_test.go` covers Store/Lookup roundtrip, nil-Store delete, missing-key `ErrNotFound`, and concurrent Store/Lookup from multiple goroutines.
- Deleted tests were scoped to dualWrite and host-key-fallback paths only. `TestToken`, `TestToken_WithProfile`, and the refresh/Challenge coverage remain.

## Design doc

See `databricks/cli` design doc at `documents/fy2027-q2/cli-ga/2026-04-21-move-token-cache-to-cli-plan.md` for the full three-PR sequence. Companion PR: https://github.com/databricks/cli/pull/5056.


hectorcast-db (approved):
LGTM

PR #1648, opened 2026-04-23
Remove legacy tag-triggered release workflow

hectorcast-db:
NO_CHANGELOG=true

## Summary
The legacy tag-triggered release workflow is superseded by the secure release pipeline in
`databricks/secure-public-registry-releases-eng`, which now dispatches `tagging.yml` here
to create the tag and consumes it via workflow_dispatch. This PR removes the obsolete
tag-trigger path while preserving the `merge_group` CI dry-run that validates the
GoReleaser configuration on every queued merge.

## What changed
- Dropped the `on: push: tags: ['*']` trigger from `.github/workflows/release.yml`.
- Removed the tag-gated "Write release notes to file" step.
- Removed the "Run GoReleaser (Release)" step (gated on `startsWith(github.ref, 'refs/tags/v')`).
- Dropped the `if: !startsWith(github.ref, 'refs/tags/v')` guard on the Snapshot step, since the workflow now only runs in `merge_group` context.

## What's preserved
- `merge_group` trigger for CI.
- GoReleaser snapshot build (dry-run validation) that runs on every queued merge.
- `tagging.yml` (untouched) — the new entry point invoked by the secure release pipeline.

## Blast radius
Nothing consumes this workflow's output outside of GitHub Releases, which the new pipeline
now produces.

### DIFF
--- a/.github/workflows/release-build-check.yml
+++ b/.github/workflows/release-build-check.yml
@@ -1,4 +1,4 @@
-name: release
+name: Release Build Check
 
 on:
   merge_group:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,6 @@
 name: release
 
 on:
-  push:
-    tags:
-      - "*" 
   merge_group:
     types: [checks_requested]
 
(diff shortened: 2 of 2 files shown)


mihaimitrea-db (approved):

hectorcast-db:
Renamed the workflow file to `release-build-check.yml` and updated the top-level `name:` to `Release Build Check` in 39ac845c. The workflow only runs a `merge_group` dry-run build — no release path remains — so both the filename and the display name now match its real purpose.

Heads-up: any branch protection rules or required-status-check settings that reference the old check name (`release`) will need to be updated to `Release Build Check` once this merges, otherwise the merge queue will block on a missing check.

mihaimitrea-db (approved):

PR #1682, opened 2026-05-15
useragent: honor AI_AGENT and pass raw fallback values through

renaudhartert-db:
## Why

The Go SDK detects AI coding agents and surfaces them as `agent/<name>` in the User-Agent. Today the generic fallback (when no proprietary env var fires) only honors the agents.md `AGENT=<name>` standard. Vercel's `@vercel/detect-agent` library uses a parallel `AI_AGENT=<name>` convention that tools in the Vercel ecosystem set instead; we currently miss those.

Separately, the existing fallback coerces any unrecognized `AGENT` value to the literal string `"unknown"`. That buries useful signal. For example a tool setting `AGENT=claude-code_2-1-141_agent` ends up as `agent/unknown`,
discarding the very signal (tool name plus version variant) we want to see. Bucketing should happen downstream of the SDK. 

## Changes

Two behavior changes in `useragent/agent.go`:

1. **`AI_AGENT` fallback.** Add `AI_AGENT=<name>` as a secondary fallback after `AGENT=<name>`. `AGENT` wins when both are set to non-empty values; empty is treated as unset for both. Explicit product matchers (e.g. `CLAUDECODE=1`) still always win over both.

2. **Raw passthrough instead of `"unknown"`.** Drop the known-product lookup in the fallback. The value is piped through the existing `Sanitize()` helper (so disallowed chars become `-` and the User-Agent allowlist `[0-9A-Za-z_.+-]+` is satisfied) and capped at 64 chars to keep the header bounded. Known products like `cursor` or `claude-code` pass through unchanged because they already satisfy the allowlist.

Same change should land in `databricks-sdk-py` and `databricks-sdk-java`; sibling PRs to follow.

## Test plan

- [x] `go test ./useragent/...` passes
- [x] `gofmt -l useragent/` clean
- [x] `AI_AGENT=<known product>` returns the product name
- [x] `AI_AGENT=<unrecognized>` returns the raw sanitized value (no longer `\"unknown\"`)
- [x] `AGENT` wins over `AI_AGENT` when both are non-empty
- [x] Empty `AGENT` falls through to `AI_AGENT`
- [x] Disallowed chars in `AGENT` / `AI_AGENT` are sanitized to `-`
- [x] Values longer than 64 chars are truncated
- [x] Explicit matcher (e.g. `CLAUDECODE=1`) still wins over both fallbacks

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,6 +6,13 @@
 
 ### New Features and Improvements
 
+* Honor the Vercel `AI_AGENT=<name>` env var as a secondary fallback for
+  AI agent detection in the User-Agent header (after the agents.md
+  `AGENT=<name>` standard). Unrecognized fallback values now pass through
+  the User-Agent sanitized and length-capped at 64 chars instead of being
+  coerced to `agent/unknown`, so versioned variants such as
+  `claude-code_2-1-141_agent` surface as-is.
+
 ### Bug Fixes
 
 ### Documentation
--- a/useragent/agent.go
+++ b/useragent/agent.go
@@ -13,12 +13,25 @@ type knownAgent struct {
 	product string
 }
 
-// agentEnvVar is the agents.md standard env var. When set to a value we
-// don't specifically recognize, detection falls back to "unknown".
-const agentEnvVar = "AGENT"
+const (
+	// agentEnvVar is the agents.md standard env var.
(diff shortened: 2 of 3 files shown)


renaudhartert-db:
Recreating from origin branch instead of fork so CI can authenticate to JFrog (OIDC tokens are not available to fork PRs). See new PR.

PR #1713, opened 2026-06-09
Encode tag key as a single path segment in tag-policy and tag-assignment APIs

lizhenxiang-db:
## Summary

Percent-encode the governed tag key when it travels as a URL *path* parameter across the tag-policy and entity-tag-assignment APIs, so keys containing `/` route to the correct endpoint.

## Why

Governed tag keys may be hierarchical and contain `/` (e.g. `a/b`). The generated clients interpolate the key straight into the request path (e.g. `fmt.Sprintf("/api/2.1/tag-policies/%v", request.TagKey)`), so a raw `/` is treated as a path separator and the request resolves to no matching endpoint (`404` / `ENDPOINT_NOT_FOUND`). `Get`, `Delete`, and `Update` are affected.

A concrete consequence: the Terraform provider cannot refresh a `databricks_tag_policy` whose key contains `/` — every plan reads it as missing and tries to recreate it, which then fails because the policy already exists.

`Create` does not have this problem: it sends the key in the JSON request body, where `/` is valid and must be stored verbatim.

### Interface changes

None. Method signatures are unchanged; only the encoding of the existing request path differs.

### Behavioral changes

The tag-key path parameter is now `url.PathEscape`d (`/` → `%2F`) before the request is sent, in:

- **`tags.TagPoliciesAPI`** — `GetTagPolicy`, `DeleteTagPolicy`, `UpdateTagPolicy`
- **`tags.WorkspaceEntityTagAssignmentsAPI`** — `GetTagAssignment`, `DeleteTagAssignment`, `UpdateTagAssignment`
- **`catalog.EntityTagAssignmentsAPI`** — `Get`, `Delete`, `Update`

The server decodes the key back to its original value, so behavior is unchanged for keys without special characters and fixed for keys that contain them. Only the tag-key parameter is encoded — `entity_type` / `entity_id` / `entity_name` are left untouched. The `Create` methods are intentionally not overridden (they send the key in the body).

### Internal changes

- `service/tags/ext_impl.go` — hand-written overrides for `TagPoliciesAPI` and `WorkspaceEntityTagAssignmentsAPI`.
- `service/catalog/ext_entity_tag_assignments.go` — hand-written overrides for `EntityTagAssignmentsAPI`.
- `NEXT_CHANGELOG.md` — Bug Fixes entry.

## How is this tested?

`qa.HTTPFixtures` unit tests in `service/tags/ext_impl_test.go` and `service/catalog/ext_entity_tag_assignments_test.go` assert the request path is encoded (`/` → `%2F`) for the tag-policy and both entity-tag-assignment APIs. `go build`, `go vet`, and `go test` pass locally for both packages.

This pull request and its description were written by Isaac.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* Encode the governed-tag key as a single path segment when it is sent as a URL path parameter, so hierarchical keys containing `/` route correctly instead of being split into extra path segments and resolving to no endpoint (`404` / `ENDPOINT_NOT_FOUND`). Affects `GetTagPolicy`/`DeleteTagPolicy`/`UpdateTagPolicy` ([tags.TagPoliciesAPI](https://pkg.go.dev/github.com/databricks/databricks-sdk-go/service/tags#TagPoliciesAPI)), `GetTagAssignment`/`DeleteTagAssignment`/`UpdateTagAssignment` ([tags.WorkspaceEntityTagAssignmentsAPI](https://pkg.go.dev/github.com/databricks/databricks-sdk-go/service/tags#WorkspaceEntityTagAssignmentsAPI)), and `Get`/`Delete`/`Update` ([catalog.EntityTagAssignmentsAPI](https://pkg.go.dev/github.com/databricks/databricks-sdk-go/service/catalog#EntityTagAssignmentsAPI)).
+
 ### Documentation
 
 ### Internal Changes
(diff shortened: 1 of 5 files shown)


renaudhartert-db (approved):

renaudhartert-db:
For context, the failing checks here are not caused by the change: this PR is opened from a personal fork, and fork PRs do not receive GitHub Actions OIDC tokens, so `setup-build-environment` fails at JFrog auth before tests run. Required checks can therefore never pass on this PR. I pushed the same commits to an origin branch as #1716 so CI can run; I will close this one once #1716 merges.

renaudhartert-db:
Merged via #1716 (identical commits, squashed into main as part of the merge queue). Closing this one since fork PRs cannot pass the OIDC-gated required checks. Thanks for the fix @lizhenxiang-db.

PR #1718, opened 2026-06-12
Deduplicate identical key/value pairs in user agent builder

Divyansh-db:
## Summary

`useragent.data.With()` appended every key/value pair unconditionally and never deduplicated. When the same user-agent dimension is injected repeatedly onto a reused or long-lived `context.Context` — for example a caller that threads a single context across many operations — the per-context user-agent data accumulates duplicate entries, so `FromContext` produces a `User-Agent` string that grows without bound (e.g. `... sdk/x sdk/x resource/y resource/y ...`). Sufficiently large headers can be rejected or truncated downstream.

Note that `With`'s doc comment claimed it "always uses the latest value for a given alphanumeric key", but the implementation always appended, and existing tests rely on multiple **distinct** values per key being preserved (e.g. multiple partners). This change keeps that behavior and only removes exact duplicates.

## Change

`data.With()` now skips appending a key/value pair when an identical pair is already present. Distinct values for the same key are still preserved, so no existing behavior or tests change. The doc comment is updated to describe the actual behavior.

## Testing

- `go test ./useragent/` — all pass, including `TestMultiplePartners` and `TestFromContext_Custom`.
- Added `TestInContext_DeduplicatesIdenticalPairs`: 100 repeated identical injections collapse to one entry each; distinct values still preserved.
- `go vet ./useragent/` and `gofmt` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code).


useragent/user_agent.go
```diff
@@ -110,7 +110,11 @@ func (u info) String() string {
 
 type data []info
 
-// With always uses the latest value for a given alphanumeric key.
+// With appends the key/value pair, skipping it when an identical key/value
+// pair is already present. De-duplicating exact pairs keeps the user agent from
+// growing without bound when the same dimension is injected repeatedly onto a
+// reused context (for example, the same resource pushed onto a long-lived
+// context many times). Distinct values for the same key are still preserved.
```
renaudhartert-db:
```suggestion
// reused context. Distinct values for the same key are still preserved.
```

renaudhartert-db (approved):

PR #1775, opened 2026-07-23
Add Pi coding agent to the known-agents table

lennartkats-db:
## Summary

Pi (https://pi.dev/, npm @earendil-works/pi-coding-agent) sets PI_CODING_AGENT in the environment of every subprocess its bash/exec tool spawns, but does not set the generic AGENT/AI_AGENT var, so it needs an explicit matcher. Without it, CLI usage driven by Pi is attributed to unknown instead of "pi".

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### New Features and Improvements
 
+* Add the Pi coding agent (`PI_CODING_AGENT`) to AI agent detection in the User-Agent header, so CLI usage driven by Pi is attributed to `pi`.
+
 ### Bug Fixes
 
 ### Documentation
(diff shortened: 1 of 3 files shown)


renaudhartert-db (commented):
Let's see if we can get this merged in Pi: https://github.com/earendil-works/pi/issues/7132

renaudhartert-db:
> Let's see if we can get this merged in Pi: [earendil-works/pi#7132](https://github.com/earendil-works/pi/issues/7132)

Yes, we can!

lennartkats-db:
@renaudhartert-db that's great!! That will also help with older SDK versions that just looked at `AI_AGENT`. Should we still merge the present PR to detect older Pi versions?

renaudhartert-db:
> @renaudhartert-db that's great!! That will also help with older SDK versions that just looked at `AI_AGENT`. Should we still merge the present PR to detect older Pi versions?

Good point, I didn't consider that side of the backward compatibility.

renaudhartert-db (approved):

PR #1793, opened 2026-08-05
Recover from concurrent OAuth token cache writes

mihaimitrea-db:
## Summary

Recovers U2M OAuth refreshes when another process wins a concurrent token cache write. The process whose write fails briefly reloads the cache and returns the fresh token stored by the winner.

## Why

Multiple CLI or SDK processes can read the same cached OAuth token, refresh it concurrently, and then race to replace the cache entry. On macOS, concurrent Keychain writes can cause one process to fail with `exit status 45`, even though another process successfully stored a valid token.

This addresses [databricks/cli#6051](https://github.com/databricks/cli/issues/6051) without adding cross-platform process locks. It does not prevent concurrent refreshes; it prevents a losing cache write from becoming an authentication failure when a fresh winner is available.

### Interface changes

None.

### Behavioral changes

- After a refreshed token fails to be stored, `PersistentAuth` reloads the cache immediately and then after 25, 50, 100, and 200 milliseconds.
- A changed, valid cached token is returned when its expiry is within one minute of the token obtained by the current process.
- The original cache update error is preserved when no suitable token appears within the bounded 375-millisecond recovery window.

### Internal changes

- Added freshness comparison and cache update recovery helpers.
- Added tests covering delayed winner discovery, changed-token detection, validity, and expiry tolerance.
- Added a bug-fix changelog entry.

## How is this tested?

- `make test` (race-enabled full unit test suite)
- `make lint`
- `go test -count=1 ./credentials/u2m/...`

The real macOS Keychain race has not yet been stress-tested on macOS. The unit test models the observed sequence: the cache write fails, the first reload still sees the original token, and a later reload sees the fresh token stored by another process.

### DIFF
--- a/credentials/u2m/persistent_auth.go
+++ b/credentials/u2m/persistent_auth.go
@@ -41,6 +41,17 @@ const (
 	// token is proactively refreshed. This prevents callers from receiving
 	// near-expired tokens that may expire before the next request.
 	tokenRefreshBuffer = 5 * time.Minute
+
+	// Cache update recovery checks immediately and then waits for 25, 50, 100,
+	// and 200 milliseconds. This gives a concurrent cache writer time to finish
+	// while bounding the delay for persistent storage failures to 375 milliseconds.
+	cacheUpdateRecoveryAttempts = 5
+
+	cacheUpdateRecoveryInitialDelay = 25 * time.Millisecond
+
+	// Concurrent refreshes can finish at slightly different times, so their
+	// expiration times need not be identical.
+	cacheUpdateRecoveryExpiryDelta = time.Minute
 )
 
 var (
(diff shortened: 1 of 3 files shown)


credentials/u2m/persistent_auth.go
```diff
@@ -316,6 +328,42 @@ func needsRefresh(t *oauth2.Token) bool {
 	return !t.Expiry.IsZero() && time.Until(t.Expiry) < tokenRefreshBuffer
 }
 
+// isFreshReplacement reports whether cached changed from old and has
+// approximately the same lifetime as candidate.
+func isFreshReplacement(old, candidate, cached *oauth2.Token) bool {
+	if cached.AccessToken == old.AccessToken || !cached.Valid() {
+		return false
+	}
+	if candidate.Expiry.IsZero() || cached.Expiry.IsZero() {
```
parthban-db: Can we avoid skipping the expiry comparison when only one side is zero? If
`candidate.Expiry` is zero, this accepts a changed cached token expiring in one
minute because `Valid()` only checks whether it is valid now, so we can return a
token inside the five-minute refresh window.

parthban-db (approved):

PR #1807, opened 2026-08-15
Accept numeric strings for int64 response fields

renaudhartert-db:
## Summary

Accepts API `int64` fields encoded as either JSON numbers or decimal strings when generated models use `marshal.Unmarshal`.

## Why

Databricks services do not consistently encode 64-bit integers: the same field may be returned as `123` or `"123"`. Go's `encoding/json` cannot accept both representations for a plain `int64`, and changing generated fields to a custom type would break the public API.

This change normalizes quoted integers before the existing model unmarshalling logic runs. It uses destination type information to rewrite only matching `int64` tokens, preserving unrelated JSON bytes such as `json.RawMessage` payloads. A companion code-generator change will make all generated model structs use `marshal.Unmarshal`.

### Interface changes

None.

### Behavioral changes

- `marshal.Unmarshal` accepts numeric and decimal-string representations of `int64` fields, including nested slices and maps.
- JSON `null` continues to leave populated scalar `int64` values unchanged.
- Invalid and overflowing decimal strings return an error.

### Internal changes

- Adds a type-aware normalization pass using `json.Decoder` offsets so only quoted `int64` tokens are replaced.
- Handles promoted fields from anonymous embedded structs.
- Preserves raw JSON values byte-for-byte when they do not require normalization.
- Updates `NEXT_CHANGELOG.md`.

## How is this tested?

- `make test`
- `make lint`
- `go test -race ./marshal`
- `go test ./marshal ./httpclient ./httpclient/fixtures ./config/experimental/auth/oidc`

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Recover from concurrent OAuth token cache writes by using the fresh token stored by another process.
 - Fix Spark runtime selection for major-only runtime versions.
+- Accept JSON numbers and decimal strings when unmarshalling `int64` API response fields.
 
 ### Documentation
 
--- a/marshal/int64.go
+++ b/marshal/int64.go
@@ -1,0 +1,248 @@
+package marshal
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"reflect"
+	"strconv"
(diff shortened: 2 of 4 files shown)


renaudhartert-db:
Replaced by a same-repository PR so OIDC-backed CI checks can run.

PR #1817, opened 2026-08-20
Add group role assumption to SDK OAuth authentication

mihaimitrea-db:
## Summary

Adds group assumption to the Go SDK's OAuth M2M and workload identity federation authentication flows. Clients configured with a group never fall back to built-in authentication methods that can only provide normal access.

This draft is stacked on #1812, which introduces the shared `group_id` configuration and U2M authorization support. U2M support is not part of this PR and will be added after support is added in the CLI.

### Interface changes

- **`DatabricksOIDCTokenSourceConfig.GroupID`** — selects the group passed to the shared Databricks WIF token exchange.
- **`Config.GroupID` behavior** — the field introduced by the stacked base PR is now honored by built-in OAuth M2M and WIF strategies.

### Behavioral changes

- OAuth M2M and all shared WIF adapters send `assume_group` when `GroupID` is configured.
- Workspace, account, unified, and discovery paths consistently forward the group selection. Public comments warn that account and unified authentication are currently unsupported.
- PAT, basic, metadata-service, Azure-native, Google-native, `github-oidc-azure`, and `databricks-cli` authentication return a clear unsupported-role error instead of producing normal credentials.
- Default-chain fallback can continue between role-capable strategies, but unsupported strategies cannot be selected. Exhaustion preserves the existing generic default-credentials error.
- Credential caches remain scoped per client, preventing clients with different groups from sharing M2M or WIF access tokens.
- OAuth failures are returned as-is, without retrying a request that omits `assume_group`.

### Internal changes

- Adds a shared unsupported-role guard for built-in credential strategies.
- Extends M2M, WIF, credential-chain, configuration precedence, cache-isolation, and token-form tests.

## How is this tested?

- Live workspace M2M, CLI, and PAT validation. WIF not covered

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -8,7 +8,8 @@
 
 - Added group role assumption to built-in SDK OAuth M2M and workload identity
   federation authentication, with safeguards that prevent other built-in
-  credential strategies from silently obtaining normal-access credentials.
+  credential strategies, including Databricks CLI authentication, from
+  silently obtaining normal-access credentials.
 
 ### Bug Fixes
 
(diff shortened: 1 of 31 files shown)


hectorcast-db (approved):
It seems that the base is not in sync? I see missing parts which are on the other PR.
Otherwise lgtm

PR #1835, opened 2026-09-08
Replace workspace integration test with HTTP fixtures

tejaskochar-db:
## Summary

Converts `TestMwsAccWorkspaces` into a deterministic unit test focused on behavior owned by the SDK: workspace creation, waiter polling to `RUNNING`, credential reassignment, and deletion.

## Why

The previous journey also created and deleted storage configurations and credentials around a live workspace. Those dependency relationships are owned by the account backend: the SDK only passes opaque resource IDs, while the server decides when a workspace has attached or detached them. An HTTP-fixture unit test can replay responses, but it cannot make those relationships exist or prove that detachment completed. Keeping that setup and cleanup in this unit test would therefore imply coverage it cannot provide; backend attachment and eventual-detachment behavior belongs in server-side end-to-end tests.

The live journey has become a release bottleneck because workspace deletion can be reported before those backend relationships are fully detached, causing credential and storage cleanup to fail. The standalone live SDK CRUD journeys for credentials and storage configurations remain in `TestMwsAccCredentials` and `TestMwsAccStorage`.

### Interface changes

None.

### Behavioral changes

None.

### Internal changes

- Exercise the account workspace client with five deterministic requests: create, poll, reassign credentials, poll, and delete.
- Remove live credential and storage provisioning, cleanup retries, and unrelated list/name helper checks from this focused test.
- Record the test change in the next changelog.

## How is this tested?

- The focused workspace test passes with ambient OAuth and Google credentials present, verifying that its fixture client remains isolated in integration runners while checking the exact request bodies, paths, and waiter results.
- `go test -mod=vendor -short -race ./internal -count=1` verifies the full internal package with the race detector.


Divyansh-db:
Two non-blocking suggestions:

1. This drops `List`, `GetByWorkspaceName`, and `WorkspaceWorkspaceNameToWorkspaceIdMap` from coverage. Those are client-side filtering/mapping logic (SDK-owned), not thin passthroughs, and this test was the only thing exercising them. Could you add a `GET` workspaces-list fixture and assert `List`, `GetByWorkspaceName`, and the name-to-ID map (same invariants the old test checked)?

2. Both waiters return on the first poll, since the `GET` fixtures return `RUNNING` immediately, so the polling loop is never actually exercised. Consider an intermediate `GET` returning a non-terminal status (`PROVISIONING`) before `RUNNING`, at least for the create waiter, so the test proves it keeps polling until `RUNNING`. The retry backoff adds ~1s per extra poll, so doing it once is enough.

Fine as a follow-up if you'd rather keep this PR focused. Rest LGTM.

Divyansh-db (approved):

tejaskochar-db:
I'd keep testing those as separate. They have the same generated logic repeated for many services so they should be tested once.

PR #1846, opened 2026-09-16
Add Config.Headers for custom request headers

Sankalp-Mittal:
## Summary

Add `Config.Headers` to set custom HTTP headers on every request made by the client, on top of normal authentication.

## Why

Callers occasionally need to attach a fixed header to every request — for example to route traffic to a test instance via `x-databricks-traffic-id`. Today the only options are wrapping `HTTPTransport` or setting headers per call. A small generic field covers every generated service method and raw client call, while keeping any such routing convention out of the SDK itself.

### Interface changes

- **`Config.Headers map[string]string`** — additional headers set on every request. Applied by a request visitor after `AuthVisitor`, so it must not be used to override authentication headers such as `Authorization`. Set programmatically; not loaded from an environment variable or a `.databrickscfg` profile.

  ```go
  cfg := &databricks.Config{
      Headers: map[string]string{"x-databricks-traffic-id": "testenv://liteswap/<target>"},
  }
  ```

### Behavioral changes

- None when `Headers` is empty (the default). When set, each header is applied on top of normal authentication.

This pull request and its description were written by Isaac.


config/config.go
```diff
@@ -240,6 +240,10 @@ type Config struct {
 	// Debug HTTP headers of requests made by the provider. Default is false.
 	DebugHeaders bool `name:"debug_headers" env:"DATABRICKS_DEBUG_HEADERS" auth:"-"`
 
+	// Headers are additional HTTP headers set on every request made by the
+	// client, on top of normal authentication.
```
hectorcast-db: I wonder if the Headers may change from call to call. If so, it may be useful to define the interface as a function which sets headers.

Sankalp-Mittal: For the case I'm gonna we using it for static headers are enough, but I'm not certain that there can be cases where the headers change from call to call, did you have any particular cases in mind?

hectorcast-db: I cannot think of a useful case right now, but since this is a public feature, I think it may be better to do be flexible and not have to introduce a second way later to set the headers and have to maintain both.

Sankalp-Mittal: added dynamic headers, along with a small helper to set static header's if needed

hectorcast-db (approved):

